PR #1, opened 2018-02-21
v4l2 control of usbtv kernel module.

cbassa:
Fushicai USBTV007 grabbers (1b71:3002) return very dark images in Ubuntu 16.04 and later due to bug fix of v4l2 controls ([https://github.com/torvalds/linux/commit/c53a846c48f21c909102677c127ece94b1247668](url)). Brightness and contrast values of the usbtv kernel module can be controlled by v4l2-ctl from the v4l-utils package.

### DIFF
--- a/devices
+++ b/devices
@@ -8,6 +8,9 @@ usage:
 restrictions:
  - only 720x576@25fps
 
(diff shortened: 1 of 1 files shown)


dvida:
Thanks!

PR #14, opened 2019-05-04
delete archived .tgz files too when no free space is available

adalava:
RMS calculates how much storage space it needs for the night and when free space is insufficient, the oldest folders from RMS_data/ArchivedFiles and RMS_data/CapturedFiles are removed until enough storage space is available.
Code currently gives up when no folders are left to delete, however there's still some compressed captures left on RMS_data/ArchivedFiles that could be removed as well. This patch adds the ability to remove these files as last resort when no free storage space is available.


dvida:
Sweet! Could you please change the target branch to "dev" so I can test it first before merging it with the master branch?

Issue #21, opened 2019-10-18
Specify latest py2 compatible astropy


apevec:
Reported by Maciej in https://groups.io/g/globalmeteornetwork/message/468

Processing dependencies for RMS==0.1
Searching for astropy>=2.0.3
Reading https://pypi.org/simple/astropy/
Downloading https://files.pythonhosted.org/packages/ea/63/02ada19599c7ad34b4984e797c555cacf2c358d48824ffe331e029574a2c/astropy-3.2.2.tar.gz#sha256=8553f4a95e9938a0c7ae520633f7172356824b8eb324dd967fca6baf00ac19bf
Best match: astropy 3.2.2
Processing astropy-3.2.2.tar.gz
Writing /tmp/easy_install-khgwDY/astropy-3.2.2/setup.cfg
Running astropy-3.2.2/setup.py -q bdist_egg --dist-dir /tmp/easy_install-khgwDY/astropy-3.2.2/egg-dist-tmp-YUOEAh
ERROR: Python >=3.5 is required by astropy

apevec:
Collecting astropy~=2.0.3
  Downloading https://files.pythonhosted.org/packages/bc/ad/f108a5bd8aba3d5f5d4d5090a8802a78e4e50e8150cb090b6d43b399db5f/astropy-2.0.15-cp27-cp27mu-manylinux1_x86_64.whl (6.1MB)

apevec:
TODO make it work with both py2 and py3

mrprompt:
@apevec you can add these two libs into requirements.txt file, it's like:

```
astropy~=3.2.2; python_version <= '3.7'
astropy~=2.0.3; python_version >= '2.7'
```

apevec:
not needed, originally reported issue is NOT reproducible
astropy>=2.0.3 in requirements.txt works fine, installing latest compatible version for the Python2 https://pypi.org/project/astropy/2.0.16/

PR #22, opened 2019-10-18
Specify latest py2 compatible astropy

apevec:
astropy 3.* is py3 only, requirements.txt need to specify astropy 2.*
while RMS is on Python 2.7

Fixes https://github.com/CroatianMeteorNetwork/RMS/issues/21

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,8 @@ pyephem>=3.7.6.0
 cython>=0.27.3
 scipy>=1.0.0
 Pillow>=4.3.0
-astropy>=2.0.3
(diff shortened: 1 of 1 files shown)


apevec:
@dvida do not merge yet there's TODO make it work with both py2 and py3

dvida:
Thanks for this, can I merge it now?

apevec:
> Thanks for this, can I merge it now?

not yet, original report is not reproducible and ~=3.0 needs to be confirmed if it is enough or needs to be higher 3.x (latest is 3.2.2)

markmac99:
this can be closed i think, the issue no longer exists? RMS working anyway.

PR #23, opened 2019-10-25
Cleanup and Docs

mrprompt:
Move some files to Guides directory, removing editor files and add documentation to use internal cards.


dvida:
Hi Thiago, thanks for doing this!
I only have one concern, and that's the config file. The way things are set up everywhere now with RMS systems, it is expected that there is a default config file. One the users flash the RMS SD card, they are forced to change the default config file before proceeding. I also don't like the idea that when people install the code that they immediately get an error because they skipped the step of copying the example config file, that that's very easy to miss.
So could you please revert the config file rename? I will accept the pull then.

mrprompt:
Sure @dvida, no problem.

dvida:
Looks good! I see you added the port 80 in the tunneling script, can you please confirm that it will not interfere with TeamViewer, because it also uses that port?

mrprompt:
@dvida sorry, I removed the TeamViewer from my Raspi, I'll try reinstall and test it. I added the 80 por because the IMX291 camera can managed from this port, without other softwares like CMS.

dvida:
No worries. I just want to make sure that folks who use teamviewer don't lose the connection once they start the tunneling script ;)

mrprompt:
Sure, like [this link](https://community.teamviewer.com/t5/Knowledge-Base/Which-ports-are-used-by-TeamViewer/ta-p/4139) says, port 80 is the last port used by TV, priority is 443 and 5938, as udp.

mrprompt:
@dvida what's you think? It's ok to merge?

PR #24, opened 2019-10-31
Prefix live view window with station code

mrprompt:
Re-enables preview window (only for non ARM/RPi) and prefix it with station ID. It's necessary when we're running more than one instance of RMS on same machine.

### DIFF
--- a/RMS/StartCapture.py
+++ b/RMS/StartCapture.py
@@ -212,15 +212,23 @@ def runCapture(config, duration=None, video_file=None, nodetect=False, detect_en
     # Initialize buffered capture
     bc = BufferedCapture(sharedArray, startTime, sharedArray2, startTime2, config, video_file=video_file)
 
+    live_view_enable = config.live_view_enable
 
-    # # Initialize the live image viewer
-    # if config.live_view_enable:
+    # Check if platform is ARM/RPi
+    platform = os.uname()[4]
 
-    #     live_view = LiveViewer(window_name='Maxpixel')
+    if platform.startswith('arm'):
+        live_view_enable = False
 
-    # else:
-    
-    live_view = None
+    # Initialize the live image viewer
+    if live_view_enable:
+        window_name = str(config.stationID) + '_Maxpixel'
+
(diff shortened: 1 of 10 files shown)


RMS/StartCapture.py
```diff
@@ -212,15 +212,23 @@ def runCapture(config, duration=None, video_file=None, nodetect=False, detect_en
     # Initialize buffered capture
     bc = BufferedCapture(sharedArray, startTime, sharedArray2, startTime2, config, video_file=video_file)
 
+    live_view_enable = config.live_view_enable
 
-    # # Initialize the live image viewer
-    # if config.live_view_enable:
+    # Check if platform is ARM/RPi
+    platform = os.uname()[4]
```
adalava, line 218: os.uname() doesn't work on Windows. You could use platform.uname() instead.

mrprompt: For now, the capture doesn't works on Windows platform, maybe it's not a problem for now because @dvida is rewriting this module (I think).

mrprompt: I'll update code and test soon.

mrprompt: Python2.7 (actualy used on RMS) haven't the machine property on result of platform.uname(), on 3.7 have. So, I think we can't use it now.

adalava: err sorry :(
actually, this is the portable way for the value you want: `platform.machine()`

dvida:
Hi,
I tested the live view on a Linux PC and it also keeps crashing. Thus, I cannot accept this pull request until this is resolved.
If it works on your PC, could you tell me the specs, OS, and the version of Python you are using?

mrprompt:
That's sucks @dvida  :(
The environment tested is Linux 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u1 (2019-09-20) x86_64. 
It's a HP Proliant ML350 G4 Server using some PCI Cards. Also, I tested this code on my RPis too (BR 4, 5, 7 and 8) and it's worked as expected. I'll make more tests, thank you.

PR #26, opened 2020-04-24
Reprocess on crash

adalava:
This changes RMS start process: if it's not capture time, look for partially-processed folders then call Reprocess code for all of them, except if FTPdetect file is found in a folder with same name on archived directory.  The results are added to the upload queue.

Please note the FTPdetect check was not fully tested yet!


dvida:
Thanks, I merged it! I'll deploy it for testing on a few systems.

PR #32, opened 2020-07-06
Camera control

markmac99:
Two sets of changes proposed for inclusion in the Dev branch:
1) Enhancements to CameraControl to allow the reading, saving, reloading and setting of camera configuration.
2) Enhancement to create MP4s of individual events as well as a slight change to the all-night timelapse.

There are also cosmetic changes to Readme.MD to bring it into line with the master branch, and I added the version of imageio in requirements.txt to maintain Python 2 compatability

** Camera Control **
Utils/CameraControl
.gitignore
Guides/imx291-config-maps.md
-------------------
Updated CameraControl considerably to use python-dvr to communicate with the camera. See lengthy inline documentation.
Updated gitignore to exclude camerasettings/*.json to avoid saving user-specific data in the Repo. Could later on remove this if we can store a good set of defaults. 
Added a file containing maps showing the location of various config params in the IMX291 cameras, as returned by dvrip

python-dvr
.gitmodules 
----------
Submodule added to support CameraControl. this module is not available via pip, hence submodule required for now.

**MP4 Generation **
Utils/FFtoFrames 
----------------
Can now be called from other python code.
Added ability to extract specific frame range. Defaults to all frames. 
(used by CreateMP4s to create movies of individual meteors)

Utils/BatchFFtoImage
Utils/GenerateTimelapse
-----------------------
Can now be called from other python code.

Utils/GenerateMP4s - NEW FILE
-----------------------------
Create short MP4 from each captured event. 
Extracts up to 30 frames before and after each event, to provide some context. 
Does not currently handle events that straddle two blocks of frames. 

build_with_anaconda.txt
-----------------------
Updated instructions to match current anaconda requirements

requirements.txt
----------------
Added version to imageio to ensure python 2.7 compatability

Readme.MD
---------
Synchronised with CMN/Master


adalava:
Hi Mark, first of all, this is a great improvement! I'm personally looking for this for a while. 
I'm going to test the new CameraControl, but it seems that other unrelated changes came along (68 commits). Would be possible to split them to another pull request? I think it would be easier to review and test each subject isolated.

markmac99:
largely this is my development style i think ! I commit a change to my 
fork each time i have satisfactorily tested it because bitter experience 
has taught me not to keep uncommitted code on my laptop! I also 
periodically resync my fork with the CMN master which creates additional 
commits.

There are actually only 11 files that have been altered and you can see 
the code differences on github. In the release notes i describe the two 
changes I've proposed, firstly the cameracontrol module (four files 
changed/added), secondly a module to create MP4s from each capture (one 
file added, three changed). In support of the latter I also updated some 
of the utility modules, mainly to allow them to be called from other 
modules. The other three changed files cosmetic changes made in Master 
by denis that i pulled back into my fork.

Is it essential that I split this into two pulls? Its quite a lot of 
work to rollback and then reapply in two separate branches, and I won't 
be able to do it this week due to the Perseids. I'm also on holiday for 
the last week in August and I'm still recovering from the virus so its 
possible i may not have time till september.

adalava:
Hi Mark, I hope you get better soon. There's no problem, it's actually up to Dennis, I only suggested that as in my experience smaller code changes and single subject makes easier for others to review and test.
I was able to cherry-pick the CameraControl related commits and integrated it in my personal fork (https://github.com/CroatianMeteorNetwork/RMS/compare/dev...adalava:CameraControl), and I'm using them right now.

Your code works pretty fine here, nice job! I tested several functions, including the save/load of all settings. One thing I noticed is it can't save/restore brightness, contrast, sharpness and saturation, but I think this is a limitation of this protocol. Camera's web interface has this limitation too, so I use ONVIF Manager for this.

markmac99:
" it can't save/restore brightness, contrast, sharpness and saturation"

Do you mean the sliders that are accessible via the "Colour" menu in 
CMS? Yes, I could not work out how to access those. I worked out most of 
the fields by trial and error!  When I have more time i will investigate 
further.  I may be able to find the right fields eventually!

This week is extremely busy the UK Meteor network (which i'm part of) 
are running outreach events for the Perseids, and we are live-uploading 
detections to our website and fireball videos to youtube. So i was up 
all night fixing problems....

dvida:
Hi Mark, thanks for the pull request, look great! Sorry for not merging it sooner, I was busy fixing up some other pull request and did not want to introduce more chaos to the repo.
One thing before I merge it - because the submodule was added, a normal git clone won't work anymore, it has to be done with the recursive flag it believe. Do you think it might be possible to pull the submodule inside setup.py?

markmac99:
Hi Denis
sure, will take a look at this. :) You can actually still use git clone by the way, but afterwards you need to cd into RMS and do "git submodule update --init" before you can use the cameracontrol module.  It all seemed to compile ok on the Pi by the way.

ps sorry i didn't reply sooner, I've had to build a new PC as my laptop started to freeze every few hours, losing all my work. now working on a shiny new machine with 48GB memory and 2TB storage...

dvida:
Thanks Mark, glad to hear you have a new machine!
Yeah, I know an additional command is needed (it can even be done within git clone), but I want to make sure that the new feature will work on existing installation, so we don't have to manually fix up 200+ stations...

markmac99:
ok, done - added submodule initialisation into setup.py, tested on my spare pi seems ok. 

Also synced requirements.txt with master (configreader==4.0.2 to prevent error in build)

PR #36, opened 2020-09-19
Camera control for Python 2.7 

markmac99:
The onvif camera control module requires a set of WSDL files (i think this stands for web sevice definition language?). The location of these is platform-specific and depends on which version of onvif was installed. I added functionality to determine the location of the WSDL files dynamically.

Tested on
-Windows10 with Anaconda
- Pi3 with Jessie
- Pi4 with Buster
I added support for MacOS but i am unable to test it as i do not have a Mac. 
Other Linux deployments should behave like the Pi. 

note: i haven't changed FOVKML.py, my repo is identical to cmn/dev but github thinks otherwise.


Utils/CameraControl27.py
```diff
@@ -1,8 +1,14 @@
 """ Controls ONVIF-compatible IP camera """
 
+# Prerequisites:
+# Python2 
+#   pip install onvif
```
mrprompt, line 5: Maybe it's a cool idea add these packages into `setup.py` and `requirements.txt` files.

markmac99: Actually, its there already. It was added some time ago. I just clarified the comments in the module header. This is from lines 30 onwards in Setup.py.
    ### HANDLE DIFFERENT ONVIF LIBRARIES FOR Py 2 AND 3 ###

    # Python 2 uses the 'onvif' library and Python 3 uses the onvif_zeep library
if sys.version_info[0] < 3:
    onvif_str = 'onvif_zeep'
    onvif_proper = 'onvif'
else:
    onvif_str = 'onvif'
    onvif_proper = 'onvif_zeep'
...
if onvif_str in reqs_stripped:
    onvif_index = reqs_stripped.index(onvif_str)

    # Replace the onvif module with the correct module for this version
    requirements[onvif_index] = onvif_proper

dvida:
Great job, thanks!

PR #38, opened 2020-11-09
Migrated some tkinter to qt

tammojan:
While waiting for my camera to arrive, I'm getting some code to work on my computer.

Note that the file dialogs still need to be ported to Qt.

I have not tested this on python2.

### DIFF
--- a/Utils/SkyFit2.py
+++ b/Utils/SkyFit2.py
@@ -13,13 +13,6 @@ import numpy as np
 import matplotlib
 import matplotlib.pyplot as plt
 
-try:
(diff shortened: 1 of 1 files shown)


tammojan:
I migrated some more functionality in skyfit2 from tkinter to pyqt. At least on my computer mixing the two simply crashes the program. I have not migrated the error messageboxes in FrameInterface.py, so they still crash for me. Otherwise I think SkyFit2 is now very usable!

(I accidentally opened another PR for this second fix, I closed that PR myself.)

PR #44, opened 2020-12-26
Convert height from EGM96 to WGS84 for CAMS

tammojan:
Currently in the CAL-files for CAMS, the height from the RMS config file is written as if it were the WGS84-height. From the documentation in the config file, the RMS height is w.r.t. the EGM96 datum. This code converts between the two (an alternative fix would be to add a WGS84 height in the config).

This introduces a dependency on pyproj, but since it's only for full CAMS compatibility it may not be necessary to add to the requirements.

### DIFF
--- a/RMS/Formats/CAL.py
+++ b/RMS/Formats/CAL.py
@@ -6,6 +6,7 @@ from __future__ import print_function, division, absolute_import
 import os
 import copy
 import math
+import pyproj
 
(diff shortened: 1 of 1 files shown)


tammojan:
Though proj is very standard, installing it requires two steps:
```
apt install proj-bin
pip install proj
```

PR #47, opened 2021-01-31
Cameracontrol enhancements

markmac99:
Small changes to Utils.CameraControl to 
- allow the setting of IP address, netmask, gateway and to toggle DHCP on and off. 
- call cameracontrol27 /onvif if running under python 2
- be able to to call from other modules, and to pass a config object rather than having to work out hte IP address. 

Tested ok under windows, being checked on pi3/python27 and pi4/python3.7

Note: changes to startcapture.py are not real, suspect unix/windows line-ending issue.


markmac99:
found an issue, will reopen when fixed

PR #52, opened 2021-02-07
Tweaks to allow movie as source instead of rstp stream

satmonkey:


satmonkey:
Tweaks to allow movie as source instead of rstp stream, fixes added

markmac99:
Hi milan, what is the naming convention for the video file?

satmonkey:
Hi Mark,
the video is expected in the same format as the FITS file, e.g. CZ0002_20210317_193338_404889.mp4
To generate at least single FITS file, it needs to be ca > 12s long. Try e.g. this:

https://user-images.githubusercontent.com/5328519/111656826-2c0d0280-880b-11eb-983f-8c9e66ce3ce7.mp4

PR #56, opened 2021-03-05
[WIP] Use python-dvr directly

apevec:
Drop git module and pin python-dvr to the latest commit.
After https://github.com/NeiroNx/python-dvr/pull/24 was merged,
python-dvr can be pip installed directly from git.

Note that git refs do not work for setuptools install_requires,
so pip install -r requirements.txt is required to be executed before
python setup.py install !

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "python-dvr"]
-	path = python-dvr
-	url = https://github.com/NeiroNx/python-dvr.git
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # RPi Meteor Station
 
-Open source powered meteor station. We are currently using the Raspberry Pi 3 (and 4!) as the main development platform and use digital IP cameras. **The code also works on Linux PCs, and everything but the detection works under Windows.** We are slowly phasing out the support for analog cameras, but they should work well regardless.
+Open source powered meteor station. We are currently using the Raspberry Pi 4 as the main development platform, and we use digital IP cameras. **The code also works on Linux PCs, and everything but the detection works under Windows.** We are slowly phasing out the support for analog cameras, but they should work well regardless.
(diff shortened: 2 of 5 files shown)


apevec:
Do not merge until confirmed it does not break python 2.7.
Ideally, we would migrate all RPi3s to latest Raspbian with python3 but not sure how to deploy that without breaking all RPi3s in production.

markmac99:
As soon as I have a chance i will take a look at this on my Pi3 with Jessie.

apevec:
I was going to say that Jessie's system dbpkg installed python-setuptool 5.5.1 is too old[*] but then realized in virtualenv latest setuptools is always installed!
In vRMS on my Pi3 I have new-enough 38.2.5 and in fresh venv on Pi3 I got 44.1.1 (latest py2.7 compatible setuptools)

[*] for https://www.python.org/dev/peps/pep-0508/#environment-markers we need at least 18.5 from Nov 2015

markmac99:
On my Pi3 with Python 2.7 i had to make an unrelated change due to an update in Paramiko which is not backwards compatible with python 2. In requirements.txt i added:
paramiko==2.4.0; python_version=='2.7'
paramiko>=2.4.2; python_version>='3.6'

I then made the same changes as apevec in requirements and setup, ran both and tested. No issues (once paramiko rolled back to 2.4.0)

[some-testing.txt](https://github.com/CroatianMeteorNetwork/RMS/files/6217599/some-testing.txt)
[pip-output.txt](https://github.com/CroatianMeteorNetwork/RMS/files/6217600/pip-output.txt)
[restart-rms.txt](https://github.com/CroatianMeteorNetwork/RMS/files/6217601/restart-rms.txt)
[setup-py-output.txt](https://github.com/CroatianMeteorNetwork/RMS/files/6217602/setup-py-output.txt)

PR #60, opened 2021-03-11
Fixed stupid bug in if statement - as reported by Thiago

markmac99:


dvida:
I already fix this in the master branch ;)

PR #63, opened 2021-03-26
Update timestamp in GenerateMP4s

tammojan:
This PR will make the time indicator in the generated MP4 reflect the actual time, not the time from the FF file from which it was extracted.


markmac99:
Looks good to me !

PR #69, opened 2021-04-29
Add flag to force cv2 to use V4L2

tammojan:
This may increase performance for direct video devices (/dev/video0).


dvida:
Thanks! Could you just switch the branch to dev, and add the "force_v4l2" option in the config file, set to false?

tammojan:
I think I cannot change the branch, and you can. If not, I can close the PR and open a new one.

PR #71, opened 2021-05-01
Optionally, add title to jpegs

markmac99:
Add an optional parameter to saveImage to allow a title to be added at the bottom left corner. 
Add optional argument to BatchFFtoImage to invoke this new code. 
The title is of the form "XX0001 YYYY-MM-DD HH:MM:SS UTC"

Tested on UK0006 and UK000F for a couple of weeks without issue. 

Note that saveImage is called from Compression, Reprocess, FRBinMosaic, MakeFlat, ManualReduction, SkyFit2, StackFFs, StackImgs.  No problems were observed with these modules which have not been altered.

### DIFF
--- a/RMS/Routines/Image.py
+++ b/RMS/Routines/Image.py
@@ -51,18 +51,18 @@ def loadImage(img_path, flatten=-1):
     return img
 
 
-def saveImage(img_path, img, addtitle=False):
+def saveImage(img_path, img, add_timestamp=False):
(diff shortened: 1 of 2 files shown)


dvida:
Thanks Mark, this is great - could you just change the name of the argument from "addtitle" to sometime like "add_timestamp"? It's more descriptive.

markmac99:
Done !

PR #75, opened 2021-05-20
escaped newline, fix parsing of old SkyFit

tammojan:


markmac99:
confirmed, I was just about to raise a defect on this :)

Issue #74, opened 2021-05-19
Fix limited dynamic range in 8bit FF frames


cbassa:
The FF frames are stored as 8bit FITS extensions. For the MAXPIXEL, MAXFRAME and AVEPIXEL extensions this is not an issue, but for the STDPIXEL frame the use of 8bits reduces the sensitivity, as the majority of STDPIXEL values are 1, with long tail to values of about 40 (out of 255). As a result, computing significance through (MAXPIXEL - AVEPIXEL) / STDPIXEL adds noise due to the lack of dynamic range in the STDPIXEL frame.

The FITS standard offers the BSCALE and BZERO keywords to apply a scaling and offset factor to scale the 8bit data. In `astropy`, the presence of these keywords in the extension header will be applied when reading the data of the extension, converting the 8bit data to floating point on the fly. For more information, see https://docs.astropy.org/en/stable/io/fits/usage/image.html.

For example for RMS, you could already get a factor of 10 more dynamic range by setting BSCALE to 0.1 in the STDPIXEL frame.

PR #77, opened 2021-05-21
implemented using bscale for storing std

tammojan:
This stores the standard deviation as uint8 with the standard FITS keyword `BSCALE=0.1`, which means all stored values will be scaled down by a factor 10 by all FITS-compliant software (including Astropy). This means that the values for the standard deviation with this change are limited to the range 0 to 25.5, and have one digit of precision.

I have tested this on my camera NL000D, so far everything works as expected, but I'll report back when I've had a clear night.

A side effect of this is that when opening FITS files with astropy and writing them back (something that RMS and CMN_binviewer do not) the values will be stored as floating point numbers, increasing the file size. This can be avoided by opening them with `astropy.io.fits.open(FFbla.fits, scale=True)` (note that the scaling always happens, this refers only to the writing behavior).

As suggested by @cbassa in #74

### DIFF
--- a/RMS/CompressionCy.pyx
+++ b/RMS/CompressionCy.pyx
@@ -157,8 +157,9 @@ def compressFrames(np.ndarray[INT8_TYPE_t, ndim=3] frames, int deinterlace_order
             # Subtract average squared sum of all values (acc*mean = acc*acc/frames_num_minus_four)
             var -= acc*mean    
 
-            # Compute the standard deviation
-            var = <unsigned int> sqrt(var/frames_num_minus_five)
+            # Compute 10 times the standard deviation (to keep one decimal when rounding to uint8)
+            # In FFbin and FFfits, the standard deviation is divided by 10 again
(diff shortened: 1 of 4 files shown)


dvida:
This needs to be conditional on whether the fits format is used or not. RMS can produce the CAMS-format bin files, and this won't work there.
Also, I don't think this is really an issue at all. Even if the stddev is only 1, the meteor still needs to be 4 levels above the background, which is really really faint.

tammojan:
I can make it conditional on using FITS I think.

I think the issue will occur more when the stddev is less than 0.5, which will be rounded to zero and make those pixels unusable. @cbassa mentioned that he ran into this when trying to detect satellites, maybe he can expand a bit.

dvida:
The line below the stddev computation limits it to a minimum value of 1, to avoid zero division errors afterwards: https://github.com/CroatianMeteorNetwork/RMS/blob/master/RMS/CompressionCy.pyx#L163
This effectively takes care of those issues.

PR #78, opened 2021-05-31
Fix timestamps in GenerateMP4s

tammojan:
The timestamps in GenerateMP4s (which since #63 update every second) still refer to the beginning of the FF-file. This fixes that bug.

I think @markmac99 [used](https://github.com/CroatianMeteorNetwork/RMS/commit/bd8f2e0cc5f7b35ed909c33aaa700fe2e0e931ff#diff-10fc682193c6d4db0d6b9e90e011cd3f37e749215b3b2838aada1af737e5132dL163) `i-first_frame` in the call to `saveFrame` to make the output JPGs start counting at 0 to make the input to ffmpeg easier. There are two problems with this:
- It's inconsistent with the interlaced variants
- It messes up the timestamps

Since both ffmpeg and avconv support specifying the number of the first frame, it's also not necessary.

### DIFF
--- a/Utils/FFtoFrames.py
+++ b/Utils/FFtoFrames.py
@@ -159,7 +159,7 @@ def FFtoFrames(file_path, out_dir, file_format, deinterlace_mode, first_frame=0,
 
         # No deinterlace
         else:
-            frame_name, frame_dt = saveFrame(frame, i-first_frame, out_dir, file_name_saving, file_format, ff_dt, fps)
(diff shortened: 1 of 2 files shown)


dvida:
Waiting for @markmac99 to confirm...

markmac99:
Looks good.  I wasn't aware of this flag.  

Tested on Windows 10 / Python 3.x and Pi3 / Jessie / Python 2.7 without issues.

tammojan:
Thanks for the blazing fast review + merge!

PR #81, opened 2021-06-26
Support UYVY pixelformat

jenyak:
This PR allows the use of cameras that provide images in the UYVY pixel format. For an example [YEVE IMX 327](http://wiki.veye.cc/index.php/VEYE-MIPI-290/327) provide image only in UYVY format.
With config like this 
`device: v4l2src device=/dev/video0 ! video/x-raw,format=UYVY,width=1920,height=1080,framerate=30/1 ! appsink sync=1`
openCV receive frames in UYVY format. This format have 2 channels and only second channel (Y or luma) is important.

### DIFF
--- a/.config
+++ b/.config
@@ -43,6 +43,7 @@ force_v4l2: false
 width: 1280
 height: 720
 fps: 25.0 ; frames per second
+uyvy_pixelformat: false
 report_dropped_frames: false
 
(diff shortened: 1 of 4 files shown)


dvida:
Thank you - due to some conflicts, I merged this manually to the dev branch.

PR #97, opened 2021-08-21
Bugfix in setcameraddress

markmac99:
Foolish error in setcameraaddress - it doesn't allow for IPaddresses either of whose middle two digits could be zero...


dvida:
I made a few small tweaks, can you just test if it works now as intended?

markmac99:
Ah, no not quite because 0 is a valid possibility for the 2nd and 3rd digits. This was specifically the bug i was fixing :) 
I've proposed a fix which i think does work

dvida:
Oooh, right!

PR #113, opened 2021-12-07
Save star and meteor count to a file

tammojan:


dvida:
Thanks, this is a great idea! Nevertheless, all information is already present in the CALSTARS file, so I don't see any reason for duplication.
Could you modify the script that plots fieldsums to also optionally take the CALSTARS file and plots the number of stars on a new axis?

tammojan:
Thanks, I didn't think of looking at the CALSTARS first... The plot is in the [mail](https://groups.io/g/globalmeteornetwork/message/5037).

PR #114, opened 2021-12-09
Do not detect mac as windows

tammojan:


dvida:
I believe the logic operators should be in this order: "and "darwin" not in"
Not should be applied on "in" and not on "darwin".

tammojan:
> I believe the logic operators should be in this order: "and "darwin" not in" Not should be applied on "in" and not on "darwin".

Ah yes, that's indeed better style. Fixed.

PR #116, opened 2022-02-03
Flux

dvida:


markmac99:
Hi i just noticed that this commit changed the behaviour of ShowerAssociation when passed a list or folder rather than a single FTP file. 
Previously the module was able to recieve multiple FTP files on the commandline, or to be passed a folder containing multiple FTP files, and then to plot them all on a single graph. 
Now however, the function only plots the first ftp file in the folder, and errors out if multiple files are passed. This is because a new function findFTPDetectFile() was introduced, which loads only the first FTP file in the folder, and generates an error if multiple files are passed in. 

For me, this new behaviour is undesirable! Previously i was able to plot multiple (colocated) cameras or multiple nights on a single plot, thus showing the total flux of meteors over say an entire shower. 

Is there a specific reason for changing the behaviour? If not, can i please revert it ?

dvida:
Sorry about that, I added the multi-FTPdetectinfo file functionality back in.

markmac99:
Thanks! I'm probably the only person who uses that functionality but its handy!

PR #120, opened 2022-02-24
[RFC] Optional tmux session in StartCapture

apevec:
tmux[1] allows access to the running session from another local shell:
(vRMS) pi@hr0017:~/source/RMS $ tmux ls
RMS: 2 windows (created Sun Feb 20 21:46:55 2022) [80x23]

(vRMS) pi@hr0017:~/source/RMS $ tmux attach -t RMS

[1] tmux is a terminal multiplexer. https://github.com/tmux/tmux/wiki

### DIFF
--- a/Scripts/RMS_StartCapture.sh
+++ b/Scripts/RMS_StartCapture.sh
@@ -15,7 +15,12 @@ mkdir -p $LOGPATH
 # Log the output to a file (warning: this breaks Ctrl+C passing to StartCapture)
 #python -m RMS.StartCapture 2>&1 | tee $LOGFILE
 
-python -m RMS.StartCapture "$@"
+# start a new tmux session if not already inside one and tmux is available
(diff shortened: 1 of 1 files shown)


apevec:
Proposing this as a draft to check if folks think this would be useful.

End goal is to have a completely headless setup, with all the customizations done during image burning via rpi-imager, without any need for manual first setup via GUI.

dario-zubovic:
IMO this makes perfect sense, especially for people who want to deploy all-sky-like stations with arrays of multiple cameras, makes the setup easier.

Early on I used to run RMS with screen instead of tmux though :)

I imagine it would be better to allow use of tmux depending on `--tmux` argument, instead of always using it when available?

apevec:
I'll close this and start looking instead at containerization https://github.com/lancer73/rms_container

PR #124, opened 2022-04-15
Fix ping on windows

markmac99:
subprocess.call expects each parameter to be in a separate list element so the -c  or -1 and 1 must be separated.


markmac99:
cancelling as adalava already submitted a fix that is pending application to Master

PR #126, opened 2022-05-20
add RMS_Backup.sh script

adalava:
It creates a .zip file containing the important config files that
should be preserved for repairing or replacing a RMS station.


adalava:
Tried to use **dev** branch but it's apparently not in sync with **master**

PR #129, opened 2022-07-27
Filter FTPdetectinfo for meteors only

satmonkey:


satmonkey:
This is a base code for filtering meteors by ML.
It works as a module and calling as a function.
The argument is a path to the FTPDetectinfo file.
There are 2 important parameters: 
prediction threshold and a model path.
For the time being hardcoded in the file, probably candidates for including them into .config.

PR #131, opened 2022-08-01
Added support for making PNG retention optional, and rerunning on the same folder. 

markmac99:
- optionally keeping the pngs:  -p 0 will delete them. Default / -p 1 will keep them. 
- rerunning with different threshold and  saving the modified ftpfiles with unique names for later review

When rerunning the module, if the _unfiltered file exists, then this is used in place of the FTPdetect file passed in on the commandline. This allows one to run and rerun the script with different thresholds more easily, and without having to think about whether the module was already run on the folder. 
The resulting filtered ftpdetect file is also copied to a 2nd file with the threshold in its name so you can compare the output of different runs. 
I also tidied up a couple of unused variables and some non-standard indentation which would have caused some versions of Python to error out.


markmac99:
@satmonkey  @dvida FYI

PR #132, opened 2022-08-05
Prefix live.jpg to station ID

mrprompt:
Added the station ID as prefix of live.jpg file, because it's conflicting when running more than one station into single machine.


markmac99:
I agree this is useful and consistent with most other RMS files. However there are many users who are sending the live file to to a website or sharing in other ways. Changing its name would break their processes. Since most / the majority will not be running multiple cameras on one machine, can i suggest that prefixing the stationid should be an option in the config file?

dvida:
Agreed, that's why I haven't merged this one yet. We could make both though, or simply make a symlink to live.jpg?

PR #133, opened 2022-08-06
Fix for remote reprocessing

satmonkey:
Archive and platepar path should correspond to Captured path. This is the fix for remote reprocessing e.g. via SMB or NFS


dvida:
Could you possibly find another way to construct a path? The path "source/RMS" should not be hardcoded anywhere. To support running multiple stations on one machine, the working directory should always be the reference one where all the operational files are (config, mask, platepar).

PR #134, opened 2022-08-12
Enhancing TrackStack

markmac99:
Largely work done by @tammojan but with some work by @markmac99 to handle error conditions and simplify the code:

Adding the ability to create a trackstack filtered by shower eg -s PER, and to create a trackstack for multiple days, and to set the output folder 
python -m Utils.TrackStack dir1 dir2 dir3 -s showerlist -o somewhere

dir_path can be a list of folders for example different nights of a shower
-s showers - optional, comma-separated list of showers e g PER,cap,ARI (case insensitive)
-o somewhere - optional, save the output in a different location to avoid overwriting an existing trackstack
-d optional, set blackpoint to minimum detected value (darkens sky considerably, may be undesirable for some users)

Note: shower-filtration requires the FTPdetect files and a valid .config file (default read from RMS folder). 
Note: tqdm is not normally installed on the pi though its deployed by anaconda on other platforms. Rather than requiring a new module, i put some logic in to use the old progress indicator if tqdm was not available. 

Allows the user to create a trackstack of multiple nights for a shower, as shown in the attached Perseid extract for the nights 8-11 Aug. 
UK002F_20220808_213003_871679_track_stack

### DIFF
--- a/Utils/TrackStack.py
+++ b/Utils/TrackStack.py
@@ -3,21 +3,33 @@
 import os
 import json
 import copy
+from glob import glob
+import argparse
+
 
 import cv2
 import numpy as np
 import matplotlib.pyplot as plt
 
+try:
+    from tqdm import tqdm
+    got_tqdm = True
+except:
+    got_tqdm = False
+
+import RMS.ConfigReader as cr
 from RMS.Astrometry.ApplyAstrometry import xyToRaDecPP, raDecToXYPP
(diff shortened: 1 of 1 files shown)


markmac99:
Tested on 
Windows 10 / Python 3.7.9 
Pi4 / Python 3.7.3
Pi3 / Python 2.7.9

PR #143, opened 2022-09-27
Fix running RMSExternalScript manually on machines with multiple stations

adalava:
Adds --config/-c parameter that can be used to provide a different RMS configuration file instead of using the default one. This is
useful when there are multiple stations running on the same machine.

Also fixes an error handling when the provided config file is not found.

Reported by: William Schauff and  Washington Oliveira

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -182,10 +182,9 @@ def loadConfigFromDirectory(cml_args_config, dir_path):
             # Load the config file from the full path
             config_file = os.path.abspath(cml_args_config[0].replace('"', ''))
 
-
-        if config_file is None:
+        if config_file is None or os.path.exists(config_file) is False:
             raise FileNotFoundError("A config file could not be found in directory: {:s}, {:s}".format(
-                dir_path, cml_args_config
+                dir_path, cml_args_config[0]
                 )
             )
 
--- a/RMS/RunExternalScript.py
+++ b/RMS/RunExternalScript.py
@@ -78,11 +78,15 @@ if __name__ == "__main__":
     arg_parser = argparse.ArgumentParser(description=""" Run external script.
         """)
 
-    arg_parser.add_argument('captured_path', nargs=1, metavar='CAPTURED_PATH', type=str, \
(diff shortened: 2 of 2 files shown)


adalava:
@dvida , I'm not sure about the default behavior, I would like to help user avoid mistakes when running it manually. When a config file is not specified, I would like to suggest one of the solutions bellow (or one you and other may come up with):
a) always load the .config file from Captured dir instead of the one in source/RMS/.config;
b) when user doesn't specify a config file, check if the file in Capture dir is exactly the same as the one in source/RMS/.config before loading it. If not, interrupt the script and warn the user that there's a mismatch and suggest force a config file with --config/-c;

I'd vote option b :)

markmac99 (changes requested):
I don't think the try/except block is needed in RunExternalScript, since ConfigReader already handles missing files and will actually exit(). 
The current behaviour of RunExternalScript is to load the config file from ~/source/RMS. I think this should not be changed to as otherwise it could cause backward compatability issues for anyone who is using the module already.

dvida:
I agree with Mark's comments, the default behavior should be preserved. I.e. if a config file is not specified, use the one in ~/source/RMS.

adalava:
There's no intention of changing the default behavior on this change: when --config parameter is not used (default case), a 'None' object is passed to loadConfigFromDirectory(), then ConfigReader will use the already existing logic[1]  to load the default ".config" file (This is similar to the original config file load code in RMSExternalScript)

The try/except block is needed because when a config file is provided and the file is not found it raises a FileNotFoundError exception that need to be handled, otherwise user will get the unhandled exception stack trace instead of a friendly error message.

The original idea was to always load the config file from "data" directory following the '.' behavior in [2]. I think it would be ideal but it would change default behavior for sure.

[1] https://github.com/CroatianMeteorNetwork/RMS/blob/c3b5b64139fb22b10c81a6ce087a98418df7e01a/RMS/ConfigReader.py#L198  (lines 159 and 198)
[2] https://github.com/CroatianMeteorNetwork/RMS/blob/c3b5b64139fb22b10c81a6ce087a98418df7e01a/RMS/ConfigReader.py#L165

markmac99:
closing as superceded and very old vs the current codebase

PR #145, opened 2022-10-05
PNG suffix correction

satmonkey:
Is the 'pred=' really needed in the file name? The PNG file name gets quite large now, and the '=' might have issues when using windows.


dvida:
I changed it to "p-" instead of "pred=", is that OK? I still want it to be indicated so that casual users know that's the prediction numbers.

adalava:
Hum... nowadays I prefer long names so it can speak for itself. The meaning of 'p' would be certainly one question people will ask on our local group. (actually even 'pred' may be not enough, for instance some people get confused about the difference between "_detected" and "_captured" stacks)

satmonkey:
that seems to be better, yes. The actual reason of this PR is that in my favorite tool Total Commander, the file name is truncated just after the timestamp...

PR #155, opened 2023-02-12
Add DrawConstellations.py

tammojan:
This adds a command-line util to draw constellations with the projection of one particular FF file. Overlaying on an image of the FF file is left to other tools (I use photoshop or GIMP for this).
There are a few todo's in the code.


g7gpr:
Is the plan to integrate tightly with TrackStack, so adding constellation lines becomes a command line argument?

g7gpr:
I'm having problems when the images cover a big angle of the sky, I think it is the default 90 degrees angle that is causing it.

PR #157, opened 2023-02-14
Use computed field of view for constellations

tammojan:
This should improve the drawing of constellations of wide-field trackstacks. Previously, a field of view of 90 degrees was hardcoded to prevent wrapping around the sphere.
I hope @g7gpr can test this.

### DIFF
--- a/Utils/DrawConstellations.py
+++ b/Utils/DrawConstellations.py
@@ -29,7 +29,7 @@ def drawConstellations(platepar, ff_file, separation_deg=90, color_bgra=None):
     lines = np.loadtxt(constellations_path, delimiter=",")
     from_ra, from_dec = lines[:, 0], lines[:, 1]
     to_ra, to_dec = lines[:, 2], lines[:, 3]
-    from_x, from_y = raDecToXYPP(np.array(from_ra), from_dec, fftime_jd, platepar)
+    from_x, from_y = raDecToXYPP(np.array(from_ra), np.array(from_dec), fftime_jd, platepar)
     ang_sep = np.rad2deg(angularSeparation(np.deg2rad(platepar.RA_d), np.deg2rad(platepar.dec_d), np.deg2rad(from_ra), np.deg2rad(from_dec)))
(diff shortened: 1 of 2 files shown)


g7gpr:
Testing now.

g7gpr:
https://github.com/tammojan/RMS/blob/30406813e7c4b277cbf1eb584b51f06b9ffe0591/Utils/DrawConstellations.py#L32

Why on this line are you only converting from_ra into an array, and not from_dec?

tammojan:
> https://github.com/tammojan/RMS/blob/30406813e7c4b277cbf1eb584b51f06b9ffe0591/Utils/DrawConstellations.py#L32
> 
> Why on this line are you only converting from_ra into an array, and not from_dec?

For no particular reason, apparently it worked also without the conversion :) I've added a commit to convert both into numpy arrays (they were both lists).

g7gpr:
AU0013_20230212_113632_086730_track_stack_2

looking better - Orion has appeared! But something strange is happening to Phoenix? Not sure.

g7gpr:
AU0013_20230213_135445_991929_track_stack_compressed

Still something odd going on - Lupus stops at Beta Lupus. However I think I know what is happening. In TrackStack, ang_sep is from the centre of the image, in DrawConstellations, it is the width of the image - analogous to radius / diameter. Running again but with the ang_sep multipled by 1.95.

g7gpr:
Or rather, the middle platepar by time, might not be in the centre of the final projection.

g7gpr:
https://1drv.ms/u/s!AlG3aYd4Ra-vjPcBg2QWX7VE1WUfRQ

g7gpr:
Setting the angle to 175 degrees works pretty well.

g7gpr:
AU000W_20230225_111925_841994_track_stack

Looking very good now. Thanks for all the work on this tool.

g7gpr:
AU000K_20230225_112032_785216_track_stack
AU0013_20230225_112223_724900_track_stack

PR #162, opened 2023-03-22
Update GRMSUpdater.sh

edharman:
Missed leading delimeter to awk regex


edharman:
Hi Denis,
This code which performs an orderly RMS_Update couldn't be fully tested prior to release and this change fixes a bug I introduced when we factored in the codebase change
/Ed

PR #175, opened 2023-06-03
Eventmonitor

g7gpr:
Modifications to config reader to support the new functions
Addition of Event Monitor
Brought functions from WMPL into Math
Modifications to StartCapture to run and stop EventMonitor
Addition of EventMonitorTests

EvMonServerSide, which is running on my machine for testing purposes only
EvMonUI

python -m Utils.EvMonUI -i launches an interactive console app for generating watchlists.


RMS/StartCapture.py
g7gpr, line 1: Not sure why git has put this as a complete file replacement, working on it.

g7gpr:
Closing this, because I cannot fix the problem with StartCapture.

PR #176, opened 2023-06-04
Eventmonitor

g7gpr:
Resubmitted after fixing the LF/CRLF issue in StartCapture.py. Apologies.

Added changes to .config - template file has event_monitor_enabled: false
Made changes to ConfigReader to support EventMonitor
Added EventMonitor
Brought functions into Math.py from WMPL, using degrees as inputs rather than Radians
Changes to StartCapture, following template of upload_manager
Added Tests/EventMonitorTests
Added Utils/EvMonUI.py, which when run with -i can create an event based on a trajectory either entered by hand, or based on a download from the GMN database.


g7gpr:
AU0006,09,1P,1Q,1R,1S are on the event_monitor branch.

PR #178, opened 2023-06-14
Improvements to CameraControl

markmac99:
Add the following to CameraControl
- GetAutoReboot - display the current auto-reboot settings from the camera. 
- CloudConnection - enable or disable the "phone-home" feature that connects the camera to the XMEye network in China. This should always be disabled even if your camera is on a private network. 
- CameraTime get|set - set or get the internal time.  Call set with no params to set to current time, or use YYYYMMDD_HHMMSS
- tidy up the documentation in Guides/code (lots of missing newlines)

Also a small change to MLFilter to remove any existing ML results when reprocessing a folder eg after applying a new platepar. Currently, if the file exists then the ML is not rerun. So even if the data are recalibrated, the previous incorrect results are displayed.

i also made a change to the requirements.txt file to support Ubuntu and Debian, by sourcing a working version of tflite for Python > 3.9. There may be other workarounds.


Guides/imx291-config-maps.md
markmac99, line 1: I tidied up this file. Previously it was missing many spaces, and so the lines wereallconcatenatedintoahugemess... I added spaces and marked the tables as code blocks to retain alignment.

markmac99:
Updated as requested.

markmac99:
Final update to set the time and disable remote-control from the RMS_SetCameraParams.sh script

PR #182, opened 2023-07-14
Eventmonitor

g7gpr:
Corrected handling of elevations


dvida:
If I may suggest the elevation range - we should limit the elevation to a range between 10 and 90 deg. If the given elevation is between 0 and 10, set it to 10. If it's outside the 0 - 90 range then set it to 45 which is statistically the most probable entry angle.

dvida:
We should also probably print a warning indicating that the elevation has been changed, so it doesn't silently change the value.

g7gpr:
Agree with the those comments, but I have a more fundamental problem. The backward propagation to the maximum luminous flight height works perfectly, range at the az and el produces a height accurate to 6 significant figures. The forward propagation is not minimum is not working properly - 10% errors. I know the input data is likely to be inaccurate, but I'd like the algorithms to be sensible. I can fix this by translating to Cartesian, and extending a unit vector, but with a few hours I can come up with a nicer polar solution.

g7gpr:
Tried a few methods. Approximation is very close, added an iteration to improve accuracy.
Implemented checks on elevation.

g7gpr:
rewrite uncertainty handling to work in cartesian rather than polar coordinates

PR #187, opened 2023-07-25
Convert worker_timeout to int for logging

g7gpr:
When running processing after capture, worker_timeout is generally not an integer, as it includes the capture duration. This change converts it to an int for logging.

### DIFF
--- a/RMS/QueuedPool.py
+++ b/RMS/QueuedPool.py
@@ -434,7 +434,7 @@ class QueuedPool(object):
 
                 # If the queue has been idle for too long, kill it
                 if (time.time() - output_qsize_last_change) > worker_timeout:
-                    self.printAndLog('One of the workers got stuck longer than {:.1f} seconds, killing multiprocessing...'.format(worker_timeout))
+                    self.printAndLog('One of the workers got stuck longer than {:.1f} seconds, killing multiprocessing...'.format(float(worker_timeout)))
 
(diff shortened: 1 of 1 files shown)


dvida:
What if it can't be converted to an integer, which seems to be causing the issue in the first place? It will crash the program.

g7gpr:
Happy to write something safer if you can make a suggestion. 

This is what I tested.

Current code - crash

```
>>> print("Int conversion {:d}".format(3.157))

  File "<stdin>", line 1, in <module>
ValueError: Unknown format code 'd' for object of type 'float''

Option 1 wrap in int()

>>> print("Int conversion {:d}".format(int(3.157)))
Int conversion 3

Option 2 Get rid of format code

>>> print("Int conversion {}".format(3.157))
Int conversion 3.157

Option 3 Format to float

>>> print("To float {:.2f}".format(3.157))
To float 3.16
>>> 
```

g7gpr:
So would you suggest putting it in a try: except: ?

dvida:
Thanks! I might not be understanding the problem, but why don't we make the format float, and convert to float? If that is the type that gets passed in anyway.

g7gpr:
I'll test that approach.

dvida:
And just in case the timeout is integer, could you do float(worker_timeout) in the print? Thanks!

Issue #171, opened 2023-05-13
Add a check so that auto_reprocess does not run if should be capturing


g7gpr:
https://github.com/CroatianMeteorNetwork/RMS/blob/b52f21834e2a430b0350cac06e7b6580bea8ad22/RMS/StartCapture.py#L612

Add a check so that if it is time to start capturing auto_reprocess completes any directory which is in process, but does not start a new one.

g7gpr:
I can probably tackle this myself and submit a PR.

PR #192, opened 2023-08-16
Stop reprocessing if capture time

g7gpr:
Wrt to Issue #171, allows at least one directory to be reprocessed, but it it is time to capture does not allow any more to be processed, until the next day.

### DIFF
--- a/RMS/StartCapture.py
+++ b/RMS/StartCapture.py
@@ -670,6 +670,14 @@ def processIncompleteCaptures(config, upload_manager):
             log.error(repr(e))
             log.error(repr(traceback.format_exception(*sys.exc_info())))
 
+        # If capture should have started do not process any more incomplete directories
+        start_time, duration = captureDuration(config.latitude, config.longitude, config.elevation)
+        if isinstance(start_time, bool):
+            if start_time:
+                log.info("Stopping reprocess, capture due to start")
+                break
+
+
 
 
 if __name__ == "__main__":


dvida:
So I'm torn on this one because I can see this get stuck in an infinite loop of half-processed observations. I'd rather have fully processed nights ASAP than wait for them to be processed or perhaps even deleted by the disk management code.

g7gpr:
It's perhaps something I'll keep for my own stations. Up to you.

dvida:
You can add that as an option in the config file, with it being disabled by default.

g7gpr:
This can't get stuck in a loop of half processed observations. If an observation starts to get reprocessed it will not be interrupted, even if this means that capture start is delayed; the break occurs at the end of processing one night of data, not during the processing of a night.

dvida:
I guess it will always reprocess one, but not more than one?

g7gpr:
If reprocess is called, then it will always reprocess one unprocessed directory, as the conditional break comes after the function that does the reprocessing. If capture is not due to start, then it will start working on the next directory. So capture may end up starting later, but it should not reprocess all through the night.

One station got turned off at dawn each morning as it had been moved onto a circuit controlled by a photocell. When this problem was resolved, it did not capture anything for four days while it worked through all the broken directories. This is to protect against that situation.

dvida:
Okay, I think this sounds reasonable now.

PR #194, opened 2023-08-26
Timesync

g7gpr:
This fixes a problem when a brand new station cannot execute RMS_Update.sh successfully because the RTC is too far out. This has occurred every time at one location. It may be an issue which is unique, and hence not be a suitable candidate for a pull. However I have noticed other stations sometimes appear to start and end capture at strange times. It may also block capture if there is no internet connection.


g7gpr:
Not relevant for most stations, closing.

PR #203, opened 2023-09-12
Fix for loss of data when reprocessing

g7gpr:
Fix concurrency problem in queued pool when reprocessing tries to load whole night of observation into the pool


g7gpr:
I've only been able to reproduce this behavior on Intel Q9560; closing as this a rare CPU.

PR #204, opened 2023-09-12
Take spaces out of network and group names

g7gpr:


g7gpr:
User reported that spaces in camera_group_name or network_name caused FFMpeg to fail. Stripped spaces out of these when creating folder structure.

PR #205, opened 2023-09-12
Log each run of EventMonitor and write start time.

g7gpr:
Request for log entry to be made on each iteration of event monitor, added logging of start time so that this does not scroll off the screen.


g7gpr:
Seen working on Pi4 platform.

g7gpr:
Testing completed

PR #208, opened 2023-09-13
Block FS files

g7gpr:


g7gpr:
Prevent FS*.bin files being processed by binviewer.

Issue #213, opened 2023-09-25
Zero length .config file in CapturedDirectory causes EventMonitor crash


g7gpr:
Report of EventMonitor process crashing when trying to parse a zero length .config file. 

Capture was not affected, however pr will be submitted.

PR #214, opened 2023-09-25
Handle zero length .config file

g7gpr:
https://github.com/CroatianMeteorNetwork/RMS/issues/213

Report of a crash of EventMonitor following parsing a zero length .config file found in a CapturedFiles directory. This fix uses the station .config file in this event.

### DIFF
--- a/RMS/EventMonitor.py
+++ b/RMS/EventMonitor.py
@@ -1776,7 +1776,23 @@ class EventMonitor(multiprocessing.Process):
             # If there is a .config file then parse it as evcon - not the station config
             for file in file_list:
                 if file.endswith(".config"):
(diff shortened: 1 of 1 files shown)


dvida:
Would you mind if I merge it into the eventmonitor branch first for some testing?
It's probably not a necessary fix though, RMS won't run well with a zero-length config file anyway (or at least it shouldn't). If it does, then we need to put the catch somewhere else up the chain.

g7gpr:
Yes, good idea. Let me know if you need me to do anything.

dvida:
I changed the base but it has some conflicts, would you mind fixing them? You know best which lines should be modified. Thanks!

PR #217, opened 2023-09-27
Multi camera gui platepar

g7gpr:
More work on handling CapturedFile directories with missing .config and platepar files.
More robust determination of fallback directory for multi-camera stations
The additional commenting is to assist with debugging on multi-camera stations.


g7gpr:
This appears to have fixed the edge case. I'll build a multi camera gui station and test it on my own hardware before merging into master.

Issue #222, opened 2023-10-02
Video for created from FR bin can not be played in some OS by standard players


aitov:
FRbinViewer extracts images from FR bin file and for linux uses ffmpeg for mp4 video generation which cannot be played by standard players except VLC. (Reproduced on Intel Mac, Android)

Root cause:
 
When saving to MP4 using a typical H.264 encoder, ffmpeg will convert to YUV color encoding but pick the highest fidelity pixel format which, for a RGB source, is YUV 4:4:4. This format is not compatible with most players. Solution is to manually force YUV 4:2:0 output. Add -pix_fmt yuv420p just before the output filename.

As I see for time lapse  this parameter already used, propose apply it to FRbinviewer too.

Issue #201, opened 2023-09-11
FRbinViewer - add timestamp to video like in BatchFFtoImage


aitov:
Currently FRbinViewer has option "-t", "--timestamp" but it adds file name and frame number.
1. Move frame num option to another parameter
2. Add date and time label same as in BatchFFtoImage util
It could be fixed by code like this:
[bin-viewer-improvements](https://github.com/CroatianMeteorNetwork/RMS/compare/master...aitov:RMS:bin-viewer-improvements#diff-ccb61317f713fcd05c36e97d2a23cd104068cade95929266b0112de6cb2e5f03R186)

Issue #200, opened 2023-09-11
FRbinViewer - fit window to screen size when playing Full HD bin files


aitov:
If camera is recording in full hd resolution and try run FRbinViewer without -x option:  player's window doesn't fit screen size and some parts of video out of view.
It could be fixed by resizing image before show like this code:
[bin-viewer-improvements](https://github.com/CroatianMeteorNetwork/RMS/compare/master...aitov:RMS:bin-viewer-improvements#diff-ccb61317f713fcd05c36e97d2a23cd104068cade95929266b0112de6cb2e5f03R211)

PR #223, opened 2023-10-02
Bin viewer improvements

aitov:
fixed for:
 #222 Video for created from FR bin can not be played in some OS by standard players
#201 FRbinViewer - add timestamp to video like in BatchFFtoImage
#200  FRbinViewer - fit window to screen size when playing Full HD bin files


markmac99 (approved):
Looks good. The addition of pix_fmt to the ffmpeg call does fix the problem identiifed.

PR #230, opened 2023-11-02
Refactoring and database interfaces

g7gpr:


g7gpr:
Not ready for master yet, but is running on non-gui and gui linux boxes.

PR #242, opened 2023-12-23
DoNotShowCatalogStarsInMaskedRegionsOfAstrometryReport

MurrayForbes:
Filter out the catalog stars in the masked regions of the image of the Astrometry report.

### DIFF
--- a/RMS/Routines/MaskImage.py
+++ b/RMS/Routines/MaskImage.py
@@ -72,7 +72,7 @@ class MaskStructure(object):
 
 def getMaskFile(dir_path, config, file_list=None):
     """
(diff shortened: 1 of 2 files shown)


markmac99:
Could you clarify the problem that this fixes please? I am not sure i see why we need to make the change. 

A couple of other points: 
recalibrated is correctly spelled as one word, not hyphenated. 
also note that "dict" is normal usage for the python dictionary type
sorry if these seem small points but the changes don't add to the clarity or fix errors in the comments, and its likely the next maintainer will change them back, so i would suggest we leave the current spelling.

PR #248, opened 2024-01-04
Tested jump to next furthest star

g7gpr:
BackSlash key (above Enter) pans and zooms the view to the star which is furthest away from all the other matched stars. 

This gives an even spread of stars across the image and reduces the time spent scrolling.

Still to implement is jump to lowest and highest magnitude unmatched star.

### DIFF
--- a/Utils/SkyFit2.py
+++ b/Utils/SkyFit2.py
@@ -1340,6 +1340,7 @@ class PlateTool(QtWidgets.QMainWindow):
             text_str += 'A/D - Azimuth\n'
             text_str += 'S/W - Altitude\n'
             text_str += 'Q/E - Position angle\n'
+            text_str += (',/. - Next unmatched star\n')
             text_str += 'Up/Down - Scale\n'
             text_str += 'T - Toggle refraction correction\n'
 
@@ -2967,6 +2968,48 @@ class PlateTool(QtWidgets.QMainWindow):
                 self.platepar.updateRefRADec(preserve_rotation=True)
                 self.checkParamRange()
 
+                self.tab.param_manager.updatePlatepar()
+                self.updateLeftLabels()
+                self.updateStars()
+
+            # Jump to faintest unmatched star
+            elif event.key() == QtCore.Qt.Key_Comma:
+
+                # in progress
+
+                self.platepar.az_centre, self.platepar.el_centre = self.faintestUnmatchedStar()
+
+                self.checkParamRange()
+                self.platepar.updateRefRADec(preserve_rotation=True)
+                self.checkParamRange()
+
+                self.tab.param_manager.updatePlatepar()
+                self.updateLeftLabels()
+                self.updateStars()
+
+            # Jump to brightest unmatched star
+            elif event.key() == QtCore.Qt.Key_Period:
+
+                # in progress
+                self.platepar.az_centre, self.platepar.el_centre = self.brightestUnmatchedStar()
+                self.checkParamRange()
+                self.platepar.updateRefRADec(preserve_rotation=True)
+                self.checkParamRange()
+
+                self.tab.param_manager.updatePlatepar()
+                self.updateLeftLabels()
+                self.updateStars()
+
+
+            # Jump to furthest unmatched star
+            elif event.key() == QtCore.Qt.Key_Backslash:
+
+                new_x, new_y = self.furthestStar()
+                new_x, new_y = int(new_x), int(new_y)
+                self.img_frame.setRange(xRange=(new_x+15, new_x-15), yRange=(new_y+15, new_y-15))
+                self.checkParamRange()
+                self.platepar.updateRefRADec(preserve_rotation=True)
+                self.checkParamRange()
                 self.tab.param_manager.updatePlatepar()
                 self.updateLeftLabels()
                 self.updateStars()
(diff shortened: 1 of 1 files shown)


g7gpr:
Changed to using ] key in the interest of better ergonomics

g7gpr:
-m followed by the location of the directory holding the mask will prevent stars which are obscured from being drawn, and attempting to match

g7gpr:
```
python -m Utils.SkyFit2 /home/david/platepars/au000u -m /home/david/platepars/au000u

```
will load the mask.bmp from /home/david/platepars/au000u/

dvida:
Very cool, this seems like a great feature!
Before I merge it, a few suggestions:
- The ] key is not too accessible on some non-english keyboards. I recommend putting a button somewhere in the GUI for this feature, or perhaps give a CTRL + something alternative.
- I suggest that if no path for the -m option is given, the working directory is used to look for the mask.

dvida:
Could you also check if loading an old state file works? I think every time we add a new argument to the class, we need to modify this function:
https://github.com/CroatianMeteorNetwork/RMS/blob/3a4d3058ae05dc30c9e5dd5922d41a8d6d739c0a/Utils/SkyFit2.py#L2194

I belive that the new arguments need to be explicitly defined here.

Have you also checked what happens when there are no stars in the calstars file, i.e. it's empty? This might be relevant when non-RMS data is used.

g7gpr:
For masks 
first choice is to to take the command line option
next choice it to use the path in the .config file
final choice is to pick up the mask.bmp in the working directory

g7gpr:
Not having a CALSTARS results in one being created. An empty file just returns an empty list, so the green circles do not get drawn. The code is directed by the catalog stars, limited by FOV and mask. This does make the coordinate handling awkward, because I need to jump to where they are on the screen with the as yet imperfect platepar; but I don't think I rely on CALSTARS.

g7gpr:
I've added in the code to load a mask when resuming from a state, but wouldn't it be better to save the mask in the statefile? This is not what you asked for, so I guess there is a good reason...

g7gpr:
This is ready for review - not ready for merging. The gui code needs some work. I need to make a lot of platepars over the next few days, so it will get a good test.

g7gpr:
This ready for review

g7gpr:
Nearest I can do is mark that star as bad, otherwise algorithm will keep jumping back to it. I think that solves the problem.

dvida:
That's great. If a star is skipped once, mark it as bad and don't let it be considered again.

g7gpr:
SkyFit2 with avoiding unsuitable stars

Red squares denote unsuitable stars, which should be avoided. Press K instead of space after selection.

dvida:
Could you change jumping to the next star to CTRL+ Space? That way, the order of operation is:
1) CTRL + R - star picking mode
2) CTRL + O - auto pan
3) CTRL + Space - suggest star
4) Select star and hit Space to confirm
5) Repeat 3-4

The way the fingers on the left hand never leave the space button.

Also, would it be possible to check if the current star is selected or not, so we can avoid a separate button for selecting bad stars? E.g. if I don't like the star, I just hit CTRL + Space to skip it.

We also need a label somewhere indicating we're in the auto pan mode.

Finally, could you swap the fit and next star buttons in the GUI? I'd like the fit button to be at the top, as people got accustomed to it being there.

g7gpr:
Sure, you did see the tick box on the settings tab? This automates it completely. Click and go.

dvida:
Brilliant!
I wonder if we should keep a count of stars somewhere handy, so the user can easily tell how many were selected and how many are left.

g7gpr:
That code exists but is if Falsed out. Putting this back to. Draft to reflect additional work required.

g7gpr:
"Also, would it be possible to check if the current star is selected or not, so we can avoid a separate button for selecting bad stars? E.g. if I don't like the star, I just hit CTRL + Space to skip it."

I think the requirement here is select a star, press enter to confirm, or ctrl+enter to mark unsuitable and skip?

dvida:
Correct. You can simply check if the currently considered star is in the list of paired stars. If it isn't and CTRL + Space is pressed, add it as unsuitable and go to the next one.

g7gpr:
```
        if source_intens > 0:
            x_centroid = x_acc/source_intens + x_min
            y_centroid = y_acc/source_intens + y_min
```

Added in a conditional to catch divide by 0, but not sure what the consequences would be

g7gpr:
Unable to implement all the requests, ctrl+space was already bound when no star was selected. The order of operation is 

```
1. CTRL + R star picking mode
2. CTRL + O auto pan
3. Click on star
4. Space to jump to next star - goto 3

In the case of a bad star

5. Click on star
6. Ctrl + Space - mark as bad

In case of wanting to jump to a different star which is not matched or marked bad

7. Shift and Space
```

For 7 I couldn't do anything clever; the state wasn't changing, because no stars were selected. Therefore it would just get stuck in a loop jumping between two stars.

dvida:
I believe I was able to achieve the desired behaviour, but more testing is needed. The furthestStar function needs to be cleaned up a bit, I think there's too much back and forth in terms of coordinate transforms that are unnecessary.

g7gpr:
I'll rewrite in image coordinates and simplify.

g7gpr:
Ready for testing

PR #254, opened 2024-01-13
Feature download mask

g7gpr:
Wrapping sftp actions in try: except:

This is still in testing. I've never observed the conditions that would require the except: clause to execute. I'll do some simulation work.


dvida:
Isn't this done, can we close this PR?

PR #256, opened 2024-01-17
FRbinViewer - append meteor image to the end of video and add meteor shower information

aitov:
Hi, 
As I can see my everyday usage of FRbinViewer it is  generating meteor mp4 video for bin files with options:

- --avg - use average pixel for background to generate pure video of meteor (with black background)
-  --timestamp - add date and time information to video

Usually bright and cool meteors shared with friends or social media and it requires additional action - append final image of meteor to the end of video for 1 or 2 seconds by some video editor.

In this pull request I added two additional  options:
- --append_ff_to_video - append to the end of video ff fits image (max pixel) for 1.5 seconds
- --add_shower_name - add same as timestamp to video shower name which detected by showerAssociation for single station

Probably these options will be helpful for someone.

Tested on Mac and RPi.

### DIFF
--- a/Utils/FRbinViewer.py
+++ b/Utils/FRbinViewer.py
@@ -32,6 +32,7 @@ from RMS.Formats import FFfile, FRbin
 import datetime
 from glob import glob
 from Utils.ShowerAssociation import showerAssociation
+from RMS.Formats.FTPdetectinfo import validDefaultFTPdetectinfo
 
 
 def view(dir_path, ff_path, fr_path, config, save_frames=False, extract_format=None, hide=False,
@@ -180,12 +181,8 @@ def view(dir_path, ff_path, fr_path, config, save_frames=False, extract_format=N
 
                 # Put the name of the FR file, followed by the frame number
                 # Put a black shadow
-                cv2.putText(img, fr_path + " frame = {:3d}".format(t), (10, 20), 
-                            fontFace=cv2.FONT_HERSHEY_SIMPLEX , fontScale=0.5, color=[0, 0, 0], 
-                            lineType=cv2.LINE_AA, thickness=2)
-                cv2.putText(img, fr_path + " frame = {:3d}".format(t), (10, 20), 
-                            fontFace=cv2.FONT_HERSHEY_SIMPLEX , fontScale=0.5, color=[255, 255, 255], 
-                            lineType=cv2.LINE_AA, thickness=1)
+                title = fr_path + " frame = {:3d}".format(t)
+                addTextToImage(img, title, 10, 20)
 
             # Add timestamp
             if add_timestamp:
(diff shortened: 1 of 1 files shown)


Utils/FRbinViewer.py
markmac99, line 1: * I noticed that dir_path and config are not declared or passed into loadShowerAssociations() so this function only works because variables with the same name are declared with global scope.  I feel the two should be passed in as parameters so that it is clear where dir_name and config come from. and theres no possibility of them being overriden by other globals.  This would also allow the function to be called from other modules which might be handy (I have a use-case in mind!). 

* Recommend to use the function validDefaultFTPdetectinfo() in RMS.Formats.FTPdetectinfo to filter for the correct FTPdetectinfo file. This routine filters out all the backups, pre-ML and other versions etc and means we're consistently finding the file. So at line 368 you can use the below, and line 369 is not needed 
        ftp_list = [x for x in ftp_list if validDefaultFTPdetectinfo(os.path.basename(x))]

* Also note that to test if a key is in a list, you can do "if key in list:" eg "if (ffName, 1.0) in associations:". This is neater than using exception handling, and it means that if a real exception occurs, you don't miss it. 

* Recommend that loadShowerAssociations() should log a warning and then return False if it can't find the FTP file, instead of exiting. The failure can then be handled by ignoring the flag since after all, the video can still be generated. In general its not best practice to exit() from within a function as it can cause unexpected aborts. 

* When looking up an ff_file in associations{} the 2nd element is the meteor number if there's more than one meteor in the FF file. This can be read from the FTPdetect file and is not always 1.0. as especially during busy showers there may be two or three meteors in an FF file.  I tihnk the code therefore needs to load the meteor number from the ftp file before doing these lookups. 

* The body of function SaveFramesForMeteorImage is indented by two tabstops instead of one. This causes the code to abort on some platforms, and my linter complains a lot! 

* line 335, "title" is indented by seven spaces instead of eight. Most python interpreters require multiples of four (or whatever tab is set to) so this might cause an abort on some platforms.  

* Also not strictly related to this PR but dir_path should be converted to an absolute path before being used. I noticed that if a relative path is passed to FRbinViewer, then the call to ffmpeg fails as it doesn't understand the abs path.  
So at line 427:   dir_path = os.path.abspath(cml_args.dir_path[0])

markmac99:
I also noticed that the videos that this process creates won't play on my Pi4. The same video works with VLC on my Windows PC, but not on VLC on the Pi4, I'd be interested to hear if you could get the videos to play on a Mac. On the Pi, VLC opens, the video screen flickers black and then VLC displays its traffic-cone icon. The video never runs,
I dont think this work has introduced the problem but maybe we can fix it. I think the issue is that by default ffmpeg creates video with the yuv444 format, whereas most video players only support yuv420p (unless you have an extended codec pack installed like k-lite for windows). I'll do some tests

markmac99:
As a concrete example of the issue with meteor# and lookup of associations, in my sample dataset from the Geminids, FRbinViewer incorrectly labelled several meteors as Geminids instead of COM or DSV because it picked the first event in the FF file.

aitov:
Hi @markmac99 , thank you for code review and comments.



> I also noticed that the videos that this process creates won't play on my Pi4. The same video works with VLC on my Windows PC, but not on VLC on the Pi4, I'd be interested to hear if you could get the videos to play on a Mac. On the Pi, VLC opens, the video screen flickers black and then VLC displays its traffic-cone icon. The video never runs,
> I dont think this work has introduced the problem but maybe we can fix it. I thought it was yuv444 vs yuv420 but that bug got fixed already. Hmm.

Yes, it was fixed by -pix_fmt yuv420p. I tested on my RPi 4 and it works for me - generated on RPi 4 mp4 could be played by VLC on RPi and on my Mac, same generated video on Mac could be played on RPi4. Probably you could share your file I will check.

 
> When looking up an ff_file in associations{} the 2nd element is the meteor number if there's more than one meteor in the FF file. This can be read from the FTPdetect file and is not always 1.0. as especially during busy showers there may be two or three meteors in an FF file. I tihnk the code therefore needs to load the meteor number from the ftp file before doing these lookups.

> As a concrete example of the issue with meteor# and lookup of associations, in my sample dataset from the Geminids, FRbinViewer incorrectly labelled several meteors as Geminids instead of COM or DSV because it picked the first event in the FF file.

So possible case when we have more than 1 meteor on FF file which used for background for FR bin. 

The question: how to identify record in FTPdetect for target  FR file?

Other comments fixed.

Thanks

markmac99:
Updates look good. I tested on Windows 10, Ubuntu (WSL) and Pi4 32bit, in various combinations, no problems. I think the pull request can be merged. 

The videos still won't play on any of my Pi4s though. I tried creating them on both Windows and Pi . In both cases the videos play ok on Windows but won't play on the Pi4. I tried on UK0006, UK000F, UK001L and UK002F as well as on my spare Pi4.  VLC version is 3.0.20 Vetinari on both the Pis and on Windows so i think this is more likely a codec issue.

markmac99 (approved):
looks good now - thanks for taking in my suggestions :)

markmac99:
ps would be interested to see if the videos play for you, though  i think they will

https://github.com/CroatianMeteorNetwork/RMS/assets/25826525/2ab773a5-75c1-4161-a260-1fb93f792a25

aitov:
Hi,
> ps would be interested to see if the videos play for you, though i think they will
> 
>  pi-version-FR_UK001L_20231214_020505_502_0855808_line_00.mp4_shwr_ff.mp4

I tested on my RPi4 - video played without any issues, so probably codec issue, not sure how to compare installed libs, but I used standard Buster image for RMS

> The question: how to identify record in FTPdetect for target FR file?

I can fix this part as well, but I need help with solution, otherwise sometimes it will print wrong shower name

markmac99:
All my pis are also Buster but were built from the image from 2018, though i keep them updated every month. I'll see if the internet has any codec/driver suggestions or ways to get debug info from VLC. 
I think i have a solution for the problem of identifying the correct meteor in an FF file, I'll do some tests and let you know my ideas. 
ps Thanks for this pull request, i think its a useful addition to the toolset!

aitov:
Hi @markmac99 , @dvida 
As I can see this PR was merged to  dev branch.
Seems like I have solution to close last item  :

> So possible case when we have more than 1 meteor on FF file which used for background for FR bin.

and ready create new PR with this fix, please let me know if it should be created for dev or master branch

dvida:
Thanks! Please make it for the dev branch. I'm planning to merge it soon.

Denis

PR #260, opened 2024-01-31
Plot, score, and log timestamp intervals for analysis

Cybis320:
This PR introduces `PlotTimeIntervals.py`, a utility that generates a plot displaying the intervals between FF file timestamps. It also assigns a score ranging from 0 to 1000, with higher scores indicating less variability in the timestamps.

During reprocessing, this utility is invoked on the `night_dir`, logging a score and incorporating it into the plot's filename. If insufficient intervals are present to calculate a meaningful score, none is returned. The plot and score aim to function as diagnostic tools for identifying timestamping and dropped frame issues.

The utility uses Pandas to find outliers, and `pandas` has been added to requirements.txt. If this is a pain point, the utility can be redisigned without this feature.

The scoring is somewhat arbritary and designed to offer what feels like the right sensitivity.
What are your thoughts?

Edit: Denis, thank you for suggesting to use tar.bz2 - I've incorporated that suggestion.

### DIFF
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -401,7 +401,8 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
         log.info('Timestamp Intervals Analysis: Jitter Quality: {:.1f}%, Dropped Frame Rate: {:.1f}%'.format(
             jitter_quality, dropped_frame_rate))
         # Add the timelapse to the extra files
-        extra_files.append(intervals_path)
+        if intervals_path is not None:
+            extra_files.append(intervals_path)
 
     except Exception as e:
         log.debug('Plotting timestamp interval failed with message:\n' + repr(e))
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -397,8 +397,8 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
 
     # Plot timestamp intervals
     try:
-        score, intervals_path =  analyze_timestamps(night_data_dir, fps=config.fps)
-        log.info(f'Timestamp Intervals Score: {score}')
+        jitter_quality, dropped_frame_rate, intervals_path = analyze_timestamps(night_data_dir, fps=config.fps)
+        log.info('Timestamp Intervals Analysis: Jitter Quality: {:.1f}%, Dropped Frame Rate: {:.1f}%'.format(jitter_quality, dropped_frame_rate))
         # Add the timelapse to the extra files
         extra_files.append(intervals_path)
 
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -395,14 +395,12 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
 
     log.info('Plotting timestamp intervals...')
 
-    # Plot timestamp interval
+    # Plot timestamp intervals
     try:
-        analyze_timestamps(night_data_dir)
-
-        timelapse_path = os.path.join(night_data_dir, timelapse_file_name)
-
+        score, intervals_path =  analyze_timestamps(night_data_dir)
+        log.info(f'Timestamp Intervals Score: {score}')
         # Add the timelapse to the extra files
-        extra_files.append(timelapse_path)
+        extra_files.append(intervals_path)
 
     except Exception as e:
         log.debug('Plotting timestamp interval failed with message:\n' + repr(e))
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -35,7 +35,7 @@ from Utils.MakeFlat import makeFlat
 from Utils.PlotFieldsums import plotFieldsums
 from Utils.RMS2UFO import FTPdetectinfo2UFOOrbitInput
 from Utils.ShowerAssociation import showerAssociation
-
+from Utils.PlotTimeIntervals import analyze_timestamps
 
 # Get the logger from the main module
 log = logging.getLogger("logger")
@@ -393,6 +393,20 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
             log.debug('Generating a timelapse failed with message:\n' + repr(e))
             log.debug(repr(traceback.format_exception(*sys.exc_info())))
 
+    log.info('Plotting timestamp intervals...')
+
+    # Plot timestamp interval
+    try:
+        analyze_timestamps(night_data_dir)
+
+        timelapse_path = os.path.join(night_data_dir, timelapse_file_name)
+
+        # Add the timelapse to the extra files
+        extra_files.append(timelapse_path)
+
+    except Exception as e:
+        log.debug('Plotting timestamp interval failed with message:\n' + repr(e))
+        log.debug(repr(traceback.format_exception(*sys.exc_info())))
 
 
     ### Add extra files to archive
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -397,7 +397,7 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
 
     # Plot timestamp intervals
     try:
-        score, intervals_path =  analyze_timestamps(night_data_dir)
+        score, intervals_path =  analyze_timestamps(night_data_dir, fps=config.fps)
         log.info(f'Timestamp Intervals Score: {score}')
         # Add the timelapse to the extra files
         extra_files.append(intervals_path)
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -35,7 +35,7 @@ from Utils.MakeFlat import makeFlat
 from Utils.PlotFieldsums import plotFieldsums
 from Utils.RMS2UFO import FTPdetectinfo2UFOOrbitInput
 from Utils.ShowerAssociation import showerAssociation
-from Utils.PlotTimeIntervals import analyze_timestamps
+from Utils.PlotTimeIntervals import plotFFTimeIntervals
 
 # Get the logger from the main module
 log = logging.getLogger("logger")
@@ -397,8 +397,9 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
 
     # Plot timestamp intervals
     try:
-        jitter_quality, dropped_frame_rate, intervals_path = analyze_timestamps(night_data_dir, fps=config.fps)
-        log.info('Timestamp Intervals Analysis: Jitter Quality: {:.1f}%, Dropped Frame Rate: {:.1f}%'.format(jitter_quality, dropped_frame_rate))
+        jitter_quality, dropped_frame_rate, intervals_path = plotFFTimeIntervals(night_data_dir, fps=config.fps)
+        log.info('Timestamp Intervals Analysis: Jitter Quality: {:.1f}%, Dropped Frame Rate: {:.1f}%'.format(
+            jitter_quality, dropped_frame_rate))
         # Add the timelapse to the extra files
         extra_files.append(intervals_path)
 
--- a/Utils/FluxAuto.py
+++ b/Utils/FluxAuto.py
@@ -456,7 +456,7 @@ def fluxAutoRun(config, data_path, ref_dt, days_prev=2, days_next=1, all_prev_ye
     metadata_dir=None, output_dir=None, csv_dir=None, index_dir=None, generate_website=False, 
     website_plot_url=None, shower_code=None, shower_suffix_filename=None, custom_binning_dict=None,
     cpu_cores=1, excluded_stations_file="excluded_stations.txt",
-    skip_allyear=False):
+    skip_allyear=False, publication_quality=False):
     """ Given the reference time, automatically identify active showers and produce the flux graphs and
         CSV files.
 
@@ -487,6 +487,7 @@ def fluxAutoRun(config, data_path, ref_dt, days_prev=2, days_next=1, all_prev_ye
         excluded_stations_file: [str] File with excluded stations and periods. It should be in the metadata
             directory.
         skip_allyear: [bool] Skip computing the flux with all years.
+        publication_quality: [bool] Produce plots in publication quality. False by default.
     """
 
 
@@ -850,7 +851,8 @@ def fluxAutoRun(config, data_path, ref_dt, days_prev=2, days_next=1, all_prev_ye
                 compute_single=False,
                 show_plot=False,
                 xlim_shower_limits=True,
-                sol_marker=sol_ref
+                sol_marker=sol_ref,
+                publication_quality=publication_quality
             )
 
             # Save the batch flux plot (full metadata)
@@ -862,7 +864,8 @@ def fluxAutoRun(config, data_path, ref_dt, days_prev=2, days_next=1, all_prev_ye
                 compute_single=False,
                 show_plot=False,
                 xlim_shower_limits=True,
-                sol_marker=None
+                sol_marker=None,
+                publication_quality=publication_quality
             )
 
             # Save the results to a CSV file
@@ -969,6 +972,9 @@ if __name__ == "__main__":
 
     arg_parser.add_argument('--skipallyear', action="store_true", \
         help="""Skip computing multi-year fluxes. Only compute fluxes for the given year.""")
+    
+    arg_parser.add_argument('--publication', action="store_true", \
+        help="""Produce plots in publication quality.""")
 
 
     # Parse the command line arguments
@@ -1018,7 +1024,9 @@ if __name__ == "__main__":
             website_plot_url=cml_args.weburl, shower_code=cml_args.shower, \
             shower_suffix_filename=cml_args.suffix, custom_binning_dict=custom_binning_dict,
             cpu_cores=cml_args.cpucores,
-            skip_allyear=cml_args.skipallyear)
+            skip_allyear=cml_args.skipallyear,
+            publication_quality=cml_args.publication
+            )
 
 
         ### <// DETERMINE NEXT RUN TIME ###
--- a/Utils/FluxBatch.py
+++ b/Utils/FluxBatch.py
@@ -1317,7 +1317,7 @@ def fluxBatch(config, shower_code, mass_index, dir_params, ref_ht=-1, atomic_bin
 
 
 def plotBatchFlux(fbr, dir_path, output_filename, only_flux=False, compute_single=False, show_plot=True,
-    xlim_shower_limits=False, sol_marker=None):
+    xlim_shower_limits=False, sol_marker=None, publication_quality=False):
     """ Make a plot showing the batch flux results. 
     
     Arguments:
@@ -1333,6 +1333,8 @@ def plotBatchFlux(fbr, dir_path, output_filename, only_flux=False, compute_singl
             by default.
         sol_marker: [float] Plot a red vertical line on the flux plot at the given solar longitude (deg).
             None by default, in which case the marker will not be plotted.
+        publication_quality: [bool] If True, make the plot publication quality with no grid, larger font,
+            and only keeping the flux and TAP subplots, not the individual parameters. False by default.
     """
 
 
@@ -1341,14 +1343,34 @@ def plotBatchFlux(fbr, dir_path, output_filename, only_flux=False, compute_singl
     plot_info = StationPlotParams()
     
     if only_flux:
+
         subplot_rows = 1
         figsize = (15, 5)
+
     else:
-        subplot_rows = 4
-        figsize = (15, 10)
+
+        if publication_quality:
+            subplot_rows = 2
+            figsize = (15, 10)
+
+        else:
+            subplot_rows = 4
+            figsize = (15, 10)
 
     fig, ax = plt.subplots(nrows=subplot_rows, figsize=figsize, sharex=True, \
         gridspec_kw={'height_ratios': [3, 1, 1, 1][:subplot_rows]})
+    
+
+    # If the plot is for publication, set the plot parameters
+    if publication_quality:
+
+        plt.rcParams.update({'font.size': 16})
+        plt.rcParams.update({'axes.labelsize': 16})
+        plt.rcParams.update({'xtick.labelsize': 16})
+        plt.rcParams.update({'ytick.labelsize': 16})
+        plt.rcParams.update({'legend.fontsize': 16})
+        plt.rcParams.update({'figure.autolayout': True})
+        plt.rcParams.update({'axes.grid': False})
 
 
     if not isinstance(ax, np.ndarray):
@@ -1421,24 +1443,31 @@ def plotBatchFlux(fbr, dir_path, output_filename, only_flux=False, compute_singl
             capsize=3,
         )
 
+        ax[0].legend()
+
         # Set the minimum flux to 0
         ax[0].set_ylim(bottom=0)
 
-        # Add the grid
-        ax[0].grid(color='0.9')
-
-        ax[0].legend()
-        ax[0].set_title("{:s}, v = {:.1f} km/s, s = {:.2f}, r = {:.2f}".format(fbr.shower.name_full, 
-            fbr.v_init/1000, calculateMassIndex(np.mean(fbr.summary_population_index)), 
-            np.mean(fbr.summary_population_index)) 
-                      # + ", $\\mathrm{m_{lim}} = $" + r"${:s}$ g ".format(formatScientific(1000*mass_lim, 0))
-                      # + "at LM = +6.5$^{\\mathrm{M}}$"
-                      )
         ax[0].set_ylabel("Flux (meteoroids / 1000 $\\cdot$ km$^2$ $\\cdot$ h)")
 
-        # Plot the marker
-        if sol_marker is not None:
-            ax[0].axvline(x=sol_marker, color='r', linewidth=1)
+
+        if not publication_quality:
+
+            # Add the grid
+            ax[0].grid(color='0.9')
+            
+            # Set the title
+            ax[0].set_title("{:s}, v = {:.1f} km/s, s = {:.2f}, r = {:.2f}".format(fbr.shower.name_full, 
+                fbr.v_init/1000, calculateMassIndex(np.mean(fbr.summary_population_index)), 
+                np.mean(fbr.summary_population_index)) 
+                        # + ", $\\mathrm{m_{lim}} = $" + r"${:s}$ g ".format(formatScientific(1000*mass_lim, 0))
+                        # + "at LM = +6.5$^{\\mathrm{M}}$"
+                        )
+            
+
+            # Plot the marker
+            if sol_marker is not None:
+                ax[0].axvline(x=sol_marker, color='r', linewidth=1)
 
 
         ### Plot the ZHR on another axis ###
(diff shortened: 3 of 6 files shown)


Cybis320:
Sample:
US9999_20240107_144910_590515_intervals_score_978

Cybis320:
And it can be used as a cmd line utility on existing archived folders:
`python -m Utils.PlotTimeIntervals "$input_path" "$fps"`

Cybis320:
Updated the PlotTimeIntervals script to recursively search through all subdirectories of a given path for FS*.tar.bz2 files. For each matched file, the script now generates a plot.

You can invoke the updated script like this:

`python -m Utils.PlotTimeIntervals /home/pi/RMS_data/ArchivedFiles 25`

This command will process all suitable files in /home/pi/RMSdata/ArchivedFiles and its subdirectories, assuming a frame rate of 25 fps for the analysis.

dvida:
Thanks! Could you also strip the f strings and use argparse (see the bottom of e.g. Utils.StackFFs) in the main section?

Cybis320:
Will do. Re pandas / outliers, yay or nay?

dvida:
Oh, I haven't noticed pandas there. I'm very reluctant about adding new libraries, remote installation can break setups in many different ways on the many machines people operate it on. I'd rather drop it if that's OK.

Cybis320:
Easy for this one. The GStreamer based bufferedCapture requires a new dependency though...

dvida:
That right, that's why we need to make the old behaviour the default one (perhaps try to fix it with the cv2 version as best as we can), and the gstreamer will only be rolled out in the new versions of the SD card image. We also need to be careful that all the new imports are conditional and don't break under existing installations.

Cybis320:
Updated the `PlotTimeIntervals` script with the following changes:

- Removed usage of f-strings and Pandas.
- Integrated `argparse`.
- The script now attempts to read the `.config` file in the same directory as the `.tar.bz2` file to determine the `fps` value. If `.config` is not found or doesn't contain `fps`, the script falls back to using the `fps` value provided as a command-line argument or 25 if no argument is provided.


You can run the updated script with the following command:
```bash
python -m Utils.PlotTimeIntervals /home/pi/RMS_data/ArchivedFiles --fps 25
```
This command processes all subdirectories in `/home/pi/RMS_data/ArchivedFiles`, using the fps value from the subdirectories `.config` files when available or defaulting to 25 fps if not.

During reprocessing, this utility is invoked automatically on the night_dir, logging a score and incorporating it into the plot's filename.

dvida:
Amazing work! And maybe just one final nitpick - could we make the score a physically meaningful number? E.g. average late time in ms. Or the percentages of FF files which are more than 1 frame early and 1 frame late than the nominal time. It would also be good to mark those lines on the plot.
The idea behind this is that it's immediately obvious to operators what the number means so they don't have to consult additional documentation or they act on a wrong assumption. Adding a short explanation directly in the plot is encouraged.

Cybis320:
Thanks, Denis, for all the great suggestions, please keep them coming! I experimented with the percentage of files within ±1/fps of the target, where 100 is perfect. See attached example, let me know what you think.

The expected and average lines are already in the legend. Did you mean to label the lines themselves?
US9999_20231221_011659_875432_detected_intervals_score_59

dvida:
That looks great, thanks! I meant that the +/-1 frame lines should be plotted around the expected value, but I guess they would be too close together to see in the plot. One thing that I noticed is that it looks like each outlier has a pair on the other side. Is this real or just a coincidence? If it's real it would mean that RMS compensates with the next frame to bring the timing back into some order. What if you applied a 3-frame moving average over the numbers? That might give us a better idea of systematic offsets vs short-term jitter? I'd leave individual points and overplot the moving average.

Cybis320:
Here's a bad SD card example:
US9999_20231205_005351_882274_intervals_score_36

Cybis320:
Here a GStreamer example:
US9999_20240201_064946_820370_detected_intervals_score_100

Cybis320:
Upper and lower interval line. Are the labels ok?
Regarding the rebound in interval, yes, I think there is a return to average effect. But it does mean that those timestamps are less accurate. In the bad-SD-card example, there is no return to average there. I'll play with moving avg next.

US9999_20231221_011659_875432_detected_intervals_score_59

Cybis320:
Maybe it needs two scores to differentiate jitter from dropped frames...

Cybis320:
Ok, moving average with long window (~10) is promising. I still need to work on it, but here are some preliminary results:

Here's CV2 with a good SD card:
US9999_20231217_072712_688922_detected_intervals_score_49

And CV2 with a bad SD card:
US9999_20231205_005351_882274_intervals_score_36

Cybis320:
Directly plotting the moving average doesn't work well. How about color coding the scatter points based on whether on not they fall in a time period where the moving average exceed a certain threshold. The above graphs would then look like this:
US9999_20231217_072712_688922_detected_intervals_score_49
US9999_20231205_005351_882274_intervals_score_36

Cybis320:
Latest iteration:

- Latest iteration: The score has been renamed to Jitter Score.

- Added a Dropped Frames score, which estimates the percentage of FF files likely to contain dropped frames. This is achieved by calculating a moving average over 50 FF files; any FF file with a timestamp during a time when the moving average exceeds 257/fps is tagged. The score is the fraction of untagged intervals over the total intervals. 1A score of 100% indicates that dropped frames are unlikely, while 0% means that all FF files contain at least one dropped frame. Intervals that are tagged are colored red. This method is not 100% accurate, as there is no perfectly accurate way to detect dropped frames with CV2.

- Tweaked the layout, legend, labels, etc.

- Files are now saved with both scores.

I need to clean up the code before pushing it, but here are some sample plots:
 

US9999_20231217_072712_688922_detected_intervals_scores_49-100
US9999_20231205_005351_882274_intervals_scores_36-37

dvida:
This is amazing, great work! I think we're 99% there. Here are a few suggestions
- Can you plot the median instead of the average interval? This will remove the focus from the outliers to the actual average performance.
- Make the +/- 1 frame lines a different colour for better visibility.
- The scoring system is excellent. For clarity, I suggest the following label changes to indicate that higher is better for the jitter score. Also, I suggest changing the dropped frame score to 100% - score so anything to do with dropped frames being around 0 is good. I also suggest concisely quantifying the idea behind each score:
    - Jitter Quality (intervals within +/-1 frame): 100%
    - Dropped Frame Score (intervals >2 frames late within 50 FF files): 0%
- Can the image size go down? The plots are over half a meg now, I'd like to see it go < 100 kb, ideally < 50 kb.

Cybis320:
All great observations and suggestions as usual. I agree with the idea to make the dropped frame number a lower-is-better value. I tend to associate 'score' with higher-is-better... Dropped Frame Rate?
Lower the size?? But, I was going to frame this artwork! Hehe.
I'll play with figure size, dpi and jpg compression to see how low I can go without losing readability.

dvida:
> Dropped Frame Rate
Sold!

Cybis320:
Mean, Median, Mode - make your choice! :)
US9999_20231221_011659_875432_detected_intervals_scores_59-0

dvida:
Thanks! Could you double-check the way the median is computed? It's very strange that it's so far off from the other two values.

Cybis320:
100 dpi > 107 KB
150 dpi > 190 KB
200 dpi > 289 KB
250 dpi > 409 KB
They look the same in the browser, but look incrementally better when opened to full size.
_______

100 dpi > 107 KB:
US9999_20231221_011659_875432_detected_intervals_scores_59-0_100dpi
______
150 dpi > 190 KB:
US9999_20231221_011659_875432_detected_intervals_scores_59-0_150dpi
______
200 dpi > 289 KB:
US9999_20231221_011659_875432_detected_intervals_scores_59-0-200dpi
_____
250 dpi > 409 KB:
US9999_20231221_011659_875432_detected_intervals_scores_59-0_250dpi

Cybis320:
Good catch on the median! It helps to sort the list first...
US9999_20231221_011659_875432_detected_intervals_scores_59-0

Cybis320:
US9999_20231205_005351_882274_intervals_scores_36-63

dvida:
np.median is the fastest way btw, if you're doing anything else :)
np.mean for the mean, etc.
Let's just keep the median to make the plot less busy.

You can try setting a small figsize perhaps that will help. I find it usual that the png's are that big while other PNGs that RMS produced are quite small even though they're also quite busy. 150 dpi is probably the sweet spot

dvida:
Oh! And how about showing the median and the standard deviation of the scatter? E.g. 10.248 +/- 0.05 s / 1.1 frame. That way the data is fully characterized.

Cybis320:
Like so?
US9999_20231221_011659_875432_detected_intervals_scores_59-0

dvida:
<chef's kiss>
Do you have anything else planned or can I test and deploy?

Cybis320:
I think it’s ready to test. Fingers crossed.On Feb 7, 2024, at 2:37 PM, Denis Vida ***@***.***> wrote:﻿
<chef's kiss>
Do you have anything else planned or can I test and deploy?

dvida:
NL000M_20231215_161435_236036_ff_intervals

I made a few little tweaks I'll push now, but it works great! This is what I see for a normal camera. Most look like this or have a handful of longer dropouts in the night. But the bulk of the points are well within the 1-frame boundary.

UK00BA_20231227_164748_034400_ff_intervals

Cybis320:
Is there any alerting system that could send you and/or the operator messages if things go haywire ?

Cybis320:
And to analyze past performance, one can also run the utility as a command line on the ArchivedFiles folder and it will generates plots on all sub folders. 

python -m Utils.PlotTimeIntervals /home/pi/RMS_data/ArchivedFiles

g7gpr:
The capture directory which caused the error was terminated shortly after capture started, this is why there were two captures on the same night. Would you like a copy of that directory?

g7gpr:
AU0028_20240204_114635_244720_ff_intervals
AU0028_20240205_114542_005307_ff_intervals
AU0028_20240206_114445_852284_ff_intervals
AU0028_20240207_115627_393724_ff_intervals

g7gpr:
I wonder if this is the camera automatic reboot?

```
python -m Utils.CameraControl SetAutoReboot Everyday,15
```

By the way, this is absolutely brilliant!

markmac99:
@dvida @Cybis320 looks good. Ran it on two of my production systems with no issues. It would be good to put this on the weblog, once its available. 

Might be worth fixing the Y axis scale to a range of 0-20 so that you can easily compare between days. Some points could then be off the top of the range, but they could be artifically forced in-range for display purposes.  I thought UK002F was much worse than UK0006 on 07/dec but it turned out that the scales were different (8-12 vs 0-20s! ). 2F was actually _better_ but because of the scale change, the scatter was more evident and at a casual glance i mistook it for worse data. 

Do we have a view on what a "bad" plot would look like? Would be useful to add a comment if we think the plot requires investigation. More than 5% possible dropped frames ? Jitter worse than 80%? 

UK0006 produced one day out of five where jitter quality dropped below 90% and poss dropped frames rose above 4%. UK002F also produced one day (different to 6) where JQ dropped below 90% but dropped frames was zero on that day.  Neither camera seems to be generating bad data.

markmac99:
Graphs in case of interest
UK0006_20240203_173148_513170_ff_intervals
UK0006_20240204_173331_412120_ff_intervals
UK0006_20240205_173514_827221_ff_intervals
UK0006_20240206_173658_910097_ff_intervals
UK0006_20240207_173842_246313_ff_intervals
UK002F_20240203_173148_041874_ff_intervals
UK002F_20240204_173330_896873_ff_intervals
UK002F_20240205_173514_634923_ff_intervals
UK002F_20240206_173658_558385_ff_intervals
UK002F_20240207_173843_059439_ff_intervals

markmac99:
@g7gpr  david - yes, that sets the camera to auto-reboot at 1500 according to its internal timestamp. The camera takes a few milliseconds to reboot so there should be little data loss (maybe one or two frames). 
The next line sets the camera's internal clock to the current time on the computer from which you run the script.  
```
python -m Utils.CameraControl CameraTime set
```
My original instructions on the wiki instructed the user to run this script _on their pc_ which ensured that the camera picked up your local time so that 1500 was well outside potential data capture times.

Cybis320:
> The capture directory which caused the error was terminated shortly after capture started, this is why there were two captures on the same night. Would you like a copy of that directory?

Thank you for the report. The code was attempting to compute the moving average when there was not enough data to do so. I just pushed a fix for that.

dvida:
I agree with Mark that some fixed limits should be used. I added a plot with residuals from the expected below the main plot with fixed limits to +/- 2 frames. This way even if there is one large outlier, the rest of the points will be visible well.
UK00B8_20231225_164617_621911_ff_intervals
UK00BA_20231227_164748_034400_ff_intervals
AU0044_20240118_115030_630120_ff_intervals

Cybis320:
That's awesome!

Cybis320:
Showing off the GStreamer method 😄 
US9999_20240208_013640_169265_ff_intervals

Cybis320:
Would you be opposed to add back some indication in the plot filename that something may be amiss?
Appending a flag word when the dropped frame rate is not zero, or above a fairly low threshold. 
Right now, one would have to open and view each plot to discover an issue. It would be convenient to just search for the keyword to find potential issues. 
However, I understand if we prefer not to cause concern with a new feature.
If we do add a flag, here are some options:
`_df` (I think that wouldn't be very search friendly)
`_fault`
`_dropped_frames`
`_flagged`
etc.

dvida:
The reason why I removed it is because all file names should be consistent and identical. The meteor stack isn't but that was a mistake made early on. However, I would fully support the addition of a report file with the numbers and a list of all intervals. If that is implemented, I suggest having both the read and write functions available.

The idea is then that the weblog code can go in and read in a bunch of info and generate a report and show a camera status summary.

Cybis320:
That makes sense. I don't know anything about the weblog so that's probably not something I'd be comfortable doing unless there's an example I can follow.

g7gpr:
Still got a problem with an edge case, on a directory which contains no .fits files. This can happen if a camera power supply fails, which happens from time to time. 

```
(vRMS) au000u@pioneer:~/source/RMS$ git log -1
commit ea699217c524463dd413a4a730bcff9b44d848aa (HEAD -> intervals, cybis320/intervals)
Author: Denis Vida <denis.vida@gmail.com>
Date:   Thu Feb 8 11:56:53 2024 -0500

    Added residuals plot
(vRMS) au000u@pioneer:~/source/RMS$ python -m Utils.PlotTimeIntervals ~/RMS_data/ArchivedFiles/
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240123_114744_582540/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240123_114744_582540/FS_AU000U_20240123_114744_582540_fieldsums.tar.bz2
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240125_114644_582696/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240125_114644_582696/FS_AU000U_20240125_114644_582696_fieldsums.tar.bz2
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240204_114011_242936/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240204_114011_242936/FS_AU000U_20240204_114011_242936_fieldsums.tar.bz2
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240201_114224_817694/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240201_114224_817694/FS_AU000U_20240201_114224_817694_fieldsums.tar.bz2
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240124_114714_747828/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240124_114714_747828/FS_AU000U_20240124_114714_747828_fieldsums.tar.bz2
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240131_114307_048976/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240131_114307_048976/FS_AU000U_20240131_114307_048976_fieldsums.tar.bz2
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240203_063937_481003/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240203_063937_481003/FS_AU000U_20240203_063937_481003_fieldsums.tar.bz2
/home/au000u/vRMS/lib/python3.9/site-packages/numpy/core/fromnumeric.py:3432: RuntimeWarning: Mean of empty slice.
  return _methods._mean(a, axis=axis, dtype=dtype,
/home/au000u/vRMS/lib/python3.9/site-packages/numpy/core/_methods.py:190: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
/home/au000u/vRMS/lib/python3.9/site-packages/numpy/core/_methods.py:265: RuntimeWarning: Degrees of freedom <= 0 for slice
  ret = _var(a, axis=axis, dtype=dtype, out=out, ddof=ddof,
/home/au000u/vRMS/lib/python3.9/site-packages/numpy/core/_methods.py:223: RuntimeWarning: invalid value encountered in divide
  arrmean = um.true_divide(arrmean, div, out=arrmean, casting='unsafe',
/home/au000u/vRMS/lib/python3.9/site-packages/numpy/core/_methods.py:257: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/au000u/source/RMS/Utils/PlotTimeIntervals.py", line 297, in <module>
    plotFFTimeIntervals(root, fps)
  File "/home/au000u/source/RMS/Utils/PlotTimeIntervals.py", line 157, in plotFFTimeIntervals
    label='Intervals, max ({:.3f}s), min ({:.3f}s)'.format(max_interval, min_interval), 
![AU000U_20240123_114744_582540_ff_intervals](https://github.com/CroatianMeteorNetwork/RMS/assets/77130399/62a8252b-1f35-47ec-aba3-f59b6e9ae824)

TypeError: unsupported format string passed to NoneType.__format__
```
The chart are looking fantastic

AU000U_20240123_114744_582540_ff_intervals

AU0028_20240204_114635_244720_ff_intervals

Cybis320:
> Still got a problem with an edge case, on a directory which contains no .fits files. This can happen if a camera power supply fails, which happens from time to time.
My bad, I should have handled those cases. When you have a chance could you try this fix. Thank you!

g7gpr:
Looking good. 

```
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240203_063937_481003/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240203_063937_481003/FS_AU000U_20240203_063937_481003_fieldsums.tar.bz2
Insufficient timestamps. At least two timestamps are required.
Loading config file: /home/au000u/RMS_data/ArchivedFiles/AU000U_20240130_114346_974032/.config
Processing /home/au000u/RMS_data/ArchivedFiles/AU000U_20240130_114346_974032
```

AU0028 processed with no problems. Thank you for the rapid fix. It is difficult for me to test for the next few hours as the uploads are taking place, but I will report back 12 hours from now after rolling out on au000u-z and au0028-f


Dave

g7gpr:
No problems with any of test directories. Would you consider adding a check to make sure that the specified directory exists?

Cybis320:
> No problems with any of test directories. Would you consider adding a check to make sure that the specified directory exists?

Done.

Also, I'm curious about what the resolution will be regarding the misbehaving machines. Are the issues due to slow/worn-out SD cards, or file system corruption? And if so, I'm wondering how the GStreamer branch (PR# 257) would behave on such machines. I'm concerned that the GStreamer appsink might swell up and then suddenly flush past a certain threshold, leading to large gaps in data. On the other hand, GStreamer handles temporary slowdowns without impacting timestamps, so maybe it could operate with degrading SD cards longer? (I'm struggling to make the GStreamer appsink behave less radically. Ideally, the sink should just drop frames piecemeal to prevent blowing up.)

dvida:
BTW, I think the accuracy with gstreamer is even better, but it's currently limited by the time precision in the FF file name. We should add the more accurate time to the FF fits header with the accuracy of microseconds.

Cybis320:
> BTW, I think the accuracy with gstreamer is even better, but it's currently limited by the time precision in the FF file name. We should add the more accurate time to the FF fits header with the accuracy of microseconds.

You're right that we a losing a little bit by rounding to the mills. Here's a plot with the minimum scale disabled:
US9999_20240208_013640_169265_ff_intervals

dvida:
BTW, do you have an idea why the median is always so consistently above the expected?

Cybis320:
> BTW, do you have an idea why the median is always so consistently above the expected?

I'm guessing the camera's clock is misconfigured? The offset is pretty consistent across cameras, ~1 second per 20 minutes. It's consistent btw CV2 and Gst.  It's exactly one frame per 5x 256 block run... hmm.

Cybis320:
There is a NTSC standard for 23.976 fps. I wonder if the camera is adjusting the fps relative to that, 24.976 is exactly what we have. Maybe setting the camera to PAL will solve that?

dvida:
We do recommend setting the camera to PAL already, check if yours is set to that. All other GMN cameras should already be set to PAL.

Cybis320:
> We do recommend setting the camera to PAL already, check if yours is set to that. All other GMN cameras should already be set to PAL.

I just checked and it was on PAL. However, setting it to NTSC or PAL does not affect the intervals. Still ~24.98 fps.
At 20 fps I get ~19.98.

dvida:
Hmmm... I wonder if it's an effect because we have two buffers and each perhaps gets timestamped differently? If we compute the time diff between ever other FF file, do we get the right number?

Cybis320:
> Hmmm... I wonder if it's an effect because we have two buffers and each perhaps gets timestamped differently? If we compute the time diff between ever other FF file, do we get the right number?

I just measured the fps independently of the compression and I'm getting the same. 24.982.

satmonkey:
NTSC fps is 29.97. I think they use the 5/6 divider for PAL to maintain the compatibility.
This yields 24.98 fps for PAL.
https://steinberg.help/wavelab_pro/v10/en/wavelab/topics/video/video_frame_rates_c.html

dvida:
Well, I'll be damned... Milan, are you seeing the same FPS with your gstreamer setup?

Cybis320:
There's more going on. I need to graph the timestamp of each frame over time but there something non-linear happening. It looks like it's maintaining 25.0000 fps for ~27sec then dips for a few seconds before going back to exactly 25, rinse, repeat.

Cybis320:
And same at 24fps. 24.0000 then dip.

dvida:
Are you running the camera at 1080p? Have you tried 720p? I remember we had FPS issues at 1080p.
Also, do you think it would be worth extending the FF format with per-frame timestamps? That way we can avoid any FPS issues.

Cybis320:
The camera is running on a 1/90000 sec heartbeat would that have anything to do with anything. 

720p right now.

Cybis320:
It's a clock synching issue. By setting buffer-mode to only use RTP timestamps (buffer-mode=0) or by synchronizing sender and receiver clock (buffer-mode=4), the fps is a steady 25.0000. I need to check how either of these options impact absolute time accuracy.

Cybis320:
Ok, it's even weirder than I imagined. First, changing buffer-mode only moves the problem down the line. I.e. when pinning the fps to 25, it causes the pts to accumulate errors over time.

Here's the really weird thing: what appears to be going on is that the camera is actually recording at 29.97 fps and dropping a 1 out of every 6 frames. I recorded a spinning disc with a dot (on a drill). The dots are evenly spaced, except every 6th dot is missing:
```
***** ***** ***** *****
```
There is on average 24.98 dots per seconds, so including the missing dot... ta-da 29.97 fps.

Cybis320:
> Ok, it's even weirder than I imagined. First, changing buffer-mode only moves the problem down the line. I.e. when pinning the fps to 25, it causes the pts to accumulate errors over time.
> 
> Here's the really weird thing: what appears to be going on is that the camera is actually recording at 29.97 fps and dropping a 1 out of every 6 frames. I recorded a spinning disc with a dot (on a drill). The dots are evenly spaced, except every 6th dot is missing:
> 
> ```
> ***** ***** ***** *****
> ```
> 
> There is on average 24.98 dots per seconds, so including the missing dot... ta-da 29.97 fps.

Apologies for the confusion, I had left the camera in NTSC for that test. In PAL, it doesn't skip frames. Phew. 
I learned that NTSC unlocks 30 fps though.

g7gpr:
It was weird, I have stacked frames of long lasting meteors, the angular distance moved per frame is roughly constant.

This is brilliant work though. Thanks for what you are doing.

Kind regards

David Rollinson


I am writing this on a phone, apologies for brevity and errors.

satmonkey:
this is the graph from pi4 buster
CZ0002_20240214_165221_506833_ff_intervals_flagged

satmonkey:
by the way, I can see some artefacts on the captured stack, coming from 3 FF files. This did not happen before with ffmpeg on that station. Will monitor this during a few upcoming nights and report then.
CZ0002_20240214_165221_506833_ff_intervals_flagged

Cybis320:
> by the way, I can see some artefacts on the captured stack, coming from 3 FF files. This did not happen before with ffmpeg on that station. Will monitor this during a few upcoming nights and report then.
> 
> CZ0002_20240214_165221_506833_captured_stack
> 
> 

By the look at the plot and the gaps in the trails, there is a lot of dropped frames. I would guess there's likely a hardware issue, no?

satmonkey:
from the reports, this does not seem so.
Fine, I will make another test on another pi4 Buster.

satmonkey:
ahh you mean those landing/taking off plane trails? That's 5 km far from the LKPR runway 24.

Cybis320:
And that was with the dev branch which has the same capture mechanism as master?

satmonkey:
I just checkout the dev and put force_cv2 into .config.

Cybis320:
> ahh you mean those landing/taking off plane trails? That's 5 km far from the LKPR runway 24.

Nice! I use to fly into PRG back in the days. And, yes, the aircraft light trails are showing gaps. Can I ask what SD cards are you using?

satmonkey:
I see now the BufferedCapture patch was not merged into dev yet, so yes, still the old method.

Cybis320:
I get 33.0 MB/s for write and 45.3 MB/S for read on a RPi4 Bookworm with a Sandisk Extreme Pro SD card.
```
(vRMS) pi@rpi4-cam:~/source/RMS $ dd if=/dev/zero of=tempfile bs=1M count=1024 conv=fdatasync,notrunc
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 32.5159 s, 33.0 MB/s
```

```
(vRMS) pi@rpi4-cam:~/source/RMS $ sudo /sbin/sysctl -w vm.drop_caches=3
vm.drop_caches = 3
(vRMS) pi@rpi4-cam:~/source/RMS $ dd if=tempfile of=/dev/null bs=1M count=1024
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 23.7063 s, 45.3 MB/s
```

Cybis320:
For fun: write: 466 MB/s and read: 384 MB/s with RPi5 with NVMe SSD:
```
root@rpi-cam1:~# dd if=/dev/zero of=tempfile bs=1M count=1024 conv=fdatasync,notrunc
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 2.79917 s, 384 MB/s
```
```
root@rpi-cam1:~# dd if=tempfile of=/dev/null bs=1M count=1024
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 2.30412 s, 466 MB/s
```

Cybis320:
Here's where I'm at with the frame rate. Even in PAL, the actual fps is ~24.982, and each frame is very evenly spaced with no significant variations (there may be some tiny variation over long periods.)

Unfortunately, the rtspsrc element is expecting exactly 25 fps, and is always attempting to go back to 25 fps. This causes the slight dips when the error reaches about 1 ms and the pts algorithm more or less smoothly adjust the pts to zero the error. It does this in a ~27s cycle with an amplitude of ~3ms. I don't see an easy way to tell the first element of the pipeline to expect 24.982 fps. It might be "easier" to write a pts smoothing algorithm in python if we want sub ms precision.

For the FF files, the first frame timestamp should be accurate to ~10ms (given the start time uncertainty and the pts wandering). Assuming 24.98 fps, the 256 frames within a FF file should maintain that accuracy (so, I don't think we need to timestamp each frame). I haven't looked but I assume the config fps is used to calculate intra FF ts? If so, should we consider changing the config fps to 24.98, or leave it... we are talking about 0.4ms over 256 frames...

g7gpr:
There are many places in RMS where small factors have been included and corrected for. I think we should find a way to include your findings.

dvida:
I'm seeing this on one Pi3 camera, does anyone else see something similar? I can share the data files via email.
image
These are produced automatically during normal processing.

PR #263, opened 2024-01-31
small change to SkyFit2 to save the FF name in the state file.

### LINKED ISSUE #262: State file doesn't contain FF name
The state file saved by SkyFit2 does not contain  the name of the FF file. 
The file name IS included if its an FR file but not if its an FF. 
This causes a problem when more than one FF file is present, for example bracketing FFs to allow better platepar refinement, or when the fireball is split across two FF files.

markmac99:
I've amended skyfit2 to include the FF filename in the state file. 
Now, if you run SkyFit2 on a folder containing more than one FF file, you can either
a) use the folder as before, move between FF files and then work on the one you want to use; or
b) load the state file. In this case, the right FF file is automatically loaded for further analysis

I've done some testing with fireballs without any issues, but would appreciate further review / input.

Closes #262

### DIFF
--- a/Utils/SkyFit2.py
+++ b/Utils/SkyFit2.py
@@ -2162,9 +2162,11 @@ class PlateTool(QtWidgets.QMainWindow):
         # Currently, any important variables should be initialized in the constructor (and cannot be classes
         # that inherit). Anything that can be generated for that information should be done in setupUI.
 
+        img_name = self.img_handle.name()
         to_remove = []
 
         dic = copy.copy(self.__dict__)
+        # print('input path', dic['input_path'])
         for k, v in dic.items():
 
             if (v.__class__.__bases__[0] is not object) and (not isinstance(v, bool)) and \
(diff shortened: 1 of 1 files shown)


satmonkey:
almost perfect!
however, when using the --fr switch, and starting the fireball reduction on the FR file, restoring the state jumps to the FF file, not FR. It allows to resume the reduction though, but this is a minor side effect.

dvida:
Great! Have you tried opening old state files that were generated with previous version of skyfit?

markmac99:
@dvida I think so, but i will double check just in case.

satmonkey:
does not work with old state files here. Takes the 1st FF file as default.

dvida:
Good, as long as it doesn't throw an error. I'll now do a final test on non-RMS data and merge.

markmac99:
@dvida confirmed that when i load an older state file, the programme behaves as previously and loads the 1st available FF file. 

This means that could be the wrong image for the state, but this is as expected since the correct filename isn't stored in the state.

satmonkey (approved):

markmac99:
@satmonkey yea thats as expected because the ff name is not available in the state :) so the behaviour is as it was previously (all that can be done).

satmonkey:
I got it :) good job

dvida:
Sorry to reopen this - has anyone tried moving the data directory to a different path? Will this new feature work then? E.g. the use cause might be when sharing the reduction with someone.

satmonkey:
yes, this works with me, too

PR #269, opened 2024-02-15
Add meteor to end of bin video - fix for multi meteors on same fits file

aitov:
This is a fix for cases when we have several tracked meteors on on same fits file and identification was wrong as it took first meteor from list

Implemented solution:
Usage of meteor start and end time to compare with start and time of bin video (calculated by frame number)
If meteor time range inside video time range - this is the best candidate.

Also during testing found that video could start with small lag (1 or 2 frames later) .
To fix it was introduced time offset for 100mls before and after video start (if fps is 25 then 1 frame ~40mls )
It fixed issue on my test suite.

What was not tested:
As  I can see bin file could be multi lines and I don't have such example and don't know when such file could be created, also exists option '--split' which also could affect calculation.

This solution will not work for simultaneous two or more meteors at same time but I think it is very rare case and need identification meteor by coordinates on video (very complex for me).

### DIFF
--- a/Utils/FRbinViewer.py
+++ b/Utils/FRbinViewer.py
@@ -33,7 +33,7 @@ import datetime
 from glob import glob
 from Utils.ShowerAssociation import showerAssociation
 from RMS.Formats.FTPdetectinfo import validDefaultFTPdetectinfo
-
+from RMS.Astrometry.Conversions import datetime2JD
 
 def view(dir_path, ff_path, fr_path, config, save_frames=False, extract_format=None, hide=False,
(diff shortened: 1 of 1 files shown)


g7gpr:
Is this code intended to add the name of the meteor shower to the end of a video?

aitov:
> Is this code intended to add the name of the meteor shower to the end of a video?

Hi, 
this code is improvement of previous pull request #256 which adds shower name to full video by parameter : "-w", "--add_shower_name" (even without meteor image at the end)

It fixes shower name identification in case if image contains more than 1 meteor

markmac99 (approved):
Looks good, passes all my tests anyway :)

PR #277, opened 2024-03-19
Add support for microseconds in FS filenames and FF headers

Cybis320:
This PR is related to PR #257. Because PR #257 improves frame timestamping and makes microseconds significant, this PR preserves the microseconds information instead of rounding to milliseconds. Due to compatibility concerns with downstream processes, the FF filename structure remains unchanged, still displaying only milliseconds. To retain microseconds, a new **EXPSTART** header field is added to store the datetime string with microseconds:
```
SIMPLE  =                    T / conforms to FITS standard                      
BITPIX  =                    8 / array data type                                
NAXIS   =                    0 / number of array dimensions                     
EXTEND  =                    T                                                  
NROWS   =                 1080                                                  
NCOLS   =                 1920                                                  
NBITS   =                    8                                                  
NFRAMES =                  256                                                  
FIRST   =               704768                                                  
CAMNO   = 'US9999  '                                                            
FPS     =            24.981216                                                  
EXPSTART= '20240319_095911_310820'                                             
END                                                                             
```

FS files cannot store microseconds in their header. However, since there doesn't seem to be compatibility issues downstream, microseconds are included in the filename:
`FS_US9999_20240318_011711_371924_1054208_fieldsum.bin`

### DIFF
--- a/RMS/Compression.py
+++ b/RMS/Compression.py
@@ -118,10 +118,14 @@ class Compressor(multiprocessing.Process):
         date_string = time.strftime("%Y%m%d_%H%M%S", time.gmtime(startTime))
 
         # Calculate miliseconds
+        micros = int((startTime - floor(startTime))*1000000)
         millis = int((startTime - floor(startTime))*1000)
         
 
-        filename = str(self.config.stationID).zfill(3) +  "_" + date_string + "_" + str(millis).zfill(3) \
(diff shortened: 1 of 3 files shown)


dvida:
Perfect - I see that this PR has been set as "draft", are you planning to do some more work on it or is it good to merge?

Cybis320:
Only a draft because I'm afraid of commitments :)

PR #278, opened 2024-03-25
FITS Handling: Introduce Lossless Compression for HDUs

Cybis320:
This PR introduces lossless compression algorithms for Header/Data Units (HDUs) within our FITS file handling module (**'RMS/Formats/FFfits.py'**). The goal is to reduce FITS file sizes to improve storage efficiency, mitigate wear and tear on storage devices, and enhance data transfer speeds, all while maintaining the data integrity and accessibility.

**Changes Made:**
- Implemented RICE_1 compression for data HDUs (maxpixel, maxframe, avepixel, stdpixel).

**Impact:**
- Expected reduction in FITS file sizes by approximately 30-60%, significantly saving disk space, reducing wear and tear on storage devices,  and improving data transfer times.
- The size of compressed 1080p FITS is expected to be similar with that of uncompressed 720p FITS.
- Machines with marginal storage device performance are expected to benefit from the smaller file size.
- CPU load is expected to increase slightly. 

**Testing:**
- Conducted compatibility testing on Rpi3, 4, and 5. 
- Tested decompression in CMNbinViewer


g7gpr:
Is this expected to be safe to run on stations that will upload, or should upload be turned off when testing this?

markmac99:
can you please share some example compressed files ? I'd like to check they're still compatible with other FITS viewing and analysis software,

Cybis320:
Here are four FF fits files. Unfortunately, It's cloudy here right now. It would interesting to see what the size will be on a starry night - the size reduction might not be as dramatic. 

1. 1080p Uncompressed: 8.3 MB
2. 1080p Compressed: 3.4 MB
3. 720p Uncompressed: 3.7 MB
4. 720p Compressed: 1.7 MB

[
[Archive 2.zip](https://github.com/CroatianMeteorNetwork/RMS/files/14752280/Archive.2.zip)
]

dvida:
Dave, I suggest testing this separately until we can confirm that there are no effects on processing.

markmac99:
As noted by email i have found that the compressed files are incompatible with other FITS handling software - both FITS Liberator and Pixinsight refuse to open them anyway (they actually crash FL!). For me this feels like an issue as we don't know what the files might be getting used for downstream of RMS.

I'm also wondering whether compression is really beneficial. Storage is cheap and 3.5MB is not really that big and the data are bzip compressed for upload to GMN so I dont think it will save space on the GMN side or save time in uploads. I agree it means less data getting written to disk each night, but the lifespan of any decent SD card is pretty long (several years) and i've actually never had a card fail due to wear and tear.

Cybis320:
On my end, CMNbinViewer, SAOImageDS9, and FITS Liberator have no issues handling it. I don't have access to Pixinsight but Pixinsight stated 'we have no interest in FITS, which we deprecated many years ago in PixInsight'.

Is anyone else having any issues? It would be strange for a Fits application to not support such a basic requirement.

Mark, you're making a very good point about the data being compressed before upload. The compressed and the uncompressed fits files, once compressed in a tar.bz2, have similar sizes. So, there is indeed no benefit as far transmitting the files. I don't know how the tar files are handled on the receiving end. 

Regarding local storage, it would clearly be beneficial. So, I think it would be valuable to get to the bottom the compatibility issue before pulling this out.

Luc

> As noted by email i have found that the compressed files are incompatible with other FITS handling software - both FITS Liberator and Pixinsight refuse to open them anyway (they actually crash FL!). For me this feels like an issue as we don't know what the files might be getting used for downstream of RMS.
> 
> I'm also wondering whether compression is really beneficial. Storage is cheap and 3.5MB is not really that big and the data are bzip compressed for upload to GMN so I dont think it will save space on the GMN side or save time in uploads. I agree it means less data getting written to disk each night, but the lifespan of any decent SD card is pretty long (several years) and i've actually never had a card fail due to wear and tear.

markmac99:
Pixinsight will keep supporting FITS for a long time, although the pixinsight guys have been pushing their own private format for years. Nobody else is interested in it! My point was that PixInsight is a -very- widely used tool in the astro-imaging world, and if it can't open the files then its a problem. 

i think the problem with Fits Liberator is that there are two versions: version 3 is distributed by ESA from their website and can't open the compressed files. Version 4 from noirlab can handle the files. 

Personally i actually don't agree there's much advantage in compressing the files, because storage is cheap and the days when 3-4MB was a big file are long gone. Saving 1-1.5MB per file isn't really very significant. I realise it'd mean we could keep the CapturedFiles data for a bit longer but its very rare that we need to look back more than a few days, and realistically we'd only be able to keep an extra day or so.  

Anyway for me, i would want this to be an optional feature.

g7gpr:
OK, will do.

Cybis320:
I added a '**hdu_compress**' configuration setting to `.config` [Compression] section (defaults to False). And I made a command line utility to convert all FITS files in a directory to uncompressed FITS. `RMS/Utils/ConvertCompressedFits.py`

markmac99:
will try it out !

Cybis320:
Update on compression ratio: my camera sees stars tonight and the compression still reduces the size by half.

dario-zubovic:
Worth noting that AstroPy also supports full-file compression with gzip, zip or bzip2. Such files have `.fits.gz`/`.fits.zip`/`.fits.bz2` extension.

IMO it should be preferred approach:
- files can at glance be recognized as compressed from extension alone
- all AstroPy-based tools will seamlessly accept such files, or alternatively, off-the-shelf zip decompressor can be used / no knowledge of fits internals is required
- I have serious doubts that "RICE_1" can achieve comparable compression ratio

As for justifiably question, I would agree that reducing size of individual files will be helpful. Besides retaining more nights on local storage before they're cleaned, it would also help will lifespan of SD cards, as they can only survive a limited number of write cycles.

markmac99:
As i've said earlier i do not see any real benefit to compressing the FITS files, and so would prefer this to be an option and not the default. 
I understand that it'd save a bit of space but as I've noted before diskspace is cheap and its rare  that we need to recover raw data from more than a few days ago, especially now we have the EventMonitor in place and can recover data in near-realtime without camera-operator intervention. I understand the point about less wear and tear, but in my experience this isn't really a significant problem plus many owners are migrating to linux and/or SSD which probably mitigates that issue anyway. 

Finally bear in mind that RMS is not the only app that reads/writes the FITS files. If we change the file format it could have unexpected impact on how end-users process the data in ways we cannot guess because we do not know what downstream processes the individual countrry networks have in place (we in UKMON don't use the FITS files, but its possible others do). I'm sure it'd be a simple fix, but i am never in favour of introducing incompatability.

adalava:
In my use case, reduce storage size in ~50% is benefical. Storage isn't cheap here, and it would allow us to use 64GB or even 32GB microSD cards on the Pis. 

It seems that required I/O write throughtput will be reduced as well, so we will be able to add more RMS instances on a x86_64 server, at cost of higher CPU usage. 

I really like this patch. But I'd suggest you to collect more data. For instance, I'd like to see how extra CPU usage will affect the PIs, specially the Pi3 as they are at maximum load already, before making it default. The compatibility with software like Mark said is also important. Some people would prefer to have compression disabled for now, so they need to be aware and choose what to. I think most people don't use any external software, so I'm in favor in having it as default as long it doesn't break any existing RMS station.

I'll be happy to test it on BR0001, BR0002 and BR0003

dario-zubovic:
To clarify what I meant above; it's not required to change file format to get benefits of compression. As such, I would be argue against this patch with current implementation of HDU-only compression.

Overall idea makes sense though. AstroPy can directly write to disk files with current FITS format, but already inside a generic compressed archive. Ie. best of both worlds, no changes to underlying format and smaller individual files, It will result in better compression ratio as well. I can open a branch if anybody is curious to test full-file compression?

adalava:
> To clarify what I meant above; it's not required to change file format to get benefits of compression. As such, I would be argue against this patch with current implementation of HDU-only compression.
> 
> Overall idea makes sense though. AstroPy can directly write to disk files with current FITS format, but already inside a generic compressed archive. Ie. best of both worlds, no changes to underlying format and smaller individual files, It will result in better compression ratio as well. I can open a branch if anybody is curious to test full-file compression?

Hm! Yes, It's a good balance if it can write the compressed file directly (and not by writting the full uncompressed file, then compress it and delete the full file).
Also would be great if the other tools like skyfit, bin viewer, stars detection and so are able to uncompress it directly to the RAM, without expanding it to the disk first. 

I would like to test it as well!

markmac99:
> > To clarify what I meant above; it's not required to change file format to get benefits of compression. As such, I would be argue against this patch with current implementation of HDU-only compression.
> > Overall idea makes sense though. AstroPy can directly write to disk files with current FITS format, but already inside a generic compressed archive. Ie. best of both worlds, no changes to underlying format and smaller individual files, It will result in better compression ratio as well. I can open a branch if anybody is curious to test full-file compression?
> 
> Hm! Yes, It's a good balance if it can write the compressed file directly (and not by writting the full uncompressed file, then compress it and delete the full file). Also would be great if the other tools like skyfit, bin viewer, stars detection and so are able to uncompress it directly to the RAM, without expanding it to the disk first.
> 
> I would like to test it as well!

Would be interested to test it too. 
Agree we should not change the file format. I'd also argue strongly for using a common compression algo like gzip or zip to ensure portability across platforms. I know there are better algos, but they are sometimes not supported across windows, macos, and various linux flavours. 
Definitely needs to be tested on Pi3 as it might require too much memory or processor capacity. While the pi3 is now very old there are still a lot of RMS stations using them.  I don't have any working Pi3 station however :(

Cybis320:
Bumping this discussion. I think we need to address FITS file sizes as they're impacting the science GMN can perform.

Storage size directly affects how many nights of data we can keep on disk for revisiting interesting events. Currently, without compression, we can only go back half as many days as we could with compressed storage.

**Current storage breakdown per 12hr night:**
- CapturedFiles: ~18GB (would be ~9GB compressed)
- Raw_video (24hr): ~20GB  
- FrameFiles (24hr): ~0.1GB

This is particularly problematic for multi-cam setups where storage quickly becomes the limiting factor.

This isn't just an upload bandwidth issue (files are already compressed for transfer). The real impact is on **data retention for analysis**.

All compression schemes (RICE, gzip, etc.) achieve roughly the same ~50% ratio. We just need to pick one and implement it.

**Option 1: Compress individual FF files**
- Pros: Convenient, direct file handling
- Cons: Reduced compatibility with very old Pixinsight version.

**Option 2: Compress entire night directories** 
- Pros: Preserves compatibility
- Cons: Less convenient for accessing individual files

Both could be made optional via config flag.

**Thoughts on implementation approach? Any strong preferences between the options?**

g7gpr:
If you compressed an entire directory, how hard would it be to retrieve an individual file? I know some compression schemes are clever, but it feels expensive. 

My own feeling is compress individual fits files from the previous night just before capture on the next night is due to start.

I don't agree that this problem affects multi-camera stations more than single camera stations. I think the greatest benefit will come from doubling the recall ability of the many 128 GB SD card stations.

markmac99:
I just want to reiterate that i still don't think this is the right approach. Storage is cheap, and it'd be a lot simpler, less risky and more compatible just to keep data uncompressed and advise buying a larger SD card.  As I noted upthread my quick tests indicate that any compression will make the data less compatible with other tools, which will make it harder for camera operators to examine and play with the data themselves. i think that ability is really important as it keeps people enaged and interested. 

I also believe we can solve the problm by making sure that all new non-core capabilities, such as all-day timelapses, raw video capture, daytime monitoring, contribution to contrail monitoring etc are disabled by default and come with a clear caveat that they'll use more storage so either a 256GB SD card or an SSD will be required.  Existing station owners would then be unaffected and anyone enabling the new features would understand the risks and impact.  I do worry that we're seeing a lot of "mission creep" thats impacting camera owners without fully informing them. 

I also think this would work - I might be missing something here, but I have disabled all unnecessary features and my stations can still retain 9-10 days of data, which is the same as it was back in 2022. So i do think this approach would obviate the need to make the data less compatible. 

If we do decide to compress data, I'd strongly recommend creating a standalone service that could be run as via systemd or be triggered by RMS using a signal. I feel it should not be done within RMS itself as this will create another thread and dependency that could fail or get stuck, leading to unexpected behaviour or data loss. Its a best-practice antipattern to make monolithic apps.

Cybis320:
Given recent developments - rising storage costs, the introduction of raw video save, and upcoming plans for daytime FITS recording - I'd like to revisit optional lossless compression for HDUs.

Currently, FITS files are a fixed size regardless of content. They compress well before upload, but consume significant local storage prior to tar archiving. Daytime frames in particular would compress exceptionally well.

Is there still resistance to making HDU compression optional? Maybe separately configurable for day vs. night capture?

dvida:
Just to give everyone some context - @Cybis320 and I discussed using compressed FFs for daytime recording. This will enable people to turn this feature on with very little impact on storage, as daytime data will compress really well. And this data will not be used operationally as any daytime events will have to be manually analyzed. So this will enable easy "better than nothing" data collection. And voila, the GMN is then recording daytime fireballs with little to no impact on resources.

markmac99:
In terms of space its clearly a useful thing. My concern, as before, is that many downstream FITS tools can't handle them eg as noted previously Pixinsight 1.8.9 and FITS Liberator 3.  There are probably only a few people who routinely use these tools on the data though. 

However, I'm a bit unclear about the plan. 
Currently i think daytime data is stored in jpeg or png format, correct? 
So would this proposal switch to FTP format FITS files, but compressed?  
Would that apply to both daytime and nighttime files?  
Will the FITS files be larger than jpegs or png?  

Above i tihnk luc mentioned needing 1.1GB plus 0.1GB per day, is that correct? 

Sorry to ask a lot of questions

dvida:
Right, I haven't explained things all too well. Currently, the daytime recordings only consist of the frame files and are quite tiny. The only way to get daytime fireballs is to record the full MKV videos, which comes with a big hit in terms of disk space.

The idea is to develop a new (optional) feature which will allow saving FFs during the day in the compressed format, and also run the fireball detector. This way, we can get away with daytime fireball monitoring with only a small impact on disk space.
Yes, the daytime compressed fits will only be analyzable in RMS tools, but it's better to collect some data than none.

markmac99:
understood, makes sense. I agree its better to collect some data than none! 

Will be keen to see the size of the compressed fits files compared to the current frames files.

Cybis320:
To clarify, the day FF files would be generated alongside the JPG ‘frames’ file, not as a replacement - they serve distinct purposes (Fireball/reentry vs climate)

markmac99:
Okay definitely need to understand the space impact. Looks like the frames files take a min of 1GB, plus extra  GB for compressed fits, will we need to recommend 256GB sd cards for pi based systems? Otherwise we'll lose night time data on recent fireballs.

Utils/ConvertCompressedFits.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def findFitsFiles(directory):
+    """Recursively find all FITS files in a directory and its subdirectories.
+
+    Arguments:
+        directory: [str] The path to the folder containing the files to convert.
+
+    Return:
+        None
```
g7gpr: Does not match actual return

Utils/ConvertCompressedFits.py
```diff
@@ -0,0 +1,99 @@
...
+
+    """
+    fits_files = []
+    for root, dirs, files in os.walk(directory):
+        fits_files.extend(glob.glob(os.path.join(root, '*.fits')))
+    return fits_files
```
g7gpr, line 44: Does not match docstring

Utils/ConvertCompressedFits.py
```diff
@@ -0,0 +1,99 @@
...
+            print(f"Failed to convert {fits_file}: {e}", file=sys.stderr)
+
+
+if __name__ == "__main__":
+
+    ### COMMAND LINE ARGUMENTS
+
+    # Init the command line arguments parser
+
+    arg_parser = argparse.ArgumentParser(description="Convert all compressed FITS files in a directory (and "
```
g7gpr: Do we have a way to reverse this operation, or will it happen automatically?

g7gpr (commented):
The console call to recursively decompress is good, but is there a way to recompress the files? From the console would be ideal. Or will RMS recurse through the directories, compressing them?

Will the compressed files work with SkyFit2, or does SkyFit2 need some work so it can open compressed files?

g7gpr:
Would you like to me to switch some stations to this branch?

dvida:
@markmac99 We still don't know what the requirements are, but this definitely won't run by default. We're making it an option and we'll do a bit of exploration first.

@g7gpr The daytime FF saving is still not implemented, and I'd rather not the normal FFs to be saved in the compressed format just yet.

To answer the questions whether SkyFit will need any work - if the reader in FFfile.py transparently works with the compressed format, then no. I belive that it does.

Cybis320:
Here’s an initial attempt to estimate the storage impact of different options. I haven’t tested the actual compression ratio on day FITS. I still need to consider the quota management scenario - this is with quota management disabled. I've assumed the default retention policy in .config, including 20 day retention for the hypothetical day FF files (same as night - which may not make sense.)

## Capture Options Summary

  | Option              | Description                                           |
  |---------------------|-------------------------------------------------------|
  | continuous_capture  | Run capture 24/7 instead of only at night             |
  | switch_camera_modes | Switch camera settings between day/night |
  | save_frames         | Save individual JPG frames at intervals           |
  | raw_video_save      | Save raw H.264 video segments                         |
  | save_frame_times    | Save FT (Frame Time) timestamp files                  |

## Data Products by Mode

  | Data Product                    | Science.                  | Night (standard)       | Night (continuous)     | Day (continuous)      |
  |---------------------------------|---------------------------|------------------------|------------------------|-----------------------|
  | FF files (256-frame compressed) | Meteor                    | ✅                     | ✅                     | ❌ (planned)           |
  | FR files (fireball raw frames)  | Meteor                    | ✅                     | ✅                     | ❌ (planned)           |
  | Detection (meteors/stars)       | Meteor                    | ✅                     | ✅                     | ❌                     |
  | FT files (timestamps)           | Meteor                    | ✅ if save_frame_times | ✅ if save_frame_times | ✅ if save_frame_times |
  | Saved frames (JPG/PNG)          | Climate                   | ✅ if save_frames      | ✅ if save_frames      | ✅ if save_frames      |
  | Raw video (H.264 segments)      | Meteor                    | ✅ if raw_video_save   | ✅ if raw_video_save   | ✅ if raw_video_save   |
  
---
## Storage Impact by Option (720p @ 25 fps)


  ---
### Per-Hour Storage

  | Data Product           | Size/Unit     | Per Hour   | 10h Night |  Per 24h  |
  |------------------------|---------------|------------|-----------|-----------|
  | FF files (FITS current)| ~3.5 MiB each | ~1.2 GB    | ~12 GB    | ~30 GB    | (Accumulates until deleted)
  | FF files (FITS comp.)  | ~1.7 MiB each | ~0.6 GB    | ~6 GB     | ~14 GB    | (Accumulates until deleted)
  | FR files (fireballs)   | Variable      | ~10-100 MB | ~0.1-1 GB | ~0.1-1 GB | (Accumulates until deleted)
  | FT files (timestamps)  | ~3 KB/block   | ~1 MB      | ~10 MB    | ~0.02 GB  | (Accumulates until deleted)
  | Saved frames (JPG q90) | ~125 KB each  | ~90 MB     | ~900 MB   | ~1.8 GB†  | (Max Fixed Overhead)
  | Timelapse fr JPG + Tar | ~80 MB each   |            | ~80 MB    | ~0.16 GB  | (Accumulates until deleted)
  | Raw Video @ 4096 k      |               | ~0.9 GB    | ~9 GB     | ~22 GB    | (Accumulates until deleted)

  †Reach max value after 24 hr and then fluctuates between 0 and max.

  ---
### Approximate Raw Video Storage (H.264 @ 50% of nominal bitrate)

  | Nominal Bitrate | Effective Rate | Per Hour | 10h Night | 24h Continuous |
  |-----------------|----------------|----------|-----------|----------------|
  | 2048 kbps       | 1024 kbps      | 0.45 GB  | 4.5 GB    | 10.8 GB        |
  | 4096 kbps       | 2048 kbps      | 0.9 GB   | 9 GB      | 21.6 GB        |
  | 8192 kbps       | 4096 kbps      | 1.8 GB   | 18 GB     | 43.2 GB        |
  | 16384 kbps      | 8192 kbps      | 3.6 GB   | 36 GB     | 86.4 GB        |

  ---
### Combined Storage Scenarios (10h night, current - uncompressed FF)

  | Scenario                 | FF    | FT    | Frames | Video (4096k) | Total  |
  |--------------------------|-------|-------|--------|---------------|--------|
  | Standard night (minimal) | 12 GB | -     | -      | -             | ~12 GB |
  | + save_frame_times       | 12 GB | 10 MB | -      | -             | ~12 GB |
  | + save_frames            | 12 GB | 10 MB | 0.9 GB | -             | ~13 GB |
  | + raw_video_save         | 12 GB | 10 MB | 0.9 GB | 9 GB          | ~22 GB |

  ---
### Continuous Capture (24h) Storage (current - uncompressed FF)

  | Option            | Day (14h) | Night (10h) | 24h Total |
  |-------------------|-----------|-------------|-----------|
  | FF files          | ❌ None   | 12 GB       | 12 GB     |
  | FT files          | 14 MB     | 10 MB       | 24 MB     |
  | Saved frames      | 1 GB      | 0.9 GB      | 1.9 GB    |
  | Raw video @ 4096k | 12.6 GB   | 9 GB        | 21.6 GB   |

  ---
### Continuous Capture (24h) Storage (with planned day-only comp FF)

  *The actual compression ration of day FF is unknown at this time*

  | Option                             | Day (14h) | Night (10h) | 24h Total |
  |------------------------------------|-----------|-------------|-----------|
  | FF files (uncompressed)            | 18 GB     | 12 GB       | 30 GB     |
  | FF files (day compressed only)     | 8 GB      | 12 GB       | 20 GB     |
  | FF files (day and night comp)      | 8 GB      | 6 GB        | 14 GB     |
  | FT files (save_frame_times)        | 14 MB     | 10 MB       | 24 MB     |
  | Saved frames (save_frames)         | 1.3 GB    | 0.9 GB      | 2.2 GB    |
  | Raw video @ 4096k (raw_video_save) | 12.6 GB   | 9 GB        | 21.6 GB   |

  ---
### 10-Day Storage Accumulation

#### Default retention settings from .config:

  | Setting              | Default | Description                              |
  |----------------------|---------|------------------------------------------|
  | `arch_dirs_to_keep`  | 20      | ArchivedFiles folders (FF, FR, results)  |
  | `bz2_files_to_keep`  | 20      | Compressed (.bz2) archive files          |
  | `capt_dirs_to_keep`  | 8       | CapturedFiles folders (before archiving) |
  | `frame_days_to_keep` | 4       | Saved frames (JPG/PNG)                   |
  | `video_days_to_keep` | 2       | Raw video segments                       |
  | `times_days_to_keep` | 8       | FT timestamp files                       |
  | `logdays_to_keep`    | 30      | Log files                                |

#### Standard Night Capture (10h nights)

  | Data Product        | Day 1  | Day 2  | Day 4   | Day 8   | Day 10  | Max (at cap) |
  |---------------------|--------|--------|---------|---------|---------|--------------|
  | FF files.           | 12 GB  | 24 GB  | 48 GB   | 96 GB   | 120 GB  | 240 GB (20d) |
  | FT files            | 10 MB  | 20 MB  | 40 MB   | 80 MB   | 80 MB*  | 80 MB (8d)   |
  | Raw JPGs†           | 0.9 GB | 0.9 GB | 0.9 GB  | 0.9 GB  | 0.9 GB  | 0.9 GB       |
  | Timelapse (mp4+tar) | 80 MB  | 160 MB | 320 MB* | 320 MB* | 320 MB* | 320 MB (4d)  |
  | Raw video @ 4096k   | 9 GB   | 18 GB  | 18 GB*  | 18 GB*  | 18 GB*  | 18 GB (2d)   |
  | **Cumulative**      | 22 GB  | 43 GB  | 67 GB   | 115 GB  | 139 GB  | 259 GB       |

  †Raw JPGs: fixed overhead, deleted after timelapse creation  
  *Capped by retention policy

#### Continuous Capture with 24h FF (hypothetical, 10h night + 14h day, day FF compressed)

  | Data Product        | Day 1  | Day 2  | Day 4   | Day 8   | Day 10  | Max (at cap) |
  |---------------------|--------|--------|---------|---------|---------|--------------|
  | FF files (24h)‡     | 20 GB  | 40 GB  | 80 GB   | 160 GB  | 200 GB  | 400 GB (20d) |
  | FT files            | 24 MB  | 48 MB  | 96 MB   | 192 MB  | 192 MB* | 192 MB (8d)  |
  | Raw JPGs†           | 1.8 GB | 1.8 GB | 1.8 GB  | 1.8 GB  | 1.8 GB  | 1.8 GB       |
  | Timelapse (mp4+tar) | 160 MB | 320 MB | 640 MB* | 640 MB* | 640 MB* | 640 MB (4d)  |
  | Raw video @ 4096k   | 22 GB  | 43 GB  | 43 GB*  | 43 GB*  | 43 GB*  | 43 GB (2d)   |
  | **Cumulative**      | 44 GB  | 85 GB  | 126 GB  | 206 GB  | 246 GB  | 446 GB       |

  †Raw JPGs: fixed overhead, deleted after timelapse creation  
  ‡Day FF compressed (~8 GB/day) + night FF uncompressed (~12 GB/night) = ~20 GB/day  
  *Capped by retention policy

### Storage Growth Summary

  | Scenario                      | Fixed Overhead | Daily Growth | 10-Day Total | 20-Day Cap |
  |-------------------------------|----------------|--------------|--------------|------------|
  | Night only (minimal)          | 0 GB           | 12 GB        | 120 GB       | 240 GB     |
  | Night + frames                | 1 GB           | 12 GB        | 122 GB       | 242 GB     |
  | Night + frames + video        | 1 GB           | 21 GB        | 140 GB       | 260 GB     |
  | 24h + frames                  | 2 GB           | 12 GB        | 123 GB       | 243 GB     |
  | 24h + FF* + frames            | 2 GB           | 20 GB        | 203 GB       | 403 GB     |
  | 24h + FF* + frames + video    | 2 GB           | 42 GB        | 247 GB       | 447 GB     |

  *day-only compression (estimated)

  - **Fixed Overhead**: Raw JPGs only (~1 GB night / ~2 GB 24h), deleted after timelapse creation
  - **Daily Growth**: Data generation rate (FF + timelapse + video)
  - **10-Day / 20-Day**: Actual storage after retention caps apply (video 2d, timelapse 4d, FT 8d, FF 20d)

markmac99:
Thanks for the very comprehensive stats. 

One thing I'd note is that night time FF file capture in the southern UK around the solstice is 18GB (currently capturing 5217 files @ 3.5MB each = 18,259 MB) as capture is currently running from 16:15 to 07:10, approx 15 hours.  The cameras in Greenland are capturing for even longer, around 18 hours = 23GB FF files.  If we take that as worst-case i estimate 24-hr continuous capture, with compressed FF during the day, could need 27GB per day in northern latitudes, without raw video.

dvida:
Actually, the cameras in Greenland capture 23 hours, pause for processing and upload, and then continue. So the worst-case scenario is effectively 24/7 nighttime capture.

Utils/ConvertCompressedFits.py
Cybis320: fixed it. this.

Utils/ConvertCompressedFits.py
Cybis320: Good point. Made it bidirectional.

dario-zubovic:
I think this is really fruitful direction, especially as compute cost vs storage cost of compression with new RPis is changing!

@Cybis320 would you mind testing out on a different branch AstroPy's whole-file compression that we discussed before? IIRC all that's needed is to provide astropy function that we already use to write out FITS files `*.fits.zip` (or `.gz`, `.bz2`, `.xz`) as filename instead of `*.fits`.

HDU compressions are cool, but I doubt any of those can outperform good old DEFLATE. And `.fits.zip`/`.fits.gz` has advantages:
- compresses whole FITS file, not just payload
- doesn't break compatibility with other software, as any off-the-shelf zip tool can be used to decompress files

Other downstream software (CMN_binViewer etc) can likewise be modified to detect `*.fits.zip` files and handle them seamlessly.

Otherwise I can give it a try later in January (on holidays without laptop currently)

Cybis320:
Hi Dario,

Surprisingly, the current method provides the highest compression ratio in my tests.

  | Method                | Avg Size   | Ratio | Savings   |
  |-----------------------|------------|-------|-----------|
  | RICE_1 (current)      | 4,183,488  | 1.99x | 49.6%     |
  | raw gzip (whole file) | 4,332,588  | 1.92x | 47.9%     |
  | HCOMPRESS_1           | 4,503,600  | 1.84x | 45.8%     |
  | GZIP_2 (DEFLATE)      | 4,921,056  | 1.69x | 40.8%     |

Also, just compressing the HDUs means everything downstream (except PixInsight) just works - no extra steps needed. 

I'll investigate again, but I'm not seeing the advantage of doing whole fits compression - but maybe I'm missing something.

Also, I tested day fits compression, and the compression ratio is surprisingly not that different from night fits - ~2x all around.

Edit: number above are for 1080p - doesn't affect the ratio though.

PR #284, opened 2024-04-14
Create CamManager.py

satmonkey:
This tool allows to quickly discover all visible XMeye cameras, no matter which IP address is set. Subsequently, it is possible to change the IP address to the desired one. No additional module to the standard RMS installation is needed. In case of missing GUI, cmdline version is available by using '-n' switch (testers welcome).
Only the wired interface is used, the name should be recognized automatically.
In short, it utilizes the UDP multicast broadcast feature built-in the XM cameras, and widely used in their VMS (CMS) tool. It has been successfully tested on the directly attached cameras, the cameras attached via unmanaged switch and via the router with wifi/lan bridge. 
The tool was created by patching the code which originally comes from: https://github.com/OpenIPC/python-dvr/blob/master/DeviceManager.py

### DIFF
--- a/Utils/CamManager.py
+++ b/Utils/CamManager.py
@@ -1,0 +1,796 @@
+"""
+MIT License
+
+Copyright (c) 2017 Eliot Kent Woodrich
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


satmonkey:
image

PR #289, opened 2024-05-24
Call FRbinViewer from generate mp4s

timbot18:
Added code to call the FRbinViewer utility from the Generate MP4 utility so both can be run based off of the magnitude and shower filters from GenerateMP4s

### DIFF
--- a/Utils/GenerateMP4s.py
+++ b/Utils/GenerateMP4s.py
@@ -178,11 +178,7 @@ def generateMP4s(dir_path, ftpfile_name, shower_code=None, min_mag=None, config=
 
         #print(com)
         subprocess.call(com, shell=True, cwd=dir_path)
-        
-   
 
-        		
-        	
         #Delete temporary images directory and files inside
         if os.path.exists(dir_tmp_path):
             try:
--- a/Utils/GenerateMP4s.py
+++ b/Utils/GenerateMP4s.py
@@ -222,7 +222,7 @@ def generateMP4s(dir_path, ftpfile_name, shower_code=None, min_mag=None, config=
         		o = "-" + o
         		frbv_switches.append(o)
         	options = ' '.join(frbv_switches)
-        	command = 'python -m Utils.FRbinViewer ' + options + " " + "-c " + frbvconfig + " " + dir_frbv_path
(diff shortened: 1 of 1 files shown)


markmac99:
@dvida  I'm keen to see this working but as written it will fail on most systems as it contains a hardcoded path to the FRbinViewer.py which is only correct on a Bookworm build Pi4/5.  Additionally frbinviewer is invoked using a system call to launch a child shell, which could consume additional memory and might cause issues on low-memory systems. 

Another issue - not due to this PR - is that FRbinViewer calls exit() if there are no FR files. This is a problem because GenerateMP4s is designed to be called from other Python code which might then be aborted if FRbinviewer exit()ed . The UKMON toolset uses GenerateMP4s from python and so i am concerned that it could cause our tools to fail.  

I suggest that I take a look at FRbinViewer to make it safe to call, and restructure it so it can be easily called by GenerateMP4s. This would allow @timbot18  to import it rather than using a system() call. Then this PR can be updated. 

I will take a look at FRbinViewer today and report back asap

markmac99:
taking another look, i think you might just be able to call the view() function in FRbinViewer.  
FRbinViewer itself is designed to process multiple FR files in a folder, however we don't need that here so we can just call view() directly, passing the foldername, ff name and fr name. 
You would need to put in place logic to parse frbv_switches and set True/False values for avg_background, timestamp, framenumber, append_ff, add_shower, and you'd need to import and call loadShowerAssociations() from FRbinviewer to add shower details if add_shower_name were true. The other optional params extract, extractformat, hide and split can be hardcoded to False, 'mp4', True and False respectively.

markmac99 (changes requested):
Line 225 will fail on Pi3 and older Pi4 systems, and on linux systems as it assumes a hardcoded path to the python codebase. However, i think a better approach would be to import the view() function from FRbinViewer and call it directly, passing the necessary arguments. See my comments on the PR. This will also avoid a bug in FRbinViewer which can cause it to exit improperly.

g7gpr:
https://github.com/CroatianMeteorNetwork/RMS/blob/10b703399a854d8a4d7b7d9f28a70c0ead18f44e/RMS/EventMonitor.py#L1751

The code linked above may be helpful as an ex ample.

PR #294, opened 2024-06-05
Fix GStreamer Recording Hang

Cybis320:
This update addresses the issue where GStreamer stops recording and hangs until the night session is terminated. The code now verifies that the pipeline has successfully reached the PLAYING state before proceeding further. It allows the camera 10 seconds after a successful ping to initiate an RTSP handshake, then makes up to five retries to start the GStreamer pipeline. If these attempts fail, it falls back to cv2.


g7gpr:
I'll be able to test this over the next 48 hours on Pi 4 32 bit, Pi 4 64 bit and Debian 12

markmac99:
finally had a chance to deploy this to one of my stations. Apologies, i've been very tied up with a website migration and family business. Will run it for a few days and try to simulate some failure conditions

dvida:
Hi all, any results from the testing? I presume there were no issues?

markmac99:
apologies, meant to confirm earlier. No issues here. Looks good.

PR #303, opened 2024-06-26
Utils.SkyFit2: add reference lenses option when building new platepar

adalava:
This tries to make much easier to build a new platepar when station uses the most common lenses: 4mm and 6mm M16 mount. The 720p reference data was taken from BR000F and BR0001, and the 1080p one was provided by Jocimar Justino.

This seems to work to me. Please check if this change is useful and if I'm missing something.

More tests are welcome. :)

Screenshot from 2024-07-24 20-55-35


dvida:
This is great - could you also add an explicit note that this is for the 720p resolution, and check that when the platepar is applied?

adalava:
> This is great - could you also add an explicit note that this is for the 720p resolution, and check that when the platepar is applied?

Since at this point the resolution is know (from the config), do you prefer to omit the incompatible references or keep listing them (with a note) and warn that  user that's incompatible?
For instance, show de dialog+warning in loop until it selects a compatible option or select "None"

dvida:
Yes, that would be great. Perhaps it would be best to list them but gray out if they are incompatible due to resolution differences.

adalava:
@dvida, the code and the screenshot was updated. I also changed the text displayed. Please check if it's ok. Thanks!

dvida:
Great work! I thought about this some more - would it be possible to remove the hardcoded values in the code and just replace them with actual platepar files (which can be stripped of the ID and location)? We can put template platepars in share/platepar_templates and name them e.g. template_720p_4mm.cal, with the added metadata in the platepar file itself (e.g. template_metadata: {...} ).

That way we can easily change the parameters, add new platepars, and we can even let users put a template of their choice in the data directory (detect all "template*.cal files) and have it automatically loaded.

Thanks!

adalava:
> Great work! I thought about this some more - would it be possible to remove the hardcoded values in the code and just replace them with actual platepar files (which can be stripped of the ID and location)? We can put template platepars in share/platepar_templates and name them e.g. template_720p_4mm.cal, with the added metadata in the platepar file itself (e.g. template_metadata: {...} ).

Great idea! I changed the PR code to read them from the disk. The template file needs a key/dict like the one bellow, or it will be ignored.
```json
    "template_metadata": 
    {
      "description": "Generic 4mm (720p)"
    },
```

Issue #304, opened 2024-07-06
EventMonitor crashes when files are in CapturedFiles


g7gpr:
```
Process EventMonitor-2:
Traceback (most recent call last):
  File "/usr/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/home/pi/source/RMS/RMS/EventMonitor.py", line 2129, in run
    self.getEventsAndCheck(last_check_start_time,next_check_start_time)
  File "/home/pi/source/RMS/RMS/EventMonitor.py", line 2106, in getEventsAndCheck
    self.checkEvents(self.config, test_mode=testmode)
  File "/home/pi/source/RMS/RMS/EventMonitor.py", line 1883, in checkEvents
    file_list = self.getFileList(observed_event)
  File "/home/pi/source/RMS/RMS/EventMonitor.py", line 1569, in getFileList
    file_list += self.findEventFiles(event, self.getDirectoryList(event), [".fits", ".bin"])
  File "/home/pi/source/RMS/RMS/EventMonitor.py", line 1524, in findEventFiles
    dirlist = os.listdir(directory)
NotADirectoryError: [Errno 20] Not a directory: '/home/pi/RMS_data/CapturedFiles/AU00XX_20240703_074311_092275FF_AU00XX_20240703_110339_991_0300032_frame_0252_Even.bmp'
```

g7gpr:
On one station the CapturedFiles directory contained not only folders, but also files which had been copied there by mistake.

g7gpr:
pr #305 addresses this issue, which is caused by files, named in a similar way to an observation directory being handled as a direcotry.

PR #305, opened 2024-07-07
Prevent getDirectoryList from returning files

g7gpr:
When CapturedFiles directory contains files, which are named in a similar way to an observation directory, and an event is created which references a time which matches the naming of the files, files get returned from getDirectoryList, rather than just directories, as described in issue #304  

This pull request removes any files, so that only directories are returned.


g7gpr:
Tested successfully on station which observed this problem.

PR #307, opened 2024-07-12
Change GStreamer Pipeline to UDP Protocol

Cybis320:
Recently, a station was having an issue whereby the RTSP server (the IP camera) was terminating the RTSP stream at regular intervals. It was solved by switching protocol from TCP to UDP.

TCP is connection-oriented, maintaining a continuous connection that requires acknowledgment for data transmission. UDP is connectionless, simply sending packets without establishing or maintaining a connection. If there are intermittent network issues or brief interruptions, TCP might interpret these as a broken connection and terminate the stream. UDP, being connectionless, would simply continue sending packets without caring about the connection state.

If network congestion occurs, TCP might reduce its transmission rate significantly or pause transmission, potentially leading to timeouts and stream termination. UDP would continue to send at a constant rate, which might result in some packet loss but would maintain the stream.

TCP retransmits lost packets to ensure reliable delivery. UDP does not retransmit lost packets. In poor network conditions, TCP might spend time retransmitting lost packets, potentially leading to buffer overflows or timeouts on the server side, causing it to terminate the stream. UDP would simply continue sending new packets, maintaining the flow of the stream even if some packets are lost.

In TCP, if a packet is lost, subsequent packets must wait for the lost packet to be retransmitted before they can be delivered to the application. UDP does not have this issue. This could cause significant delays in TCP transmission during packet loss, potentially leading to timeouts or the server interpreting the delay as a dead connection.

So UDP might lead to more reliable stream and more consistent timestamping at the expense of getting corrupted frames.

It would be nice if this PR could be tested on the most troublesome stations to see if it helps. 

My understanding is that cv2 defaults to TCP... so it might even improve connection reliability of gst over cv2.

Also, UDP would be a must for stations running multiple clients on one server (multiple machines streaming from one camera)

One minor issue when running UDP is that GStreamer might issue a warning message about memory size:
```
warning: Could not create a buffer of requested 524288 bytes (Operation not permitted). Need net.admin privilege?
0:00:00.107728875 32187     0x24a0d6a0 WARN                  udpsrc gstudpsrc.c:1647:gst_udpsrc_open:<udpsrc0> have udp buffer of 212992 bytes while 524288 were requested
```
The stream will proceed fine despite the warning, but increasing the system’s maximum UDP buffer size by modifying the /etc/sysctl.conf file solves the issue. Adding the following lines to /etc/sysctl.conf :
```
net.core.rmem_max=524288
net.core.wmem_max=524288
```

After making these changes, apply them by running:
```
sudo sysctl -p
```

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -541,7 +541,7 @@ class BufferedCapture(Process):
 
         # Define the source up to the point where we want to branch off
         source_to_tee = (
-            "rtspsrc buffer-mode=1 protocols=tcp tcp-timeout=5000000 retry=5 "
+            "rtspsrc buffer-mode=1 protocols=udp retry=5 "
             "location=\"{}\" ! "
             "rtph264depay ! tee name=t"
             ).format(device_url)


markmac99:
I'll try it out on UK0006, my oldest and least reliable camera. 

However, we probably need to make this optional, with a default of TCP and with a setting in .config. Many firewalls and routers block UDP by default (other than dns and ntp), and many network admins disallow UDP internal traffic. Any machine can broadcast UDP packets and because there's no ACK and no destination is required, a malicious UDP client can easily flood a network or overwhelm poorly configured clients, leading to outages and compromised security. UKMON have actually encountered this situation in some schools. If the camera is connected via a switch or router, then there might be a problem with comms.

Cybis320:
Good Points. I agree it’d be best to make it optional if we decide to incorporate it at all. 

I’m curious if UDP will solve any issues out there as it did for this one machine and if it would make things worse for anyone. With that information, we could decide which of UDP or TCP should be the default. Ideally we should test it on a plain vanilla RPi currently running gst cleanly to make sure it doesn’t make things worse. Mostly, it should be tested on machine having disconnect issues. Machines dropping frames because they can’t keep up with the frame rate will probably not see any improvements - I can even image scenarios where it’d make things worse on such machines if the buffers are somehow smaller - so that’d be something to keep an eye on.

- Luc

g7gpr:
I can deploy on a vanilla RPI4, but we need to make this configurable, possibly with automated fall-back to TCP.

g7gpr:
Running on AU0004,6, 28-2F

satmonkey:
Running on CZ0001 now

satmonkey:
no problem seen during last night capture. FF intervals report as expected.

g7gpr:
Also no problems seen. AU0028-2F, single machine running 8 cameras might have had marginally better performance. Unfortunately not able to test on Pi4 yet.

markmac99:
Am i ok just to make the one change to device_str in BufferedCapture? I noticed that there are a huge number of other changes in that file in the dev branch, and i don't want to be testing all those at the same time as it might make it difficult to be sure of the cause, if there are any issues.

Cybis320:
Argh. I apologize for that - I didn't realize dev had so much work done on it. I should have based it on master and flagged it as draft. Should I close this PR and resubmit based on Master? And, yes, the only change that this was meant to test is the pipeline str in BufferedCapture.

> Am i ok just to make the one change to device_str in BufferedCapture? I noticed that there are a huge number of other changes in that file in the dev branch, and i don't want to be testing all those at the same time as it might make it difficult to be sure of the cause, if there are any issues.

dvida:
Hi Luc,
Yes please, and make it an option in the config file. I suggest that this is only used on systems that need it.

Cybis320:
Closing this PR. Opened PR #307 based on Master.

PR #309, opened 2024-07-19
Fix Rolling Shutter Correction sometimes using image width instead of…

Cybis320:
Thank you Jamie Shepherd for noticing this bug. Jamie noticed that if RMS is in auto-detect mode, then the correct divisor of 720 is applied, as it is fetched from the .config file as self.config.height. In manual use of SkyFit2, it fetches up the width of  1280 instead the height of 720.

There might be a similar inversion in FFTalign line 45:
```
def addPoint(img, xc, yc, radius):
    """ Add a point to the image. """

    img_w = img.shape[1]
    img_h = img.shape[0]
```
Shouldn't the 1 and 0 be swapped?

### DIFF
--- a/Utils/SkyFit2.py
+++ b/Utils/SkyFit2.py
@@ -5575,7 +5575,7 @@ class PlateTool(QtWidgets.QMainWindow):
                 img_h = self.config.height
 
             else:
-                img_h = self.img.data.shape[0]
+                img_h = self.img.data.shape[1]
 
             # Compute the corrected frame time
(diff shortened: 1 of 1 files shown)


dvida:
Did you confirm this by printing the values? shape[0] is usually the height

Cybis320:
Hmm, you are right. Something else is going on. I'll dig further into it.

Cybis320:
Ok, reopening. I printed self.img.data.shape[0] in skyfit and it does give me 1280. I think img.shape[0] in FFTalign is probably correct (720).

PR #310, opened 2024-07-19
Add GStreamer Pipeline TCP or UDP Protocol Config Switch

Cybis320:
I made the PR relative to Master and added a new config parameter `protocol`. The default is tcp. Setting `protocol: udp` in `.config` switches the GStreamer pipeline to UDP.

UDP should provide more reliable connections, especially on Multicam Linux machines.

One minor issue when running UDP is that GStreamer might issue a warning message about memory size:
```
warning: Could not create a buffer of requested 524288 bytes (Operation not permitted). Need net.admin privilege?
0:00:00.107728875 32187     0x24a0d6a0 WARN                  udpsrc gstudpsrc.c:1647:gst_udpsrc_open:<udpsrc0> have udp buffer of 212992 bytes while 524288 were requested
```

The stream will proceed fine despite the warning, but increasing the system’s maximum UDP buffer size by modifying the `/etc/sysctl.conf` file solves the issue. Adding the following lines to `/etc/sysctl.conf `:
```
net.core.rmem_max=524288
net.core.wmem_max=524288
```
After making these changes, apply them by running:
```
sudo sysctl -p
```

### DIFF
--- a/.config
+++ b/.config
@@ -52,6 +52,10 @@ reboot_lock_file: .reboot_lock
 
 device: rtsp://192.168.42.10:554/user=admin&password=&channel=1&stream=0.sdp ; device id
 
+; Transport Layer Protocol: 'tcp' or 'udp'. Defaults to tcp. UDP typically
+; provides more stable connectivity.
+protocol: tcp
+
 ; Media Backend: options are gst, cv2, or v4l2.
 ;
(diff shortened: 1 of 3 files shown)


Cybis320:
Sorry. I'm closing this and starting over. I LOVE git... grrr

PR #312, opened 2024-07-23
Fix GStreamer hanging after multiple attempts

Cybis320:
Occasionally, GStreamer would hang after 5 attempts during a night run. The retry loop doesn't reset/initialize the attempt counter and would hang on subsequent reconnect attempt after a disconnect. This PR initialize the attempt counter.


Cybis320:
My fix made no sense... sorry.

PR #313, opened 2024-07-25
Replace datetime.datetime.utcnow() by datetime.datetime.now(datetime UTC)

adalava:
.utcnow() is being deprecated:
   "DeprecationWarning: datetime.datetime.utcnow() is deprecated and
   scheduled for removal in a future version. Use timezone-aware
   objects to represent datetimes in UTC"

Output differences:
```python
>>> str(datetime.datetime.utcnow()) + ' UTC'
'2024-07-25 13:12:15.773880 UTC'

>>> str(datetime.datetime.now(datetime.UTC)) + ' UTC'
'2024-07-25 13:12:23.862181+00:00 UTC'
```

### DIFF
--- a/RMS/Astrometry/ApplyAstrometry.py
+++ b/RMS/Astrometry/ApplyAstrometry.py
@@ -45,6 +45,7 @@ from RMS.Formats.FFfile import filenameToDatetime
 from RMS.Formats.FTPdetectinfo import (findFTPdetectinfoFile,
                                        readFTPdetectinfo, writeFTPdetectinfo)
 from RMS.Math import angularSeparation, cartesianToPolar, polarToCartesian
+from RMS.Misc import rms_datetime
 
 pyximport.install(setup_args={'include_dirs':[np.get_include()]})
 from RMS.Astrometry.CyFunctions import (cyraDecToXY, cyTrueRaDec2ApparentAltAz,
@@ -833,7 +834,7 @@ def applyAstrometryFTPdetectinfo(dir_path, ftp_detectinfo_file, platepar_file, U
 
 
     # Calibration string to be written to the FTPdetectinfo file
-    calib_str = 'Calibrated with RMS on: ' + str(datetime.datetime.utcnow()) + ' UTC'
+    calib_str = 'Calibrated with RMS on: ' + str(rms_datetime.utcnow()) + ' UTC'
 
     # If no meteors were detected, set dummpy parameters
     if len(meteor_list) == 0:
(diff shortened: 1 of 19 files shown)


adalava:
I'll change this PR to "draft" since this doesn't work on python 2.x  and the chance isn't urgent. Apparently where's no solution that works for both Python 2 and 3.

Results for python2:

```bash
$ python2
Python 2.7.18 (default, Jan 26 2024, 00:00:00) 
[GCC 14.0.1 20240125 (Red Hat 14.0.1-0)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
```
```python
>>> import datetime
>>> datetime.datetime.now(datetime.UTC)
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'UTC'
>>> datetime.UTC
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'UTC'
>>>
```

adalava:
What about creating a wrapper function to replace those calls, so it can choose the correct approach for each environment?

Something like:
```python
from RMS.Misc import rms_datetime

utctime = rms_datetime.utcnow()
```

dvida:
Agreed, can you add a check on the python version and then have it use the appropriate function?

satmonkey:
Currently, I register two ways to handle this:

- handle the Python 2/3 compability, e.g.:
if sys.version_info.major < 3:

- by the exception thrown when importing:
except ImportError:

I suggest to use one of these.

M.

adalava:
I've made the changes to create the wrapper functions. When reviewing, please double check that "from RMS.Misc import rms_datetime" or similar exists on all files changed.

I'll keep the PR as "draft" since I won't be able to test it properly until later next week. Also, I think this should be tested on "dev" branch for a longer period. It appears inoffensive but  changes many files that seems to be rarely touched (or deprecated like istream) and I'm not sure that I'll able to test every piece of code... :/

Cybis320:
This is great, thank you!
I think the DeprecationWarning message suggestion doesn't work. May I suggest to change the wrapper function to:
```
class RmsDateTime:
    """ Class to hold utcnow() wrapper function definition.
        Select the best approach to retrieve current UTC time according to Python version.
    """
    if sys.version_info[0] < 3:
        @staticmethod
        def utcnow():
            # Python 2: Use the existing utcnow, which is not timezone-aware.
            return datetime.datetime.utcnow()
    else:
        @staticmethod
        def utcnow():
            # Python 3: Get timezone-aware UTC time and then make it naive.
            return datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```

To use in the code:
```
from RMS.Misc import RmsDateTime

RmsDateTime.utcnow()
```

PR #314, opened 2024-07-27
Fix microsecond error

Cybis320:
Adds missing microsecond=True argument.

### DIFF
--- a/Utils/PlotFieldsums.py
+++ b/Utils/PlotFieldsums.py
@@ -64,7 +64,7 @@ def plotFieldsums(dir_path, config):
                 print('File {:s} is corrupted!'.format(file_name))
 
             # Extract the date and time from the FF file
-            dt = filenameToDatetime(file_name)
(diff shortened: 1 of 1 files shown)


g7gpr:
Reprocessed same archive on AU0028, and no logging on microseconds. However, just forcing microsecond to be true might be missing something.

g7gpr:
On reflection, if there is a reason why we are not automatically parsing to see the length to indicate milliseconds or microseconds, forcing microseconds at this point in the code is OK.

dvida:
@g7gpr agreed - @Cybis320 how about this, can you confirm that it works?

g7gpr:
Keen to see this progress - what is the status?

Cybis320:
Yes, it worked for me:
FS_US9995_20240730_121615_726865_0832768_fieldsum.bin, 2024-07-30 12:16:15.726865
FS_HR002M_20240306_045241_556_1044224_fieldsum.bin, 2024-03-06 04:52:41.556000

PR #318, opened 2024-07-31
Config File Enhancements - Restyling, Template, and Audit Tool

Cybis320:
I'm proposing the following changes to `.config`:

Restyling:
- Move inline comments to a separate line (per Denis request).
- 80-character line limit (per Denis request).
- Various cleanups

`.config` is the only RMS source file that does not automatically update on production machines. Individual operators are responsible for any change to their `.config` files. However, operators may not be aware that new options have been introduced, or old options removed. Therefore, the '.config' file tends to be out-of-date with the rest of the code base on many installations.

I'm proposing the following to try to make things easier for operators:

- Add a template config file, `.configTemplate`,  in the RMS root directory. Yes, operators could just look at the GitHub version, but I think having it in the RMS root folder would be more convenient and more likely to be noticed.
- Add an explanatory note at the beginning of `.config`
- Add an audit tool to quickly highlight the difference between a `.config` file and the template
- Generate an automatic report at the start of capture where it is more likely to be noticed
- Generate a text file in archive where it can be referenced.

To make things easier for developers, the tool can also compare the template config to options supported  in `ConfigReader.py`. The tool highlights obsolete options and options not present in the template.

I've also removed duplicate entries in ConfigReader.

Here's an example of the report generated to the user. There might be reasons not to include certain options in the report - this can be done with an omit list.

```
================================================================================
                           CONFIG COMPARISON RESULTS                            
================================================================================

             OPTIONS MISSING IN .CONFIG FILE (PRESENT IN TEMPLATE)              
             ****** CONSIDER MANUALLY UPDATING .CONFIG FILE ******              
                          Default values will be used:                          
--------------------------------------------------------------------------------
 • camera_buffer
 • camera_latency

                     OPTIONS COMMENTED OUT IN .CONFIG FILE                      
                          Default values will be used:                          
--------------------------------------------------------------------------------
 • public_elevation
 • public_latitude
 • public_longitude
 • star_catalog_band_ratios

================================================================================
                         Total options in template: 157                         
                       Total options in .config file: 155                       
                              Common options: 155                               
================================================================================

```

Here's an example with `--dev` flag:
```

================================================================================
                           CONFIG COMPARISON RESULTS                            
================================================================================

             OPTIONS MISSING IN .CONFIG FILE (PRESENT IN TEMPLATE)              
             ****** CONSIDER MANUALLY UPDATING .CONFIG FILE ******              
                          Default values will be used:                          
--------------------------------------------------------------------------------
 • camera_buffer
 • camera_latency

            OPTIONS MISSING IN TEMPLATE FILE BUT IMPLEMENTED IN RMS:            
--------------------------------------------------------------------------------
 • brightness
 • contrast
 • dark_file
 • event_monitor_db_name
 • force_v4l2
 • mask_download_permissive
 • mask_remote_name
 • min_lines
 • platepar_template_dir
 • remote_mask_dir
 • use_dark

                     OPTIONS COMMENTED OUT IN .CONFIG FILE                      
                          Default values will be used:                          
--------------------------------------------------------------------------------
 • public_elevation
 • public_latitude
 • public_longitude
 • star_catalog_band_ratios

                    OPTIONS COMMENTED OUT IN TEMPLATE FILE:                     
--------------------------------------------------------------------------------
 • public_elevation
 • public_latitude
 • public_longitude
 • star_catalog_band_ratios

================================================================================
                         Total options in template: 157                         
                       Total options in .config file: 155                       
                              Common options: 155                               
================================================================================

```

Usage example:
`python -m Utils.AuditConfig`
`python -m Utils.AuditConfig --dev`

### DIFF
--- a/.config
+++ b/.config
@@ -197,6 +197,9 @@ slideshow_enable: false
 ; Automatically reprocess broken capture directories (due to e.g. power cut or system crash)
 auto_reprocess: true
 
+; Prioritize capture over reprocessing - do not start reprocessing a new directory if should be capturing
+prioritize_capture_over_reprocess: false
+
 ; Flag file which indicates that the previously processed files are loaded during capture resume
 capture_resume_flag_file: .capture_resuming
 
--- a/.config
+++ b/.config
@@ -1,83 +1,110 @@
-; IMPORTANT: There always must be at least one space after the argument value and before the semicolon in front of the comment.
-; For example, "parameter: value ; comment" is correct, while "parameter: value; comment" is incorrect.
+; IMPORTANT: Unlike other RMS files, the '.config' file is never automatically
+; updated. When changes are introduced to the config file in RMS, these changes
+; will NOT be automatically reflected in the active '.config' file on 
+; individual machines. Users should manually update their '.config' file if 
+; changing any option from its default value is desired, or to simply stay 
+; up-to-date. If an option is missing from the config file, a default value 
+; will be automatically applied.
+; An up-to-date template config '.configTemplate' file, containing the current
+; options, is located in the RMS root directory.
+; Run python -m Utils.AuditConfig to compare the .config file to the template. 
 
 [System]
 
-stationID: XX0001
-latitude: 43.19301 ; WGS84 +N  (degrees )
-longitude: -81.315555 ; WGS84 +E (degrees)
-elevation: 327 ; mean sea level EGM96 geoidal datum, not WGS84 ellipsoidal (meters)
+; The assigned station ID
+stationID: US9999
+
+; WGS84 +N (degrees)
+latitude: 43.19301
+
+; WGS84 +E (degrees)
+longitude: -81.315555
+
+; Mean sea level EGM96 geoidal datum, not WGS84 ellipsoidal (meters)
+; Can be obtained from Google Earth or other apps. Raw GPS altitude should not
+; be used as the software converts from EGM96 to WGS84 internally.
+elevation: 327
 
 ; Should be set only if full CAMS compatibility is desired
-cams_code: 0 
+cams_code: 0
+
+; Weblog and PerfMonitor
+; ----------------------
 
 ; Show this camera on the GMN weblog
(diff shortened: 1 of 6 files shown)


Cybis320:
Additionally. I don't think these two options are used anymore, right?:
`event_monitor_db_name`
`force_v4l2`

dvida:
Great job! I'd be happy to merge it, however some values have to be reset to the previous values. E.g. the station code has to be XX0001 and the weblog should be enabled.

 event_monitor_db_name is used in the code but for safety (people shouldn't mess with the name) is not given in the config file. We should also remove the name of the platepar and the mask file, as those names are hardcoded upstream in some scripts.

 force_v4l2 has been removed, but some old config files still support it, so we need to read it and set the new media_backend options appropriately, so we still need to support that option.

Another point - if a whole block is commented out and there are breaks in the block, the line break should also start with ;

E.g.
; Server will generate obfuscated coordinates for public datasets.
; Optionally, uncomment and populate the three parameters below to use a
; specific public location if desired. The recommended precision is two decimal
; places for latitude and longitude, and ~10 m for elevation.
;
; WGS84 +N (degrees)
; public_latitude: 0.0
;
; WGS84 +E (degrees)
; public_longitude: 0.0
;
; mean sea level EGM96 (meters)
; public_elevation: 0.0

This gives better readability.


There's also a matter of having to keep both .config and the template version updated and identical at all times, doubling the work and making it prone to errors. Is there perhaps another mechanism we might employ?

Cybis320:
Sorry about XX0001. I'll remove `event_monitor_db_name`, `platepar_name`, and `mask` from the `.config` and the reporting (but leave them in configreader).

I hear you about the pain of keeping two versions in sync. This can probably be done automatically in RMS_backup by:

1. only having one `.config` in the source code
2. in `RMS_update`, rename the `.config` to `.configTemplate` before restoring the backed up user `.config`
3. The end result is the user `.config` and the renamed source file are both there

Issue #319, opened 2024-07-31
kht module detection is failing when python minor version > 9


g7gpr:
In ConfigReader.py find BinaryPath generates a python version such as '3.11' using the code at 

https://github.com/CroatianMeteorNetwork/RMS/blob/3a20e04d8fe8104b02af82b3fc34d3e00ace2bd4/RMS/ConfigReader.py#L117

however the kht module is found in a directory named

```
lib.linux-x86_64-cpython-311
```

which does not include a dot. This causes the kht module not to be found, leading to the kht_wrapper bug. 

Earlier version of python/cython placed the binary files in a directory with a dot separator

```
lib.linux-x86_64-3.7
```

I suspect that this could be quickly fixed by giving an explicit path in the .config file, working on a patch on 

https://github.com/CroatianMeteorNetwork/RMS/blob/3a20e04d8fe8104b02af82b3fc34d3e00ace2bd4/RMS/ConfigReader.py#L117

dvida:
I believe I encountered the same issue and the solution was to purge all previous builds, but I might be wrong.

dvida:
These are the names on build dirs on a Pi3 running Python 2:
![image](https://github.com/user-attachments/assets/79d4adac-06f4-43c3-850f-495eac0e89ae)

I belive the build script already tries to delete a build directory, but that may fail if e.g. the previous build was done with sudo. But Dave's approach is the true way forward, one a check for Py2 is added.

g7gpr:
Merged into prerelease, so closing now.

PR #320, opened 2024-07-31
Fix for kht_binary_search

g7gpr:
Issue #319 notes that detector can fail with the following error

```
2024/07/31 18:45:27-INFO-QueuedPool-line:204 - Traceback (most recent call last):
  File "/home/au0004/source/RMS/RMS/QueuedPool.py", line 356, in _workerFunc
    result = func(*all_args, **self.func_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/au0004/source/RMS/RMS/DetectStarsAndMeteors.py", line 98, in detectStarsAndMeteors
    meteor_list = detectMeteors(img_handle, config, flat_struct=flat_struct, dark=dark, mask=mask)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/au0004/source/RMS/RMS/Detection.py", line 1107, in detectMeteors
    line_list = getLines(img_handle, config.k1_det, config.j1_det, config.time_slide, config.time_window_size,
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/au0004/source/RMS/RMS/Detection.py", line 413, in getLines
    kht.kht_wrapper.argtypes = [npct.ndpointer(dtype=np.double, ndim=2),
    ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ctypes/__init__.py", line 389, in __getattr__
    func = self.__getitem__(name)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ctypes/__init__.py", line 394, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: python: undefined symbol: kht_wrapper
```

This is because the binary is not correctly detected when python minor version is greater than 9. This PR addresses this issue. 

Adding the full binary name in .config is a good workaround.

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -125,7 +125,7 @@ def findBinaryPath(config, dir_path, binary_name, binary_extension):
             binary_dir_version = binary_dir.split('-')[-1]
 
 
-            # the binary directory may or may not contain a dot in the directory name
+            # the binary directory may or may not contain a dot in the version
             # e.g lib.linux-x86_64-3.7 vs lib.linux-x86_64-cpython-311
             if '.' in binary_dir_version:
                 py_version = "{:d}.{:d}".format(sys.version_info.major, sys.version_info.minor)
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -114,7 +114,10 @@ def findBinaryPath(config, dir_path, binary_name, binary_extension):
     else:
         # If there are more candidates, find the right one for the running version of python, platform, and
         #   bits
-        py_version = "{:d}.{:d}".format(sys.version_info.major, sys.version_info.minor)
+        if sys.version_info.minor <= 9:
+            py_version = "{:d}.{:d}".format(sys.version_info.major, sys.version_info.minor)
+        else:
+            py_version = "{:d}{:d}".format(sys.version_info.major, sys.version_info.minor)
 
         # Find the compiled module for the correct python version
         for file_path in file_candidates:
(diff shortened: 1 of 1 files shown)


dvida:
A potential edge case: what if it's running on Python 2.7, like on some RPi3s?

g7gpr:
Python 2.7 has a minor version number <= 9, so will search for major.minor, which works, I think? 

Did I miss something here?

Purging would solve the problem, as the code would pick the only build available.

dvida:
To play Devil's advocate, how about the final version of Python 2, which is 2.12?

g7gpr:
Will check.

g7gpr:
https://mail.python.org/archives/list/python-announce-list@python.org/thread/OFCIETIXLX34X7FVK5B5WPZH22HXV342/

Afaik 2.7.18 was the last python release. Was 2.12 a special Pi version?

I think my strategy is flawed. A better approach would be split the directory name by '-', take the last part, and see if it contains a dot.

dvida:
Whoops, by bad. Then the current mechanism is fine, let's not complicate things.

I'll wait for one more person to test and then I'll merge it.

g7gpr:
I reworked it; I think your point is good, and the revised code is more robust.

Cybis320:
I tested the PR on python3.7 and python 3.11 and it works.

PR #330, opened 2024-08-05
Automatic download and display of a video montage

g7gpr:
This utility manages downloading of files into a temporary directory, or a directory specified by a user, and generating an ffmpeg command string to combine these files into a video mosaic. 

Optionally, it can operate in a kiosk mode, command switch -a, where it plays the videos continually, and updates with an interval set by the command line.

Shape of the video mosaic can be selected, at present it must be a rectangular number, and each tile must be populated. 
 
The simplest command 

```
python -m Utils.VideoMosaic
```
Downloads the videos from 4 cameras, and produces a mosaic. 

https://1drv.ms/v/c/6165251cc10ab7b0/EQ0q2RibEA5Asue1u3CsGp4BdNW_HueyL7JJ-R97EEDFJA?e=ueuoNB

A more sophisticated command 

```
python -m Utils.VideoMosaic -c AU001U,AU001V,AU001W,AU001X,AU001Y,AU001Z -r 3840 1440 -s 3 2 -o ~/station_video.mpg -w ~/videos/pioneer/ -k
```
https://1drv.ms/v/c/6165251cc10ab7b0/EegFzd1vI2tHrR-9MaXthd0Bet8s011jXuDiTmTyHpgQqA?e=M8WhD2

downloads the videos of  6 stations, renders them at 3840 by 1440, 3 across by 2 down, and saves to station_video in the user's root directory. The files were downloaded into ~/videos/pioneer, and were kept in that location. The working directory does not need to be emptied before subsequent runs. Where videos are of different lengths, the shorter videos are compensated by reducing the frame rate.

Station IDs can be passed as upper or lower case letters.

### DIFF
--- a/Utils/VideoMosaic.py
+++ b/Utils/VideoMosaic.py
@@ -459,7 +459,7 @@ def videoMosaic(station_ids, x_shape=2, y_shape=2, x_res=1280, y_res=720, equali
                     generation_duration),verbosity=2)
 
         if keep_files:
-        printv("Downloaded files in {:s}".format(working_directory),verbosity=2)
+            printv("Downloaded files in {:s}".format(working_directory),verbosity=2)
     else:
         for input_video in input_video_paths:
             os.unlink(input_video)
--- a/Utils/VideoMosaic.py
+++ b/Utils/VideoMosaic.py
@@ -48,6 +48,9 @@ import subprocess
 import cv2
 from RMS.Misc import mkdirP
 import time
+import json
+from random import shuffle
+from RMS.Misc import sanitise
 
 if sys.version_info[0] < 3:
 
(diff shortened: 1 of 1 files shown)


Cybis320:
Looks fantastic. I'll test - just need to find time.

g7gpr:
The all-sky video timelapse is interesting. I've had it working once. However it needs to run in the opposite direction. Instead of mapping observed pixels to a projection, I think it would work better if we started from the projection pixels and then found the sensor pixels which matched them. I think this would fix the aliasing problems. Then for extra work, fill in the gaps by using images from a different time.

g7gpr:
image
This works by pre-computing transforms from the destination image pixels back to the different camera pixels, which avoids aliasing problems. It almost runs fast enough for real time. I've experimented with filling in the gaps using images from different times, but it is not satisfactory. A view of the sky with more than one moon does not seem authentic.

PR #333, opened 2024-08-05
Update requirements.txt as tflite-runtime 2.14.0 now support python 3.11

Cybis320:
https://pypi.org/project/tflite-runtime/


dvida:
Thanks - it still doesn't support python 3.12, can you limit to to <=3.11?

Cybis320:
We could, but do we check every package? The previous `<'3.10'` made sense because we had a separate solution for 3.10. I haven't check if there is an adhoc solution for 3.12. If there isn't, then it's easier to maintain if we leave it fail until a version of tflite-runtime becomes compatible with 3.12, no?
Is this package the current bottleneck to 3.12?

Issue #316, opened 2024-07-29
Add text status file to archive


g7gpr:
Create a text file, in an extensible format, containing salient data of previous observation containing information such as 

```
#RMS_update date and the commit hash from the last RMS_Update on the following lines:
Date:   Mon Jul 22 10:47:01 2024 -0400   
commit: 3a20e04d8fe8104b02af82b3fc34d3e00ace2bd4 (HEAD -> master, origin/master, origin/HEAD, dalava/master)                            
Merge: 3344c26 2aee258           
pi_version: Raspberry Pi 4 Model B Rev 1.1     
os_version: Debian GNU/Linux 11 (bullseye)     
storage_space: 117G      
fits_files: 3079         
fits_missing: 0          
minutes_missing: 0   
storage_status: 4      
real_meteors: 43       
number_fits_detected: 37    
photometry_status: good   
dropped_frames:    2    
jitter_quality: 100.0%    
dropped_frame_rate: 0.0%          
last_calculated_FPS: 24.980993   
sensor : IMX291
field_of_view : 90 x 50
```

add to the archive to be uploaded.

peschman:
Yes, including camera sensor type, lens and capture method might be good.
Luc is planning to add  'protocol: udp' in config at some point in the near future.

g7gpr:
protocol udp is in test at the moment

g7gpr:
```
date                                 : 20240730_064325
commit                               : 24481635d80ec742e3a9f01bc946f803cef1dc74
merge_parents                        : 073138b2989b7039913e7833470a9291d1eb0f63
hardware_version                     : x86_64
os_version                           : Linux-5.10.0-29-amd64-x86_64-with-glibc2.36
storage_total                        : 232.73GB
storage_used                         : 166.08GB
storage_free                         : 54.76GB
fits_files                           : None
fits_missing                         : None
minutes_missing                      : None
capture_directories                  : None
detections_before_ml                 : None
detections_after_ml                  : None
number_fits_detected                 : None
number_detections_before_ml          : None
number_detections_after_ml           : None
number_fits_detected                 : None
photometry_good                      : None
dropped_frames                       : None
jitter_quality                       : None
dropped_frame_rate                   : None%
last_calculated_fps                  : None
sensor_type                          : None
lens                                 : None
protocol                             : None
media_backend                        : None
```

Blocked out the concept.

g7gpr:
```
repo                            : g7gpr/feature/summary_file
[...]

PR #335, opened 2024-08-06
Generate a summary file of information

g7gpr:
This PR is in response to issue https://github.com/CroatianMeteorNetwork/RMS/issues/316

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -19,6 +19,9 @@ from __future__ import print_function, division, absolute_import
 import os
 import sys
 import traceback
+
+import RMS.Reprocess
+
 # Set GStreamer debug level. Use '2' for warnings in production environments.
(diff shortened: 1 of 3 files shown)


dvida:
Thank you for the tremendous work!

Before I merge it, I would like to global call to the obs_summary object remoted and have a more transparent flow of information from buffered_capture instead of direct writes to a global variable in another module, which is just calling for trouble. Can we have something that instead stores the status within bufferedcapture and gets read after the capture is done (it can even be though a method in BufferedCapture), and then gets passed to the report script?

g7gpr:
Oh, the PR to this repo is meant to still be in draft; I must have taken this PR out of draft rather than the one to my repo. 

Yes, the writes to the global object are going to be replaced by setParam, and getParam. This is still at a very early stage of development. It's weeks away, at least.

g7gpr:
Closing this, the merge is getting too complicated.

PR #338, opened 2024-08-07
Prompt for Password if needed when Installing Missing Libraries

Cybis320:
## Background
`RMS_Update.sh` checks for the presence of key libraries and attempts to install them if missing. Currently, if the user lacks passwordless sudo privileges, the installation is skipped with a brief note suggesting granting such privileges. However, granting passwordless sudo access poses security risks.

## Current Situation
- `RMS_Update.sh` skips library installation if the user doesn't have passwordless sudo privileges.
- It suggests granting passwordless sudo privileges and skips installation.
- The message disappears quickly and is easy to miss.

## Proposed Changes
This PR updates `RMS_Update.sh` to implement a more secure and user-friendly approach:

1. Check for missing libraries.
2. If libraries are missing:
   a. Attempt installation without a password.
   b. If a password is required, prompt the user.
   c. Allow 30 seconds for password entry.
   d. Permit up to 3 attempts for correct password entry.
3. If password entry times out or fails after 3 attempts, skip installation and continue with the script.

## Benefits
- Provides users an opportunity to enter their password when needed.
- Eliminates the need for granting less secure passwordless sudo privileges.
- Increases visibility of potential installation issues.
- Enhances overall security while maintaining user convenience.

## Implementation Details
- Added a new function `tryPasswordlessSudo()`, and `sudoWithTimeout()` to handle password prompts and validation.
- Implemented a passwordless sudo attempt before falling back to password prompt.
- Utilized a 30-second timeout for password entry and a maximum of 3 attempts.

## Testing
- Tested on systems with and without passwordless sudo access.
- Verified correct behavior for various scenarios: successful and unsuccessful installation, password timeout, incorrect password attempts, and passwordless sudo privilege and not.

### DIFF
--- a/Scripts/RMS_Update.sh
+++ b/Scripts/RMS_Update.sh
@@ -63,13 +63,86 @@ git pull
 
 ### Install potentially missing libraries ###
 
-# Check if sudo requires a password
-if sudo -n true 2>/dev/null; then
-    sudo apt-get update
-    sudo apt-get install -y gobject-introspection libgirepository1.0-dev
-    sudo apt-get install -y gstreamer1.0-libav gstreamer1.0-plugins-bad
+# Function to check if a package is installed
+isInstalled() {
+    dpkg -s "$1" >/dev/null 2>&1
+}
+
+# Function to prompt for sudo password with timeout
+sudoWithTimeout() {
+    local timeout_duration=30  # Timeout in seconds
+    local attempts=3
+    local prompt="[sudo] password for $USER: "
+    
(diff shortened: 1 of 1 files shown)


dvida:
Excellent addition! I'd like to have someone test it first in another environment. I assigned @g7gpr to test it on his systems. I'd be happy to push it to prerelease if that helps.

g7gpr:
Hi, starting work on testing this now.

g7gpr:
Good on Debian 10,11,12. Not tested on pi as always has passwordless sudo

Issue #225, opened 2023-10-13
Invalid ECSVs files from SkyFit2


hdevillepoix:
SkyFit2's saveECSV() function does not save valid ECSV.

Problems comes from hard-coded columns declaration, not matching the CSV header line:

```
# datatype:
# - {name: datetime, datatype: string}
# - {name: ra, unit: deg, datatype: float64}
# - {name: dec, unit: deg, datatype: float64}
# - {name: azimuth, datatype: float64}
# - {name: altitude, datatype: float64}
# - {name: mag_data, datatype: float64}
# - {name: x_image, unit: pix, datatype: float64}
# - {name: y_image, unit: pix, datatype: float64}
```

 VS

```
datetime,ra,dec,azimuth,altitude,x_image,y_image,integrated_pixel_value,mag_data
```

https://github.com/CroatianMeteorNetwork/RMS/blob/c42ec82ece5b6d4a8bdbe4e618998b76358fcea2/Utils/SkyFit2.py#L5318

Easy permanent fix would be to use astropy's one-liner [Table write](https://docs.astropy.org/en/stable/io/ascii/write.html) function: `.write(ecsv_file_path, format='ascii.ecsv', delimiter=',', overwrite=True)`
Adds a dependency, but passes on the burden to Astropy to write [standard ECSV](https://github.com/astropy/astropy-APEs/blob/main/APE6.rst).
More than happy to issue a PR if the astropy dependency is acceptable.

May also build a case for having a GFE linter in https://github.com/UKFAll/standard ?

dvida:
Hi Hadrien,
Thank you for bringing this to my attention. I will try to fix it ASAP. As far as I understand, I only have to modify the header so that column names are in the correct order and that they contain all column declarations?
I would love to add astropy, but there are some old camera stations running Raspberry Pi 3s and Python 2 on which it's not possible to install astropy. Because of this I have to be very careful about introducing new dependencies. There is also an issue remotely installing new dependencies on 1000+ cameras across the world. It is automated, but if something goes wrong we have a big problem on our hands.

hdevillepoix:
Hey Denis,

Yeah as far as I can tell, for reading the file with Astropy's implementation of ECSV, it's just the header block that's the issue.


The [STIL/Topcat ECSV implementation](https://www.star.bris.ac.uk/~mbt/topcat/sun253/inEcsv.html) has got an additional issue with spaces in the numerical columns.
[...]

PR #345, opened 2024-08-12
Fix/skyfit/ecsv

g7gpr:
This pull request is in response to issue #225, and changes the formatting of ECSV files.


dvida:
I belive I fixed it myself, now the PR only suggests removing the integrated pixel value, which should stay in the ECSV file as it is a part of the GFE format. I also added a few more things.
I belive that the only issue was that the header wasn't matching the data, but I fixed that now.

PR #350, opened 2024-08-14
Silence Syntax warning

Cybis320:
Added 'r' prefix to regex patterns and 'u' prefix to Unicode strings containing special characters to ensure compatibility with both Python 2.7 and 3.x, resolving SyntaxWarnings.


Cybis320:
Hmm, standby the 'u' didn't silence it.

Cybis320:
It works now.

PR #352, opened 2024-08-15
Additions to config file

g7gpr:
These support quota management by disc use and deleting captured directories by count.

### DIFF
--- a/.config
+++ b/.config
@@ -224,27 +224,29 @@ capt_dirs_to_keep: 8
 
 ; the next settings configure the archive management by space used on disc
 
-; disabling quota management generates the reports, lists the files that would be deleted
-; but prevents any action from being taken
+; disabling quota management generates the reports, lists the files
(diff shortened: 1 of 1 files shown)


Cybis320:
We should try to keep to a 80-character line limit. The comments are exceeding that a little.

Cybis320:
This line break keeps it under 80:
```
; disabling quota management generates the reports, lists the files that would
; be deleted but prevents any action from being taken
; true = disabled, false = enabled

quota_management_disabled: true

; any one of the next three parameters missing or commented out disables the
; quota management each of the parameters is in GB.
```

Cybis320:
@g7gpr,  What is the status on this? I see that the quota management was merged into ConfigReader and DeleteOldObservation but not the associated config options. Should we reopen this PR? See issue #387

g7gpr:
Yes, we should reopen. I deleted my branch as I thought this had been merged.

PR #353, opened 2024-08-15
Utcnow fix

Cybis320:
Sorry, it was too time consuming to modify PR #313 so I recreated it here. Thank you, @adalava for your work on this.

### DIFF
--- a/RMS/Astrometry/ApplyAstrometry.py
+++ b/RMS/Astrometry/ApplyAstrometry.py
@@ -51,6 +51,7 @@ from RMS.Astrometry.CyFunctions import (cyraDecToXY, cyTrueRaDec2ApparentAltAz,
                                         cyXYToRADec,
                                         eqRefractionApparentToTrue,
                                         equatorialCoordPrecession)
+from RMS.Misc import RmsDateTime
 
 # Handle Python 2/3 compatibility
 if sys.version_info.major == 3:
@@ -876,7 +877,7 @@ def applyAstrometryFTPdetectinfo(dir_path, ftp_detectinfo_file, platepar_file, U
 
 
     # Calibration string to be written to the FTPdetectinfo file
-    calib_str = 'Calibrated with RMS on: ' + str(datetime.datetime.utcnow()) + ' UTC'
+    calib_str = 'Calibrated with RMS on: ' + str(RmsDateTime.utcnow()) + ' UTC'
 
     # If no meteors were detected, set dummpy parameters
     if len(meteor_list) == 0:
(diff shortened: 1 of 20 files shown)


dvida:
Thank you - can you confirm that it has been tested with the latest version of the prerelease branch? Or should we assign it for testing to someone who as a py2 setup?

Cybis320:
Have not tested on Py2 and did some limited testing on Py3 - nothing bad happend. I have tested that the py2 logic works on py3. Meaning if the if statement works on py2, I'm fairly confident the rest of the code won't cause issues. It does touch a lot of files but the changes are pretty minor. I'm not sure if it needs to be tested separately or if general prerelease testing is sufficient.
Maybe a quick test to confirm that this part does indeed work as expected on py2 - and I don't see why it wouldn't:
```python
sys.version_info[0]
```
I removed `import datetime` in several files when pylance indicated it wasn't accessed - I'm assuming pylance is reliable that way.

g7gpr:
Can test on py2 in a few hours.

g7gpr:
Apologies, unable to test on Python 2.x, there are too many dependencies on libraries which are not available.

Cybis320:
Do you mean that master is fine on py2 but prerelease isn't?

g7gpr:
Unfortunately I've messed up my pi3 so much, I need to re-image it, to be certain that I am testing in the correct environment.

Cybis320:
Gotcha. Any chance to narrowly test that if clause actually switches to datetime.utcnow() in a py2 env?On Aug 17, 2024, at 7:49 AM, David Rollinson ***@***.***> wrote:﻿
Unfortunately I've messed up my pi3 so much, I need to re-image it, to be certain that I am testing in the correct environment.

g7gpr:
Yes, just ran out of time yesterday.

g7gpr:
Used the following code fragment for testing

```
import datetime
import sys



class RmsDateTime:
    """ Class to hold utcnow() wrapper function definition.
        Select the best approach to retrieve current UTC time according to Python version.
    """
    if sys.version_info[0] < 3:
        @staticmethod
        def utcnow():
            # Python 2: Use the existing utcnow, which is not timezone-aware.
            return datetime.datetime.utcnow()
    else:
        @staticmethod
        def utcnow():
            # Python 3: Get timezone-aware UTC time and then make it naive.
            return datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)




print("Running under python {}.{}".format(sys.version_info.major    , sys.version_info.minor))

print("Time is {}".format(RmsDateTime.utcnow().strftime('%Y%m%d_%H%M%S.%f')))
```

Output was

```
Running under python 2.7
Time is 20240817_230222.261742

Process finished with exit code 0
```

and

```
/home/david/vRMS/bin/python3.9 /home/david/source/RMS/RMS/Astrometry/RMSDateTimeTest.py 
Running under python 3.9
Time is 20240817_230442.955765

Process finished with exit code 0
```

So for this test case, it works

Cybis320:
Thank you, Dave. I think it's probably ok to test in prerelease then.

g7gpr (approved):
Run in isolation, this looks good. However there are many other parts of the codebase which I cannot get to run in Python 2.7

PR #358, opened 2024-08-16
Feature filenameontrackstack

g7gpr:
This feature was requested by a single operator many months ago, and adds the ability to add either

0) nothing
1) the filename
2) the stationID, date and meteor count to a TrackStack

I don't believe that this functionality is required by many operator, however before I delete this branch, offering this PR.


markmac99 (approved):
Looks good. I tested it with a few conditions, -t 0, 1, 2 as well as a few chaos engineering options, -1, 999999999, and potato. 
I would suggest merging this in as it would probably be useful.

PR #361, opened 2024-08-17
Read in platepar rather than relying on existing variables

g7gpr:
Variable pp may not be set after a night of heavy cloud, therefore reading in the station platepar


dvida:
I'd love to merge it, but it has conflicts that need to be addressed.

g7gpr:
Tested working on one station in US, 10 stations in WA, including reprocessing previous corrupted nights' of observation data. One station left on prerelease showed same issue.

g7gpr:
Conflicts resolved.

PR #365, opened 2024-08-18
Update Reprocess.py to avoid error if jitter_quality unavailable

markmac99:
Handle the case when plotFFTimeIntervals() returns None which arises if the FS archive bz2 file is not available.

### DIFF
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -398,8 +398,11 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
     # Plot timestamp intervals
     try:
         jitter_quality, dropped_frame_rate, intervals_path = plotFFTimeIntervals(night_data_dir, fps=config.fps)
-        log.info('Timestamp Intervals Analysis: Jitter Quality: {:.1f}%, Dropped Frame Rate: {:.1f}%'.format(
-            jitter_quality, dropped_frame_rate))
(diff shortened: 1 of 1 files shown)


Cybis320:
Already fixed in prerelease, I believe.

markmac99:
okay, so are we alright with it crashing on live cameras? I prefer not to have crashes in "producition" :) But if prerelease is nearly ready to be merged then i guess its ok.

Cybis320:
Is it causing the production cameras to fail? Is the code not enclosed in a try-except block? It's causing the utility to crash during short test runs, which is undesirable and has been addressed in dev and prerelease branches. However, isn't the crash ultimately inconsequential? It doesn't produce an interval plot that it couldn't have produced anyway.

I'm trying to understand the urgency of this issue, but I'm concerned I'm missing some understanding.

I don't think prerelease will be ready to merge into master for another few weeks as it packs a lot of changes that will need testing.

markmac99:
yea, the code is surrounded by a try/except, but ahem exception handling is not intended to be a solution to coding errors :) 

As for effect: it caused my cameras to have errors in the logs. I check the logs for errors, so this took up my time. It also threw me off the scent of a terminal error elsewhere, which had caused RMS to crash. I do think if we find small fixable bugs we should fix them as soon a possible.

Issue #366, opened 2024-08-18
Prerelease: ObservationSummary Error on Multicam station


Cybis320:
On start capture with prerelease:
```
2024/08/18 15:54:43-ERROR-DownloadPlatepar-line:46 - Connecting to server failed!
2024/08/18 15:54:43-INFO-Reprocess-line:78 - Loaded platepar from RMS directory: platepar_cmn2010.cal
2024/08/18 15:54:43-INFO-StartCapture-line:273 - Initializing frame buffers...
2024/08/18 15:54:43-INFO-StartCapture-line:297 - Initializing frame buffers done!
Loaded 0 backed up results...
2024/08/18 15:54:44-INFO-QueuedPool-line:204 - Loaded 0 backed up results...
Using 1 cores
2024/08/18 15:54:44-INFO-QueuedPool-line:204 - Using 1 cores
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/bolide/source/RMS/RMS/StartCapture.py", line 873, in <module>
    runCapture(config, duration=duration, nodetect=cml_args.nodetect, upload_manager=upload_manager, \
  File "/home/bolide/source/RMS/RMS/StartCapture.py", line 386, in runCapture
    log.info(startObservationSummaryReport(config, duration, force_delete=False))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bolide/source/RMS/RMS/Formats/ObservationSummary.py", line 162, in startObservationSummaryReport
    repo = git.Repo(search_parent_directories=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bolide/vRMS/lib/python3.12/site-packages/git/repo/base.py", line 289, in __init__
    raise InvalidGitRepositoryError(epath)
git.exc.InvalidGitRepositoryError: /home/bolide/source/Stations/US9992
```

Cybis320:
Cleaning up *.so files solved the issue. Closing.

PR #367, opened 2024-08-18
 Prerelease: issue #366 Finding correct repo dir on multicam stations

Cybis320:
#366 The repo is not found on the parent dir on multicam stattion. Would it be sufficient to point to ~/source/RMS and wrap it in a try?


Cybis320:
Messed up the base branch standby

Issue #351, opened 2024-08-15
CryptographyDeprecationWarning: Blowfish


Cybis320:
I was wondering why we are restricting paramiko in such a way:

```python
paramiko==2.4.0,<=2.8.1; python_version=='2.7'
paramiko>=2.4.2,<=2.8.1; python_version>='3.6'
```

Paramiko 2.12.0 supports Python 2.7. Are recent paramiko versions not working for us somehow? I've installed paramiko 3.4.1, and nothing seems to break, although I haven't tried transferring files. The nice thing is that it gets rid of the deprecation warning. 

Should we find out what minimum version suppresses the deprecation warning and update the requirements accordingly? Perhaps something like:

```python
paramiko>=2.4.0,<=2.12.0; python_version=='2.7'
paramiko>=3.x.x; python_version>='3.6'
```

dvida:
Agreed, newer versions on python 3 should be tested. There was an issue that was breaking file transfers, so we degraded it. This was a few years ago.

Cybis320:
Looks like version >=2.9.0 removed the use of Blowfish and other older ciphers by default. Changing our requirements to this version will effectively update all current installations to the most recent Paramiko version compatible with the installed version of Python:
 
1. Our current requirements are:
   ```
   paramiko==2.4.0,<=2.8.1; python_version=='2.7'
   paramiko>=2.4.2,<=2.8.1; python_version>='3.6'
   ```
   This caps Paramiko at version 2.8.1 for both Python 2 and 3.

2. By changing to `paramiko>=2.9.0`, we're specifying a minimum version higher than our current maximum. This will trigger updates on all systems when RMS_update runs. 

3. The update outcomes will be:
   - Python 2 environments will install Paramiko 2.12.0 (the last version supporting Python 2.7)
   - Python 3 environments will install Paramiko 3.4.1 (the latest version as of now)

4. It eliminates the need to differentiate between Python 2 and 3 in our requirements.

5. It resolves the Blowfish deprecation warning across all our systems.

6. It makes the code more secure. 

Proposed updated requirement:
```
paramiko>=2.9.0
```

But obviously we should test that it doesn't affect file transfers negatively.

markmac99:
[...]

PR #375, opened 2024-08-21
Change requirements to paramiko>=2.9.0

Cybis320:
Addresses #351. Need to test that SSH connections still work with python 2 & 3.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 gitpython>=2.1.8
-# Use versions up to 2.4.0 due to compatibility issues with OpenSSL on Pi3/Jessie
+# Use versions 2.4.0 due to compatibility issues with OpenSSL on Pi3/Jessie
 paramiko==2.4.0; python_version=='2.7'
 paramiko>=2.9.0; python_version>='3.6'
 numpy>=1.13.3,<1.22.0 ; python_version=='2.7'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 gitpython>=2.1.8
-paramiko==2.4.0,<=2.8.1; python_version=='2.7'
-paramiko>=2.4.2,<=2.8.1; python_version>='3.6'
+paramiko>=2.9.0
 numpy>=1.13.3,<1.22.0 ; python_version=='2.7'
 numpy>=1.21.0,<1.24.0 ; python_version>='3.6' and python_version<='3.9.2'
 numpy>=1.26.0 ; python_version>'3.9.2'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,7 @@
 gitpython>=2.1.8
-paramiko>=2.9.0
+# Use versions up to 2.4.0 due to compatibility issues with OpenSSL on Pi3/Jessie
+paramiko==2.4.0; python_version=='2.7'
+paramiko>=2.9.0; python_version>='3.6'
 numpy>=1.13.3,<1.22.0 ; python_version=='2.7'
 numpy>=1.21.0,<1.24.0 ; python_version>='3.6' and python_version<='3.9.2'
 numpy>=1.26.0 ; python_version>'3.9.2'


g7gpr:
I'm trying to fit in testing this.

markmac99:
I I'm pretty sure we need to retain the existing condition for python 2.x or at least pin it to paramiko <= 2.12 because paramiko 3,x is not python 2 compatible.  I'll fire up my legacy Pi3 later today and check.

Cybis320:
I think by specifying >= 2.9.0, it'll end up installing 2.12.0 on py 2 anyway.

dvida:
The Zen of Python (PEP 20) states "Explicit is better than implicit". I suggest putting a hard <3.0 for Python 2.

Cybis320:
Denis, we certainly can do that, although I'm not sure PEP 20 is totally applicable here. The simpler approach is hardly relying on 'invisible' logic that PEP 20 is cautioning against. It simply gives pip the maximum degree of freedom to find a solution. In general, by being overly restrictive for no established reason, we may miss out on future security fixes (as it did here), needlessly restrict pip's freedom, and reduce maintainability (as it did).

I think a less-than restriction in requirements should only exist for established compatibility issues - not unproven anticipated ones. And, should be regularly revisited to see if the restriction is still necessary. 

If we were to apply the same logic of adding Python version-specific restrictions to all the other modules in requirements, we could end up with a bit of a mess. We don't currently do this for most modules even though the exact same potential for version conflicts exists.

pip will not install Paramiko 3.x on Python 2, as Paramiko 3.x currently requires Python 3.6 or newer.

That being said, it would be easy enough to add restrictions if we find it necessary (which we may shortly find out is the case). This way, we maintain simplicity and flexibility :)

markmac99:
I have to say, i'm with denis - better to be explicit especially with a legacy platform. I doubt very much that paramiko 3 will ever be backported to python2, given that it went EOL several years ago now, so constraining it to < 3.0 should suffice.

Also it avoids pip wasting time trying various versions till it finds one that works, and potentially making a poor choice. In principle pip's dependency checker is pretty smart but there's no harm in helping it.

markmac99:
Having all sorts of interesting problems testing this. I burned a new SD card with the terminal Pi3 image (dated 20191203), and booted up ok, but RMS won't update due to unable to find a version of imreg_dft, configparser and others. The image's SSL certs are also out of date and i can't update them easily as Jessie is out of support. This may take some time!

markmac99:
Got RMS working and updated, Unfortunately however updating the requirements to paramiko >=2.9.0 generates an error. I'm already on the latest compatible version of pip (20,3.4). 
```
pip install -r requirements.txt
...
Collecting paramiko>=2.9.0
  Downloading paramiko-2.12.0-py2.py3-none-any.whl (213 kB)
     |████████████████████████████████| 213 kB 3.8 MB/s
...
Collecting cryptography>=2.5
  Downloading cryptography-3.3.2.tar.gz (539 kB)
     |████████████████████████████████| 539 kB 5.7 MB/s
  Installing build dependencies ... |                                                                                        done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/pi/vRMS/bin/python /home/pi/vRMS/local/lib/python2.7/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpksFK_W
         cwd: /tmp/pip-install-fshc_P/cryptography
    Complete output (14 lines):
    running dist_info

        =============================DEBUG ASSISTANCE=============================
        If you are seeing a compilation error please try the following steps to
        successfully install cryptography:
        1) Upgrade to the latest pip and try again. This will fix errors for most
           users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
        2) Read https://cryptography.io/en/latest/installation.html for specific
           instructions for your platform.
        3) Check our frequently asked questions for more information:
           https://cryptography.io/en/latest/faq.html
        =============================DEBUG ASSISTANCE=============================

    error: 'egg_base' must be a directory name (got `/tmp/pip-modern-metadata-Ysr4C2`)
    ----------------------------------------
ERROR: Command errored out with exit status 1: /home/pi/vRMS/bin/python /home/pi/vRMS/local/lib/python2.7/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpksFK_W Check the logs for full command output.
```
i tried upgrading setuptools in case that would help, but it got worse.. The error indicates that the code for cryptography is incompatible with the version of OpenSSL thats available on Pi3/jessie.  Errors like these appeared several dozen times
```
_openssl.c:2161:19: error: expected identifier or ‘(’ before numeric constant static const long SSL_ST_RENEGOTIATE = 0;
crypto.h:499:5: note: expected ‘void * (*)(size_t)’ but argument is of type ‘void * (*)(size_t,  const char *, int)’
```
I'll fiddle about to see if i can work out a combination thats suitable, but we should probably just stick with version 2.4.0 on this platform.

Cybis320:
Thank you for your sacrifice, Mark :) It will never be forgotten. I agree we should stick to what works on Python 2.7. Now, let's see what happens with Python 3.

markmac99:
No problem :)

By the way it looks ok on Buster32, Python 3.7, Pi4. I was able to upload dummy data to my sftp server with RMS/UploadManager.py.  (I edited the file slightly to target a user on my server). 

Note that pynacl also gets updated (1.4.0 to 1.5.0).

Cybis320:
> No problem :)
> 
> By the way it looks ok on Buster32, Python 3.7, Pi4. I was able to upload dummy data to my sftp server with RMS/UploadManager.py. (I edited the file slightly to target a user on my server).
> 
> Note that pynacl also gets updated (1.4.0 to 1.5.0).

Excellent. Which version of paramiko did pip end up installing, 3.4.1? Did the Blowfish Deprecation Warning disappear?

markmac99:
3.4.1 and no blowfish warnings :)

Cybis320:
For visibility, I just want to note here that I change the PR to:
https://github.com/CroatianMeteorNetwork/RMS/blob/4cc4362a61b7f4709acd4c7cac8ceafa779b0c23/requirements.txt#L2-L4

g7gpr:
Apologies, I'm not going to be able to test this in the next few days.

Issue #376, opened 2024-08-21
No mask found even though mask is present


Cybis320:
On prerelease Multicam, with `mask.bmp` located in `~/source/Stations/[Station_ID]/` I get:
```
2024/08/21 02:43:30-INFO-DetectionTools-line:69 - No mask file has been found.
```

Cybis320:
@g7gpr, is there a change in the required folder structure? I'm also getting:
```
Starting RMS...
Using config from /home/bolide/source/Stations/US9992/.config
/home/bolide/source/Stations/US9992/.config
Disabled upload because the default station code is used!
```
The `.config` in the RMS root folder has the default station ID but the `.config` in `~/source/Stations/[Station_ID]/` is correct.

Cybis320:
In StartCapture, it's loading the config from the working dir, which is set to ~/source/RMS by the new multicam StartCapture script:
```python
# Load the config file
config = cr.loadConfigFromDirectory(cml_args.config, os.path.abspath('.'))
```

Cybis320:
The same issue is present in multiple locations: BufferedCapture, DownloadMask, EventMonitor, and StartCapture

Cybis320:
Also, ConfigReader doesn't log, it just print to the terminal, so the issue is not visible in the log.

g7gpr:
Are you passing the station name as an argument to StartCapture?

Cybis320:
Yes.

g7gpr:
I think what you are seeing is the config reader initialization where it reads the default config from . 

https://github.com/CroatianMeteorNetwork/RMS/blob/6f42cc1fd2f40b65b95e6d20a3724dcb8b900835/RMS/ConfigReader.py#L235

But this gets overwritten by the config specified on the command line. 

As for the mask, this is very similar to the error that is raised when a mask of the wrong format is used; for example the mask might have been saved with color information, rather than with the "Do not write color space option selected"

In any case, I can't replicate this issue. However, I'm happy to submit a pull request to revert the change if needed.

Cybis320:
Dave, you are right! Thank you! I connected the wrong dots here. Apologies for the extra work.

The issue actually was not that the mask.bmp was in the wrong format. It turns out 'No mask file has been found' is outputted even if a mask is found but is all white. So it's all good!

I might post a PR to make the log entry less confusing.

[...]

Issue #336, opened 2024-08-06
loadImageCalibration is called exessivley 


Cybis320:
The `loadImageCalibration` function is called twice per block. This function attempts to load the mask, flat, and dark images twice every 10 seconds. Among other things, the function checks if the mask is all white with the condition `if np.all(mask.img == 255):`.
This seems a little unnecessary - shouldn't the function only be called once per run instead?

dvida:
This is done to check whether any of those files have been updated on the fly so the capture doesn't have to be stopped. An option would be to check if they have been modified and then load them again. However, is it really a problem to read a few files every 10 seconds?

Cybis320:
I'm really not sure if it's actually a problem. Checking for all white pixels repeatedly on the same image doesn't strike me as particularly efficient. I'm also not sure if it doesn't generate unnecessary IO operations which we don't want on flash based storage. It could be adding to the storage bandwidth bottleneck. Pure speculation on my part, but worst case scenario, at 1080p we are talking about 3x2MB per 5 sec, that 1.2 MB/s. That's the transfer speed of some of our worst storage devices.
What processes would update the files on the fly? Is there anything that would update mask on the fly?

dvida:
Operators can update both files after an initial run and calibration, and there is also DownloadPlatepar and DownloadMask which can pull the files from the server. This is done when the operators cannot do it themselves.
A simpler solution would be just to modify the code to check the modification time to reduce IO.

Cybis320:
Those changes can only occur in between runs, not in the middle of a run, right?

satmonkey:
I can confirm what Denis says. The correct mask is needed for the stars count and eventually subsequent meteor detection trigger.
Creation od the mask (and platepar) is usualy done during the run, at no staturation and if stars are present.

Cybis320:
Good to know. I've always stopped and restarted capture to that - not realizing it was even an option.

Cybis320:
So, in this scenario, would it be sufficient to have the code look for the file(s) until found and then continue the run with the first found file(s)?

dvida:
Nope, as the files can be updated in the night. Tracking the modification date is the best option really.

PR #382, opened 2024-08-22
Optimize mask, flat, and dark loading with caching

Cybis320:
Optimize image calibration loading with caching and improved logging (issue #336)

- Implement ImageCalibrationLoader class for efficient file caching
- Reduce I/O operations by only reloading files when modified
- Add detailed logging for mask, dark, and flat field image loading
- Fix logging accuracy for all-white mask scenarios (issue #376)
- Implement singleton pattern for consistent cache usage across modules
- Optimize all-white mask check to avoid unnecessary rechecks

This commit should reduces strain on SD cards and hopefully speed up performance on slow storage.
Tested on MCL. Need testing on Pi 4 and 3.


Cybis320:
This PR just caches the non-bin versions. Should I cache the bins too? Not sure how often the conversions occur, but it wouldn’t be too much work.

g7gpr:
Can test on Pi4, Pi5 tonight.

Cybis320:
Thank you Dave. 
I flagged the PR as draft for the moment. It probably work fine as is but it could still be improved. 
I’ll cache the binned versions too. The cache also currently keeps old versions instead of only keeping the most current. Also, there should be a better cleanup process. 
Sorry for asking to test a bit too early.

g7gpr:
It was running fine on Debian, Pi5, 4 last night. I'll check after sunrise, happy to leave those stations there.

Cybis320:
I'm testing this version tonight. It caches all binned and non binned mask, dark, and flat files. The cache should update incrementaly if there is any changes to the files. This should reduces IO operations significantly and reduce CPU load somewhat as binning is only done when a new file is detected and checking for all white mask is only done when a new mask file is detected - as opposed to every time the files are used (which is continuously).

So far, the only testing I've done is that it doesn't crash on MCL using mask only and that it loads the correct mask.

What needs to be tested is that the cache actually updates on new files and that it works with flat and dark. It also needs to be tested that the expected files are loaded.

I'm unsure about the actual usage of dark and flat. I note that dark is disabled by default and there are currently no options exposed  in config to enable.

Cybis320:
I verified the cache does update when a mask file is updated. Now, if someone could test the flat and dark functions which I have no experience with.

PR #384, opened 2024-08-24
Fix docstrings

g7gpr:
Corrections to docstrings.


g7gpr:
Tested on Pi4, Pi5 and Debian 11

PR #394, opened 2024-08-28
Save an unprocessed frame every nth frames and generate a timelapse

Cybis320:
This PR save a video frame evert nth frames (determined by config option) in a new directory in the data directory and optionally generate a timelapse from the frames.

### DIFF
--- a/.config
+++ b/.config
@@ -198,6 +198,13 @@ captured_dir: CapturedFiles
 ; Name of directory within data_dir for archived files
 archived_dir: ArchivedFiles
 
+; Name of directory within data_dir for JPEG files
+jpg_dir: JpgFiles
+
+; Name of subdirectory within jpg_dir for JPEG files
+jpg_subdir: frames
+
+
 ; Name of directory within data_dir for log files
 log_dir: logs
 
@@ -218,12 +225,32 @@ logdays_to_keep: 30
 ; number of bz2 compressed archive folders to keep. Default 20
 bz2_files_to_keep: 20
 
-
 ; Toggle showing maxpixel on the screen during capture
 live_maxpixel_enable: false
 
 ; Toggle saving a single jpg with the latest maxpixel image in the data dir
 live_jpg: false
+
+; Toggle saving video frames to JPEG at a set interval to the jpg_dir
+save_jpgs: false
+
+; Set JPEG compression quality for the save_jpgs feature
+jpgs_quality: 90
+
+; Set the frame interval for saving video frames to JPEG. Must be greater than 5
+; CAUTION: Lower intervals can generate a large amount of data.
+; Example: An 8-hour run at 25 FPS will generate different data sizes
+;          depending on the interval and resolution selected.
+;
+;          |     Interval    |    Resolution     |
+;          | (Frames) |  (s) |   720p  |  1080p  |
+;          |----------|------|---------|---------|
+;          |        5 | 0.2  |  ~20 GB |  ~50 GB |
+;          |       25 |   1  |   ~4 GB |  ~10 GB |
+;          |       50 |   2  |   ~2 GB |   ~5 GB |
+;          |      125 |   5  | ~800 MB |   ~2 GB |
+;          |      250 |  10  | ~400 MB |   ~1 GB |
+jpgs_interval: 125
 
 ; Toggle daytime slideshow of last night's meteor detections
 slideshow_enable: false
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -33,6 +33,7 @@ from multiprocessing import Process, Event, Value
 
 import cv2
 import numpy as np
+from math import floor
 
 from RMS.Misc import ping
 from RMS.Routines.GstreamerCapture import GstVideoFile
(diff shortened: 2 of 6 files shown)


dvida:
Thanks!

A few items to address:
- The saving should be done in a separate thread. I suggest creating a new worker process which will be started and waiting for any images in the queue. My concern right now is that the frame read will be delayed if the disk IO is bad, creating timing issues. This might not be relevant for gstreamer where we have PTS, but it will be with opencv which relies on in-code timestamping.

- I suggest making feature names format-agnostic. E.g. let's say we're not happy with JPG quality and decide to change to PNGs. Our function names and arguments will be called "saveJPGs", introducing confusion.

Cybis320:
Still a rough draft. Replaced 'jpg' in variables names with 'frame'. Save frame is now sent to QueuedPool, but I'm unsure if it's actually a good idea as it would potentially double the volume of writes by pickling then saving. I think I'll post a separate issue about QueuedPool which I wonder might be too pickle happy.
I still have to implement optionally saving as png.

Cybis320:
How's this file structure and naming?:
```
~/RMS_data/
  └── XX0001/
      ├── ArchivedFiles/
      ├── CapturedFiles/
      └── FramesFiles/
          └── XX0001_yyyymmdd_hhmmss_nnnnnn/
              ├── Frames/
              │   ├── XX0001_yyyymmdd_hhmmss_nnnnnn.jpg
              │   ├── XX0001_yyyymmdd_hhmmss_nnnnnn.jpg
              │   ├── XX0001_yyyymmdd_hhmmss_nnnnnn.jpg
              │   └── ...
              └── XX0001_yyyymmdd_hhmmss_frames_timelapse.mp4
```
Struggling with the naming - totally open to suggestions.

After generating the timelapse the `Frames` dir is optionally tarballed or deleted.

dvida:
The file naming is looking good. As for the structure, we might have to think about how to keep one folder for 24/7 monitoring. There will probably have to be some restructuring. When the collection runs 24/7, we should probably pick a common time to create a day-long timelapse. E.g. from sunrise to sunrise, or simply from 00:00 UT + 24 hours.

Cybis320:
> The file naming is looking good. As for the structure, we might have to think about how to keep one folder for 24/7 monitoring. There will probably have to be some restructuring. When the collection runs 24/7, we should probably pick a common time to create a day-long timelapse. E.g. from sunrise to sunrise, or simply from 00:00 UT + 24 hours.

A possible option is to store the frames in a year/day_number/hour folder structure. 2024/267/15, would store 09/23/2024 from 1500 to 1559 UTC.

Cybis320:
Something like this?:
```
~/RMS_data/
└── XX0001/
    ├── ArchivedFiles/
    ├── CapturedFiles/
    └── FramesFiles/
        ├── 2024/
        │   ├── 001/
        │   │   ├── 00/
        │   │   │   ├── XX0001_20240101_000001_122577.jpg
        │   │   │   ├── XX0001_20240101_000011_122585.jpg
        │   │   │   └── ...
        │   │   ├── 01/
        │   │   │   ├── XX0001_20240101_010001_122577.jpg
        │   │   │   ├── XX0001_20240101_010011_122585.jpg
        │   │   │   └── ...
        │   │   ├── ...
        │   │   └── 23/
        │   │       ├── XX0001_20240101_230001_122577.jpg
        │   │       ├── XX0001_20240101_230011_122585.jpg
        │   │       └── ...
        │   ├── ...
        │   ├── 365/
        │   └── 366/
        ├── 2025/
        │   ├── 001/
        │   │   ├── 00/
        │   │   │   ├── XX0001_20250101_000001_122577.jpg
        │   │   │   ├── XX0001_20250101_000011_122585.jpg
        │   │   │   └── ...
        │   │   ├── 01/
        │   │   ├── ...
        │   │   └── 23/
        │   ├── ...
        │   └── 365/
        └── ...
```

dvida:
I recommend avoiding uninformative structures such as the day count, as they will only introduce confusion. Python's datetime library can easily provide the proper dates. Is there an advantage to using day numbers?

Cybis320:
The main advantage is that it matches how satellite (including GOES) imagery is organized.

One folder per hour is also efficient for cloud-based storage, where folders containing too many files are often slow.

But I agree that Day of Year (DOY) is not particularly human-friendly—it’s machine-friendly though. I think it’s popular with satellite operators because it makes interval calculation easy (they don’t have to worry about leap years).

dvida:
Why not both? E.g. 20240923-267_HHMMSS

Structure:
```
        ├── 2024/ (year)
        │   ├── 202409/ (month)
        │   │   ├── 20240923-267/ (day)
        │   │   │   ├── 20240923-267_00/ (hour)
```

I recommend including all previous parent dirs in the name of all children for full transparency. I do this in the GMN and this allows automated sorting of directories by time even if they're copied out of the expected structure.

Cybis320:
That’s a good idea and you make a good point about not loosing track when using folders in isolation.Is the month layer necessary? Dropping it would match the GOES structure and still give us all the context we need. Structures:        ├── 2024/
        │   ├── 20240923-267/
        │   │   ├── 20240923-267_00/
I know that once we settle on a scheme and start accumulate data, it’s not easy or cheap to make changes in the cloud. On Sep 24, 2024, at 9:32 PM, Denis Vida ***@***.***> wrote:﻿
Why not both? E.g. 20240923-267_HHMMSS
Structures:
        ├── 2024/ (year)
        │   ├── 202409/ (month)
        │   │   ├── 20240923-267/ (day)
        │   │   │   ├── 20240923-267_00/ (hour)

I recommend including all previous parent dirs in the name of all children for full transparency. I do this in the GMN and this allows automated sorting of directories by time even if they're copied out of the expected structure.

dvida:
For sure, dropping the month works and I guess it will also make finding the right day by the day number easier if all directories are together.

PR #395, opened 2024-08-29
Fix loading operational mask when computing flux

dvida:
All raw collecting areas on operational systems have been computed without masks. This PR fixes that, hopefully resulting in correct fluxes.

### DIFF
--- a/RMS/Formats/Platepar.py
+++ b/RMS/Formats/Platepar.py
@@ -1485,7 +1485,7 @@ def findBestPlatepar(config, night_data_dir=None):
     """
 
     # Get the platepar (first from the night directory, then the default one)
-    platepar = Platepar.Platepar()
+    platepar = Platepar()
     
     if night_data_dir is not None:
         night_platepar_path = os.path.join(night_data_dir, config.platepar_name)
--- a/RMS/Formats/Platepar.py
+++ b/RMS/Formats/Platepar.py
@@ -1495,11 +1495,15 @@ def findBestPlatepar(config, night_data_dir=None):
     if (night_data_dir is not None) and os.path.exists(night_platepar_path):
         platepar.read(night_platepar_path, use_flat=config.use_flat)
 
+        return platepar
+
     elif os.path.exists(default_platepar_path):
         platepar.read(default_platepar_path, use_flat=config.use_flat)
 
-    else:
-        return None
+        return platepar
+
+    
+    return None
 
 
 if __name__ == "__main__":
--- a/RMS/Formats/Platepar.py
+++ b/RMS/Formats/Platepar.py
@@ -1470,6 +1470,38 @@ class Platepar(object):
         return out_str
 
 
+def findBestPlatepar(config, night_data_dir=None):
+    """ Try loading the platepar from the night data directory, and if it doesn't exist, load the default one
+    from the config directory.
+    
+    Arguments:
+        config: [Config] Configuration object.
+
+    Keyword arguments:
+        night_data_dir: [str] Path to the night data directory. None by default.
+
+    Return:
+        platepar: [Platepar] Loaded platepar object. None if not found.
+    """
+
+    # Get the platepar (first from the night directory, then the default one)
+    platepar = Platepar.Platepar()
+    
+    if night_data_dir is not None:
+        night_platepar_path = os.path.join(night_data_dir, config.platepar_name)
+
+    default_platepar_path = os.path.join(config.config_file_path, config.platepar_name)
+
+    if (night_data_dir is not None) and os.path.exists(night_platepar_path):
+        platepar.read(night_platepar_path, use_flat=config.use_flat)
+
+    elif os.path.exists(default_platepar_path):
+        platepar.read(default_platepar_path, use_flat=config.use_flat)
+
+    else:
+        return None
+
+
 if __name__ == "__main__":
 
     import argparse
--- a/Utils/RecomputeCollectionAreas.py
+++ b/Utils/RecomputeCollectionAreas.py
@@ -43,6 +43,9 @@ def recomputeCollectionAreas(root_dir_path, ncores=1):
 
     dir_list = []
 
+    print()
+    print("Building the list of directories...")
+
     # Go through all directories in the STATIONID/NIGHT/ directory structure
     for station_id in sorted(os.listdir(root_dir_path)):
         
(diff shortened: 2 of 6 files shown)


Cybis320:
Also, I think there is a similar issue with the platepar.

dvida:
I added explicitly passing a platepar. However, a platepar is always available in the captured files directory, so that's fine. Otherwise, things would simply not work and the fluxes would not make any sense whatsoevery.

Cybis320:
This still risks loading the wrong platepar, no?
```python
    if platepar is None:
        platepar = Platepar.Platepar()
        platepar.read(os.path.join(dir_path, config.platepar_name), use_flat=config.use_flat)
```
`platepar = Platepar.Platepar()` will create a default platepar with default values. If for whatever reasons there is no platepar in `dir_path`, those default values will not be updated. And there will be no platepar in dir_path is the recalibration fails.

dvida:
I'm considering this a high priority fix. The platepar doesn't matter here as there will be no flux if the recalibrated platepars don't match the observations, the mask is the only critical thing here.

This works in my local tests. I'd like to merge this ASAP.

I'm recomputing all files on the server, hopefully we'll have updated flux profiles soon.

markmac99:
I'll do my best to test this asap but see private message.

dvida:
@Cybis320 if you look upstream in processNight, the command is not run if a platepar is not available or loaded.

However, I updated the code to make a check in the default dir second. I prefer the night dir first as the platepars might change over time so the current one might not always be appropriate for reprocessing.

Cybis320:
> @Cybis320 if you look upstream in processNight, the command is not run if a platepar is not available or loaded.
> 
> However, I updated the code to make a check in the default dir second. I prefer the night dir first as the platepars might change over time so the current one might not always be appropriate for reprocessing.

Looks great! Thank you for humoring me :) Also, I can reuse the findBestPlatepar in my adsb code, so that's great too.

Cybis320:
The findBestPlatepar is missing a couple 'return', maybe?

RMS/Formats/Platepar.py
```diff
@@ -1470,6 +1470,38 @@ def __repr__(self):
...
+    platepar = Platepar.Platepar()
+    
+    if night_data_dir is not None:
+        night_platepar_path = os.path.join(night_data_dir, config.platepar_name)
+
+    default_platepar_path = os.path.join(config.config_file_path, config.platepar_name)
+
+    if (night_data_dir is not None) and os.path.exists(night_platepar_path):
+        platepar.read(night_platepar_path, use_flat=config.use_flat)
+
```
Cybis320, line 1497: return platepar? and line 1500 too.

dvida:
sorry, working through meetings and gaps in time. it should be fixed now. thanks!

g7gpr:
Testing on Debian 10,11,12, Pi4, Pi5

Utils/RecomputeCollectionAreas.py
```diff
@@ -0,0 +1,118 @@
...
+        # Load the config file
+        config = loadConfigFromDirectory(['.'], dir_path)
+
+        # Load the platepar
+        platepar = Platepar.Platepar()
+        platepar.read(os.path.join(dir_path, config.platepar_name), use_flat=config.use_flat)
+
+        # Load the mask
+        mask = getMaskFile(dir_path, config)
+
```
Cybis320, line 26: Are we not repeating the initial error here? Looking for platepar and mask in dir_path? Or is it meant to be run on a different folder structure?

dvida: When running live, the code will already pass the correct mask and the platepar. When post-processing, RMS has to use the platepar and mask in the directory. E.g. imagine I'm reprocessing data from 2 years ago when the mask and platepar were different.

dvida: I see the potential confusion - the directories this is made to reprocess are ArchivedFiles directories, not CapturedFiles.

Cybis320: Oh, I see. Running this code on the Archived folder is fine. As long as the folder contains the appropriate platepar and mask without fail the code is fine.

markmac99:
How should i be testing this? I checked out the PR to my test Pi4 and ran RMS.Reprocess on the ArchivedFiles folder for a dataset from May 2024, but the flux files were not updated.  The dataset contains 41 meteors.

Cybis320:
I think mostly it's running reprocess on a CaptureFiles directory that needs testing. Whether it picks up the correct mask and platepar. For instance, if the correct mask is not all white but you see an all-white message in the console that would indicate that the wrong mask is used. 
Another interesting test is whether flux picks the correct platepar even if missing from the night dir. that's trickier to notice through. The way I noticed is that if the correct platepar is 1080p but flux picks a default 720p platepar, it generate a message. 
(Sorry typing on my cellphone)

markmac99:
I've tested running RMS.Reprocess against both a CapturedFiles folder and an ArchivedFiles folder without error.  However, i am not sure what i should expect or how i'd know if it was broken! 

As there is no mask in the CapturedFiles folders, the mask in the RMS source folder must be being used.  I don't see any "all white" messages, even though the mask is the default all white. But this seems to be normal behaviour. 

There is a mask in the ArchivedFiles folder, but cannot confirm whether it was loaded in preference to the default mask as the code just prints "loaded mask; mask.bmp" at various places, without mentioning where it was loaded from :) But again this seems to be normal! 

The flux files were also unchanged (datestamp didn't update). I think i was expecting these to be updated, even if the mask was the same as before? 

So unfortunately i am not sure what this change is doing, or how i confirm it has worked.

Cybis320:
I think, the `flux_col_areas_*.json` file should first be renamed for the file to be recalculated. The resulting json should be different if the original was computed with an incorrect all-white mask and the correct mask is non-white.

dvida:
There were some issues with computing the masked ratio on some versions of numpy, I fixed that now.

How to test: 
Choose a camera that has a portion of the FOV masked out (preferably a good chunk of one corner). Run the normal capture. Next day, check the flux_col_areas*.json file and see if the computed area for the masked corner is zero.

The old version of the code ignored the mask, so the collecting areas were also full. That's probably why the flux had those deep dips.


I tested a reprocessing run locally without any issues. I'll test it on a live machine and hopefully merge it tomorrow.


The server is recomputing all previously computed areas.

PR #396, opened 2024-08-29
Load pp and mask from same dir as config

Cybis320:
Here's another way to go about it.

### DIFF
--- a/Utils/Flux.py
+++ b/Utils/Flux.py
@@ -1208,11 +1208,10 @@ def detectClouds(config, dir_path, N=5, mask=None, show_plots=True, save_plots=F
 
     # Get the platepar
     platepar = Platepar.Platepar()
-    platepar.read(os.path.join(dir_path, config.platepar_name), use_flat=config.use_flat)
-
+    platepar.read(os.path.join(config.config_file_path, config.platepar_name), use_flat=config.use_flat)
 
     # Locate and load the mask file
-    mask = getMaskFile(dir_path, config, file_list=file_list)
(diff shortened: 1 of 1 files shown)


g7gpr:
Not a fan of this. Mask in Captured directory should always be used in preference, as camera mask in config dir may have changed.

Cybis320:
Problem is I don't see a platepar is the capture directory.

Cybis320:
We could extract a platepar from the json in the capture dir.

Cybis320:
Actually, no mask or platepar in the capture dir. At least, not in prerelease MCL.

g7gpr:
Ah, of course, it gets copied in at the end of capture.

Cybis320:
Not on my end for some reasons. No mask or platepar here in capture dirs. Flux then use no mask and default platepar values from platepar.py with no warnings. I'll look at why they are not there. If you have the code doing the copying handy, that would save me looking for it.

Cybis320:
Looking at other stations running master, I do see the platepar but no mask in Capture. Investigating why I don't have a platepar on my end. In any case, the code shouldn't be allowed to use the default platepar values if no platepar is found.

Cybis320:
Found the issue I think:
If fit_status is False, then the platepar doesn't get copied and flux subsequently use the wrong platepar.

https://github.com/CroatianMeteorNetwork/RMS/blob/441008b356fa203df26d8c5094c82797de5cf7bf/RMS/Reprocess.py#L214-225

Cybis320:
Closing. see #395.

PR #410, opened 2024-09-06
save kml files with decimal area_ht in filename

MaadhyamRana:
This PR updates the storage format for the FOV slice area height in both the KML file and the filename. The height is now stored as a decimal (rounded to 3 decimal places), rather than as an integer.

### DIFF
--- a/Utils/FOVKML.py
+++ b/Utils/FOVKML.py
@@ -14,7 +14,7 @@ from RMS.Routines.MaskImage import loadMask
 from RMS.Routines.FOVArea import fovArea
 
 
-def fovKML(dir_path, platepar, mask=None, area_ht=100000, side_points=10, plot_station=True):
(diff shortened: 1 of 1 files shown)


dvida:
Can you make the decimal points optional? The exact original functionality should be preserved using the default function arguments.

Cybis320:
I wonder if we should start thinking about adding a feature to limit the range of the  FOV to realistic values. The bottom corners of some of the FOVs is showing very optimistic values. It's all calculated correctly in the sense that the line of sight intersect the height at the correct distance, but very low angle lines are traversing the densest part of the atmosphere and are getting attenuated very strongly. Also the cameras probably can't resolve that kind of details at those distances even in perfectly clear air.

dvida:
@Cybis320  - I belive this would involve limiting the bottom elevation. It's possible, but I think it's lens and usage-dependent. E.g. for fireballs which can be caught >400 km away on all-sky cameras, having coverage at elevations near the horizon is desirable. For contrails which presumably need a minimum resolution to be resolved, it's probably higher.

Issue #411, opened 2024-09-07
Observation Summary - stations operating without effective time sync


g7gpr:
At least one station has been running without time sync
Add some feature into observation summary so that the time sync platform and status is reported

PR #413, opened 2024-09-07
Report on time sync status

g7gpr:
Issue #411 notes that at least one station has been operating without valid time sync. This pr includes time sync status into the observation summary report, and a very approximate estimation of error that could be used to recover data at the server.


g7gpr:
Tested on Pi5 platform and Debian

g7gpr:
```
clock_error_seconds             : 0.9
clock_synchronized              : yes
```

PR #422, opened 2024-09-14
updated to shared memory method for raw frame saving

MaadhyamRana:
This PR updates the raw frame saving method to use dual shared memory buffers instead of RMS.QueuedPool.

The buffers switch and save frames to disk when there are 10 raw frames available (though this is configurable in RMS/BufferedCapture.py).

Additionally, a JSON file is created with every run of BufferedCapture to make timelapse frame number and corresponding timestamp pairs, so that it's easy to work with such data going forward.
 
Issues: The code currently stores an extra blank file for every 10th raw frame saved. This is likely due to how leftover raw frames from the buffer are processed in [line 1318] of BufferedCapture.py

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -18,8 +18,8 @@ from __future__ import print_function, division, absolute_import
 
 import os
 import sys
+import ctypes
 import traceback
-import RMS.Reprocess
 
 # Set GStreamer debug level. Use '2' for warnings in production environments.
 os.environ['GST_DEBUG'] = '3'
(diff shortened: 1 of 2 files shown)


MaadhyamRana:
Above issues with blank frame saving have been fixed

Cybis320:
Great work! Just started looking through it. 
Would it be possible to not rely on fps to determine timestamps. I'd feel better if the timestamps for each frame are determined by the BufferedCapture timestamps. (the config.fps is not the most accurate.)

MaadhyamRana:
Done, it now uses another shared array for handling timestamps.
Tested on the New Mexico camera with no issues.

Cybis320:
Sweet! Thank you!

dvida:
Great job on this. Before we proceed, I'd like to see a few changes.
First, we should enable saving frames by default so we can start collecting a lot of data for training ASAP.

Next, the frame save interval should be defined in seconds instead of frames. E.g. some people might run cameras at different FPSs, e.g. 25/30/50/60, and we probably want uniform cadence across the network.

Once that is implemented, I'll merge it.

Cybis320:
Good point, I’ll take care of the time vs frame interval.

PR #423, opened 2024-09-15
Extension to EventMonitor to allow stars, or any RaDec point to be tracked between two julian dates

g7gpr:
This PR adds the ability to monitor a specific radec (j2000) coordinate set, and from the observations, produce a json file with information about the magnitude of any observations at those coordinates. This json contains image data, which is rendered as thumbnails, and also magnitude plots. The data is derived from searches not only of the CapturedFiles directories, but also the ArchivedFiles directories.
r_138 3_d_-69 72_jd_2460567 0_2460568 0_AU0006_magnitude
r_138 3_d_-69 72_jd_2460567 0_2460568 0_AU0006_contact_sheet
There are two ways to specify events. 
```
StarRa                   : 347.3
StarDec                  : -7.8
JDStart                  : 2460567.0
JDEnd                    : 2460567.3
UseCalstar               : 1
Suffix                   : saturn

END
```
This uses the calstar file to plot magnitudes, and locks the thumbnail plots to the nearest observed object. However, when the calstar file is not available, the UseCalstar can be set to zero.

r_239 9_d_25 9_jd_2460563 0_2460567 0_AU000U_contact_sheet
```
StarRa                   : 239.9
StarDec                  : 25.9
JDStart                  : 2460563.0
JDEnd                    : 2460567.0
UseCalstar               : 0
Suffix                   : T_CrB

END
```

Running 
```
python -m RMS.Astrometry.VariableStar
```

Will load all the data from the Calstars files in ArchivedDirectories into an sqlite database in ~/RMS_data/magnitudes.db, with each observation linked to the catalogue key of the local star database. 

If this is run again, it will import the observations since the latest entry into the database.


g7gpr:
Tested on Debian and Pi4

PR #424, opened 2024-09-20
Define .config Path Relative to RMS Root

Cybis320:
Currently, RMS can only function when the working directory is the RMS root directory because the `.config` path is defined relative to the working directory in the .pyxbld files. This is inconvenient when, for instance, we want to use RMS functions in other projects. This PR defines the `.config' path relative to the RMS root directory regardless of the actual working directory path.

### DIFF
--- a/RMS/Astrometry/CyFunctions.pyxbld
+++ b/RMS/Astrometry/CyFunctions.pyxbld
@@ -5,9 +5,13 @@ def make_ext(modname, pyxfilename):
     # Use extra compile arguments for this Cython
     import RMS.ConfigReader as cr
     from distutils.extension import Extension
+    from RMS.Misc import getRmsRootDir
+    import os
 
     # Load the configuration file
-    config = cr.parse(".config")
+    config_filename = '.config'
+    config_path = os.path.join(getRmsRootDir(), config_filename)
+    config = cr.parse(config_path)
 
     # Use additional compile arguments
     ext = Extension(name = modname,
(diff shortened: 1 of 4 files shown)


markmac99 (commented):
Is this going to conflict wth the multicamera build, which locates the config file in a camera-specific folder?

satmonkey:
it seems so. In multi-station configuration, the .config path is passed as parameter -c, which most of RMS functions understand.

markmac99:
> it seems so. In multi-station configuration, the .config path is passed as parameter -c, which most of RMS functions understand.

True. And I guess the pyxbld stuff already reads from the default config in these situations so at least this will ensure it is consistent.

Cybis320:
No. Actually, in multicam the pyxbld files are still grabbing the config file from the working directory not the station’s. On Sep 23, 2024, at 8:10 AM, Mark McIntyre ***@***.***> wrote:﻿
@markmac99 commented on this pull request.

Is this going to conflict wth the multicamera build, which locates the config file in a camera-specific folder?

Cybis320:
To clarify, the PR does not affect how multicam or single cam operate. 

Currently, both single and multicam setups use the built parameters specified in the config file located in the working directory, no matter what other config file might be specified at launch. 

The PR only change is that the built parameters will be coming from the config file in the RMS root directory not the working directory.

satmonkey:
Then, there is no problem with the single user multicam. The current
workflow starts with creating the single camera at standard location, and
then cloning the .config for multicam. The build info remains untouched.

Dne po 23. 9. 2024 18:20 uživatel Luc Busquin ***@***.***>
napsal:

Cybis320:
Without the PR, the pyxbld files use a hardcoded  `./.config`. It causes an error If any RMS module is accessed when the working directory does not contain a .config file. 

The use case for the PR is when trying to work with RMS module outside our standard environment. The PR has no impact on current single or multicam operations.

markmac99:
All makes sense. Thanks.

PR #425, opened 2024-09-22
Don't log GST imported status at logger initialization

Cybis320:
Reduce logging verbosity when GStreamer cannot be imported. This PR prevent logging GST imported status at logger initialization and remove some f-strings.


markmac99 (approved):
looks good.  I have run some tests on my PC without any problems.

PR #426, opened 2024-09-22
Add support for VideoFormat, ClearFog, and BroadTrends Camera Settings

Cybis320:
This PR adds the following settings to `RMS_SetCameraParams.sh`:
```
python -m Utils.CameraControl SetParam General VideoFormat PAL
python -m Utils.CameraControl SetParam Camera ClearFog enable 0
python -m Utils.CameraControl SetParam Camera ClearFog level 50
python -m Utils.CameraControl SetParam Camera BroadTrends AutoGain 0
python -m Utils.CameraControl SetParam Camera BroadTrends Gain 50
```
Setting VideoFormat to PAL ensures that the camera is not accidentally set to NTSC, which would affect the evenness of the frame rate.

**ClearFog** and **BroadTrends** are set to their default values in case they were changed to other settings for day observation. Setting **BroadTrends AutoGain** to 0 allows preservation of highlights during day observation.

The PR also changes the name and format of the resulting JSON files saved with **CameraControl SaveSettings**. The JSON files are now indented. An additional `location.json` file is created to save the VideoFormat. The net#.json files are renamed to more explicit names.

### DIFF
--- a/Scripts/RMS_SetCameraParams.sh
+++ b/Scripts/RMS_SetCameraParams.sh
@@ -27,6 +27,7 @@ python -m Utils.CameraControl CameraTime set
 python -m Utils.CameraControl CloudConnection off
 
 # set the Video Encoder parameters
+python -m Utils.CameraControl SetParam General VideoFormat PAL
 python -m Utils.CameraControl SetParam Encode Video Compression H.264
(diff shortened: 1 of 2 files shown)


markmac99 (approved):
Looks good to me. Useful additions

PR #427, opened 2024-09-24
continuous daytime capture

MaadhyamRana:
This PR adds continuous daytime capture feature to the capture code. When enabled, the code runs capture throughout day and night, while also switching between the implemented daytime and nighttime modes within CameraControl.py.

### DIFF
--- a/.config
+++ b/.config
@@ -232,7 +232,7 @@ capt_dirs_to_keep: 8
 frame_dirs_to_keep: 8
 
 ; number of raw video folders (days) to keep. Default is 2
-video_dirs_to_keep = 2
+video_dirs_to_keep: 2
 
 ; the next settings configure the archive management by space used on disc
 
--- a/.config
+++ b/.config
@@ -115,12 +115,8 @@ gst_colorspace: BGR
 ; Examples: decodebin, avdec_h264, nvh264dec
 gst_decoder: avdec_h264
 
-; Location of the raw videos to be saved to disk (None means no saving).
-raw_video_dir: None
-
-; Save the raw video to the night directory.
-; If True, the raw_video_dir is ignored.
-raw_video_dir_night: False
+; Toggle raw video saving in data_dir/video_dir
+raw_video_save: true
 
 ; Duration of the raw video segment (seconds).
 ; If -1, the segment duration will be 256/FPS seconds
@@ -198,6 +194,12 @@ captured_dir: CapturedFiles
 ; Name of directory within data_dir for archived files
 archived_dir: ArchivedFiles
 
+; Name of directory within data_dir for saved frame files
+frame_dir: FramesFiles
+
+; Name of directory within data_dir for raw video files
+video_dir: VideoFiles
+
 ; Name of directory within data_dir for log files
 log_dir: logs
 
@@ -226,6 +228,11 @@ bz2_files_to_keep: 20
 ; number of CapturedFiles folders to keep. Default is 8
 capt_dirs_to_keep: 8
 
+; number of frame folders (days) to keep. Default is 8
+frame_dirs_to_keep: 8
+
+; number of raw video folders (days) to keep. Default is 2
+video_dirs_to_keep = 2
 
 ; the next settings configure the archive management by space used on disc
 
@@ -260,6 +267,34 @@ live_maxpixel_enable: false
 ; Toggle saving a single jpg with the latest maxpixel image in the data dir
 live_jpg: false
 
+; Toggle saving video frames at a set interval to the frame_dir
+save_frames: true
+
+; The file extension for saved frames ('jpg' or 'png')
+frame_file_type: jpg
+
+; Set the saved framed JPEG quality for jpg file type [0-100]
+jpgs_quality: 90
+
+; Set the saved framed PNG compression for png file type [0-9]
+png_compression: 3
+
+;  Set the time interval in seconds for saving video frames.
+; CAUTION: Lower intervals can generate a large amount of data.
+; Example: An 8-hour run at 25 FPS will generate different data sizes
+;          depending on the interval and resolution selected.
+;
+;          | Inter|    Resolution     |
+;          |  (s) |   720p  |  1080p  |
+;          |------|---------|---------|
+;          | 0.2  |  ~20 GB |  ~50 GB |
+;          |   1  |   ~4 GB |  ~10 GB |
+;          |   2  |   ~2 GB |   ~5 GB |
+;          |   5  | ~800 MB |   ~2 GB |
+;          |  10  | ~400 MB |   ~1 GB |
+frame_save_interval: 5
+
+
 ; Toggle daytime slideshow of last night's meteor detections
 slideshow_enable: false
 
@@ -287,7 +322,7 @@ postprocess_at_end: false
 postprocess_delay: 0
 
 ; Ran capture continuously, through day/night
-continuous_capture: false
+continuous_capture: true
 
 [Build]
 
@@ -702,6 +737,9 @@ stack_mask: false
 
 ; Automatically generate a timelapse using all FF files
 timelapse_generate_captured: true
+
+; Automatically generate a timelapse using all saved frame files
+timelapse_generate_from_frames: true
 
 
 
--- a/.config
+++ b/.config
@@ -286,7 +286,8 @@ postprocess_at_end: false
 ; postprocess_at_end is false, or after capture ends if true.
 postprocess_delay: 0
 
-
+; Ran capture continuously, through day/night
+continuous_capture: false
 
 [Build]
 
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -81,76 +81,38 @@ class BufferedCapture(Process):
         """
         
         super(BufferedCapture, self).__init__()
+        
+        # Store configuration and paths (immutable data is safe to pass to child process)
+        self.config = config
+        self.video_file = video_file
+        self.night_data_dir = night_data_dir
+        self.saved_frames_dir = saved_frames_dir
+        self.daytime_mode = daytime_mode
+
+        # Store shared memory arrays and values for compressor (these are designed for multiprocessing)
         self.array1 = array1
         self.startTime1 = startTime1
         self.array2 = array2
         self.startTime2 = startTime2
-        
         self.startTime1.value = 0
         self.startTime2.value = 0
-        
-        self.config = config
-        self.media_backend_override = False
-        self.video_device_type = "cv2"
 
-        self.video_file = video_file
-
-        self.night_data_dir = night_data_dir
-        self.saved_frames_dir = saved_frames_dir
-
-        # A frame will be considered dropped if it was late more then half a frame
-        self.time_for_drop = 1.5*(1.0/config.fps)
-
-        # Initialize Smoothing variables
-        self.startup_flag = True
-        self.last_calculated_fps = 0
-        self.last_calculated_fps_n = 0
-        self.expected_m = 1e9/self.config.fps
-        self.reset_count = -1
-
-        self.dropped_frames = Value('i', 0)
-        self.device = None
-        self.pipeline = None
-        self.start_timestamp = 0
-        self.frame_shape = None
-        self.convert_to_gray = False
-
-        # Params to help with camera mode switching
-        self.daytime_mode = daytime_mode
-
+        # Initialize shared values for raw frame saving (these are designed for multiprocessing)
         if self.config.save_frames:
 
-            # Params for raw frame saving
+            # Frame saving block size - these many raw frames are written to buffer before saving to disk
             self.num_raw_frames = 10
-            self.raw_frame_count = 0
 
-            # Initialize shared values for frame + timestamp saving
             self.startRawTime1 = Value('d', 0.0)
             self.startRawTime2 = Value('d', 0.0)
-
-
-            # Initialize shared memory arrays for raw frame saving
-            self.sharedRawArrayBase = Array(ctypes.c_uint8, self.num_raw_frames * config.height * config.width)
-            self.sharedRawArray = np.ctypeslib.as_array(self.sharedRawArrayBase.get_obj())
-            self.sharedRawArray = self.sharedRawArray.reshape(self.num_raw_frames, config.height, config.width)
-
-            self.sharedRawArrayBase2 = Array(ctypes.c_uint8, self.num_raw_frames * config.height * config.width)
-            self.sharedRawArray2 = np.ctypeslib.as_array(self.sharedRawArrayBase2.get_obj())
-            self.sharedRawArray2 = self.sharedRawArray2.reshape(self.num_raw_frames, config.height, config.width)
-
-
-            # Initialize shared memory for timestamp saving
             self.sharedTimestampsBase = Array(ctypes.c_double, self.num_raw_frames)
-            self.sharedTimestamps = np.ctypeslib.as_array(self.sharedTimestampsBase.get_obj())
-            
             self.sharedTimestampsBase2 = Array(ctypes.c_double, self.num_raw_frames)
-            self.sharedTimestamps2 = np.ctypeslib.as_array(self.sharedTimestampsBase2.get_obj())
 
-            self.raw_frame_saver = RawFrameSaver(self.saved_frames_dir, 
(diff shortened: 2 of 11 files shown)


Cybis320:
This is great. Can't wait to try it out.

dvida:
This is a great start! As suggested by @Cybis320 (I presume the message was edited), there should be a fundamental difference between nighttime and daytime capture. Perhaps BufferedCapture could be modified so it switches between the two modes on an external call (e.g. bc.switchDaytimeCapture() and bc.switchNighttimeCapture()).
The nighttime capture should remain unchanged, but the daytime capture should only capture those individual frames and optionally capture the video to disk.
This will keep the current diurnal cycle and keep the meteor capture unaffected. The new daytime frame capture will collect contrail data during the day and the daytime timelapse can be uploaded with the daily uploads. The daytime video capture will enable daytime fireball observations and perhaps used for other things too.

Cybis320:
Something like that?
```mermaid
graph TD
    A[StartCapture] --> B[Initialize RTSP Stream]
    B --> C[Start Pipeline Sinks]
    C --> I[App Sink]
    C --> J[ MKV Video Sink]
    J --> N[Start MKV Making]
    I --> M[Save every Nth Frame]
    I --> D{Day or Night?}
    D -->|Day| E[Stop Compressor]
    E --> F[Set Day Camera Settings]
    D -->|Night| G[Set Night Camera Settings]
    G --> H[Start Compressor]
    H --> L[Detection]
    F --> K[Time Delay]
    L --> K
    K --> D
```

dvida:
That's a great flowchart, that should exactly be it!

Cybis320:
> That's a great flowchart, that should exactly be it!

That might be my new fav thing to do :) I just discovered you can do this:

(replace ' with ` to actually display the chart)
'''mermaid
graph TD
    A[StartCapture] --> B[Initialize RTSP Stream]
    B --> C[Start Pipeline Sinks]
    C --> I[App Sink]
    C --> J[ MKV Video Sink]
    J --> N[Start MKV Making]
    I --> M[Save every Nth Frame]
    I --> D{Day or Night?}
    D -->|Day| E[Stop Compressor]
    E --> F[Set Day Camera Settings]
    F --> X[Post Process]
    D -->|Night| G[Set Night Camera Settings]
    G --> H[Start Compressor]
    H --> L[Detection]
    X --> K[Time Delay]
    L --> K
    K --> D
'''

MaadhyamRana:
Needs testing, but the updates are summarized below.

**.config and ConfigReader.py:**

- Includes options for _continuous capture, raw video saving, raw frame saving_ and _timelapse generation_ from the raw frames. These are enabled by default for this version.

- Raw frames go to ~/RMS_data/FramesFiles by default
- Raw videos go to ~/RMS_data/VideoFiles by default
- Raw frame interval (in seconds), quality, and file type (jpg/png) is configurable


**BufferedCapture.py and RawFrameSave.py:**

- During daytime, compression is not performed so the shared arrays (with the Compressor) aren't accessed
- A similar shared memory setup to the Compressor is implemented for frame saving (RawFrameSave.py), saving one frame every defined interval (intended to run regardless of day/night). Frames and corresponding timestamps are transferred through this array 
- Accurate timestamping of raw videos is still a WIP. Currently, raw video clip saving have been tied to a custom gstreamer callback function (move_segment), which moves every video clip to its correct path under VideoFiles. TimestampRMS does not temporarily work as a result.
- RawFrameSaver saves frames and their corresponding frame count / timestamps in 10 frame blocks (can be configured in BufferedCapture.py)
- The frame counts / timestamps are saved in a separate "framestamps.json" file aside the (current day's) frames directory


** CameraControl.py and CameraModeSwitcher.py **

- Added functions to CameraControl to switch to night and daytime modes.
- CameraModeSwitcher calls these functions when the sun is 10 degrees below horizon (at sunset, sunrise), and waits until next switch. This function actually facilitates the switch by modifying a shared variable (daytime_mode) at the right time.


** DeleteObservations.py ** 

- Added functions to clean up and manage special directory structure of FramesFiles and VideoFiles (explained below)
- Last 8 days of FrameFiles are kept when cleaning up
- Last 2 days of VideoFiles are kept when cleaning up


** StartCapture.py **

- The wait function has been modified to conditionally break when camera mode switches at sunset/sunrise
- Control flow of runCapture has been modified to allow both standard and continuous capture modes based on the config - might need cleaning up
- Continuous capture can only be exited using Ctrl + C for now


The new directory structure for VideoFiles and FramesFiles is as follows (Thanks @Cybis320):

~/RMS_data/
└── XX0001/
    ├── ArchivedFiles/
    ├── CapturedFiles/
    ├── VideoFiles/                 # similar structure as FramesFiles, but for mkv video clips 
    └── FramesFiles/
        ├── 2024/
        │   ├── 20240101-01/
        │   │   ├── 00/
        │   │   │   ├── XX0001_20240101_000001_122577.jpg
        │   │   │   ├── XX0001_20240101_000011_122585.jpg
        │   │   │   └── ...
        │   │   ├── 01/
        │   │   │   ├── XX0001_20240101_010001_122577.jpg
        │   │   │   ├── XX0001_20240101_010011_122585.jpg
        │   │   │   └── ...
        │   │   ├── ...
        │   │   └── 23/
        │   │       ├── XX0001_20240101_230001_122577.jpg
        │   │       ├── XX0001_20240101_230011_122585.jpg
        │   │       └── ...
        │   ├── ...
        │   ├── 20241230-365/
        │   └── 20241231-366/
        ├── 2025/
        │   ...
        └── ...


** Reprocess.py and GenerateTimelapse.py **

- Frame timelapses, archived frames are created and generated only for past days, as capture might be running for the current day.
- These exist at the "year" directory level. After creation, the corresponding day directory (ex. 20241230-365) in FramesFiles is deleted.

RMS/BufferedCapture.py
```diff
@@ -519,6 +587,34 @@ def handleGrayscaleConversion(self, map_info):
         return gray_frame
 
 
+    def move_segment(self, splitmuxsink, fragment_id):
```
dvida: All function names should be camelCase, fix throughout.

.config
```diff
@@ -226,6 +228,11 @@ bz2_files_to_keep: 20
 ; number of CapturedFiles folders to keep. Default is 8
 capt_dirs_to_keep: 8
 
+; number of frame folders (days) to keep. Default is 8
+frame_dirs_to_keep: 8
+
+; number of raw video folders (days) to keep. Default is 2
+video_dirs_to_keep = 2
```
Cybis320: Should be ': ' instead of ' = '

RMS/BufferedCapture.py
```diff
@@ -1081,6 +1187,47 @@ def run(self):
...
+
+                    # Write raw frame and timestamp to one of the two corresponding buffers
+                    if raw_buffer_one:
+                        self.sharedRawArray[self.raw_frame_count, :, :] = frame
+                        self.sharedTimestamps[self.raw_frame_count] = frame_timestamp
+
+                    else:
+                        self.sharedRawArray2[self.raw_frame_count, :, :] = frame
+                        self.sharedTimestamps2[self.raw_frame_count] = frame_timestamp
+
```
Cybis320: I'm getting this error when attempting to capture during the day:
"ValueError: could not broadcast input array from shape (1080,1920,3) into shape (1080,1920)"
This error occurs because we're trying to write a 3-channel color frame (shape 1080x1920x3) into a 2D array buffer (shape 1080x1920).

RMS/BufferedCapture.py
```diff
@@ -110,6 +115,43 @@ def __init__(self, array1, startTime1, array2, startTime2, config, video_file=No
...
+            # Initialize shared memory arrays for raw frame saving
+            self.sharedRawArrayBase = Array(ctypes.c_uint8, self.num_raw_frames * config.height * config.width)
+            self.sharedRawArray = np.ctypeslib.as_array(self.sharedRawArrayBase.get_obj())
+            self.sharedRawArray = self.sharedRawArray.reshape(self.num_raw_frames, config.height, config.width)
+
+            self.sharedRawArrayBase2 = Array(ctypes.c_uint8, self.num_raw_frames * config.height * config.width)
+            self.sharedRawArray2 = np.ctypeslib.as_array(self.sharedRawArrayBase2.get_obj())
+            self.sharedRawArray2 = self.sharedRawArray2.reshape(self.num_raw_frames, config.height, config.width)
+
```
Cybis320: Should we initialize the arrays after the frame shape is known? And handle the switch between greyscale and 3 channels?

Cybis320: maybe first initialize the arrays as None:
```python
self.sharedRawArrayBase = None
self.sharedRawArray = None
self.sharedRawArrayBase2 = None
self.sharedRawArray2 = None
self.rawFrameSaver = None
```
 Then initialize with a method when the shape is known:
 ```python
def initRawFrameArrays(self, frameShape):
        """Initialize the raw frame arrays once we know the frame shape"""
        if not self.config.save_frames:
            return

        # Calculate buffer size based on frame shape
        if len(frame_shape) == 3:
            # Color frame (height, width, channels)
            bufferSize = self.num_raw_frames * frameShape[0] * frameShape[1] * frameshape[2]
            arrayShape = (self.num_raw_frames, frameShape[0], frameShape[1], frameShape[2])
        else:
            # Grayscale frame (height, width)
            bufferSize = self.num_raw_frames * frameShape[0] * frameShape[1]
            arrayShape = (self.num_raw_frames, frameShape[0], frameShape[1])

        # Initialize shared memory arrays for raw frame saving
        self.sharedRawArrayBase = Array(ctypes.c_uint8, buffer_size)
        self.sharedRawArray = np.ctypeslib.as_array(self.sharedRawArrayBase.get_obj())
        self.sharedRawArray = self.sharedRawArray.reshape(arrayShape)

        self.sharedRawArrayBase2 = Array(ctypes.c_uint8, buffer_size)
        self.sharedRawArray2 = np.ctypeslib.as_array(self.sharedRawArrayBase2.get_obj())
        self.sharedRawArray2 = self.sharedRawArray2.reshape(arrayShape)

        # Initialize the raw frame saver
        self.rawFrameSaver = RawFrameSaver(self.saved_frames_dir, 
                                           self.sharedRawArray, self.startRawTime1, 
                                           self.sharedRawArray2, self.startRawTime2, 
                                           self.sharedTimestamps, self.sharedTimestamps2,
                                           self.config)
```

**Comment by Cybis320** [comment]
Testing with the initRawFrameArrays method above. It's capturing in day mode right now :) Thank you @MaadhyamRana !

File: RMS/RawFrameSave.py [hunk 5]
```diff
@@ -0,0 +1,290 @@
...
+                block_start_time = timestamp
+                self.total_saved_frames = 0
+
+
+            # Generate names for the file and path
+            date_string = time.strftime("%Y%m%d_%H%M%S", time.gmtime(timestamp))
+            timed_dir_string = time.strftime("%Y/%Y%m%d-%j/%Y%m%d-%j_%H", time.gmtime(timestamp))
+
+            # Calculate milliseconds
+            millis = int((block_start_time - floor(block_start_time))*1000)
```
Cybis320: Why use the block start time for the millis, why not the frame timestamp? Currently, each block of 10 frames has the same milliseconds from the first frame in the block.

MaadhyamRana:
The code has been modified so that it checks if the raw frame shape has changed at the frame level (instead of at the block level), thanks for the help @Cybis320 !

Cybis320:
Thank you @MaadhyamRana ! Testing right now on my station... sunset in 1 hr.

Cybis320:
Unfortunately, it did not like the camera switch at sunset. I'm not sure if it's because all six cameras are being sent the same commands at the same moment in a rapid fire manner, but all six camera just hung for 30 minutes at the changeover, and only show signs of having changed some of their settings.
I made some changes to the CameraControl and  CameraModeSwitcher and re-started capture on all six. Will see what happens at sunrise.

RMS/BufferedCapture.py
```diff
@@ -905,41 +958,169 @@ def releaseResources(self):
...
+                                                    self.sharedTimestamps, self.sharedTimestamps2,
+                                                    self.config)
+                    self.raw_frame_saver.start()
+                    log.debug("Raw frame handling initialization complete")
+                    
+                except Exception as e:
+                    log.error("Failed to initialize raw frame handling: {}".format(e))
+                    log.debug(repr(traceback.format_exception(*sys.exc_info())))
+                    self.raw_frame_saver = None
+                    raise
```
Cybis320: I think we can safely delete L1087-L1107 since we are initializing the arrays on frame shape change later. Also the frame_shape at this point is the code is not accurate yet as it's not gone through the greyscale conversion if applicable.. 
Later at L1264,  `if frame.shape != self.current_raw_frame_shape:` is the test to (re)initialize the shared arrays. At that point frame.shape should be correct (as it's been read() ) and current_raw_frame_shape is None before the first initialization.

MaadhyamRana:
Video and Frame directories (days) are deleted first in case of less disk space, so the relatively smaller Capture data is preserved

RMS/Reprocess.py
```diff
@@ -465,6 +466,77 @@ def processNight(night_data_dir, config, detection_results=None, nodetect=False)
...
+                    # Search for current day's timelapse in the corresponding year's directory
+                    found_files = glob.glob(os.path.join(year_dir, f'{day}_frames_timelapse.mp4'))
+
+                    # If not found, generate timelapse for the current day
+                    if not found_files:
+                        log.info(f"No frames timelapse for {day} found in {year_dir}, generating new timelapse...")
+
+                        # Make the name of the timelapse file from day directory
+                        # The day's timelapse is stored in its corresponding year's directory
+                        frames_timelapse_path = os.path.join(year_dir, day + "_frames_timelapse.mp4")
```
Cybis320: Should we add the station ID, e.g. XX0001-20241102-307, to the mp4, bz2, and json file names? Otherwise we end up with the same name for all stations.

Cybis320:
@MaadhyamRana, I'm seeing cases where camera switching commands are not being received by the cameras. This seems to happen during the transition from night to day mode. One possibility is that the machine is overwhelmed at that exact moment. I'm testing to see if postprocess_delay will help. I'm also experimenting with setting a low niceness value for BufferedCapture.
In addition, we need a way to resend commands on reconnects. Currently, there's no capturing during the first 25 minutes of post-processing. When the cameras resume capture, they are supposed to be in day mode, but because they missed the day commands, they remain in night mode. Therefore, we need to implement a mechanism to resend commands when cameras reconnect.

MaadhyamRana:
When there is no capturing for those first 25 minutes, is BufferedCapture looping on L1168?
If so, we can import CameraControl and use the present daytime_mode.value to resend the appropriate commands (SwitchDayTime or SwitchNightTime) from within that loop.

Cybis320:
it tries pinging the camera for 25 minutes. I don't understand why exactly. The overwhelmed RTSP client might have an effect on the RTSP server upstream somehow? That would be strange especially since it broadcast in UDP - the server shouldn't care about what's happening downstream. Or maybe, the pinging utility itself is stuck? I try probing the rstp server itself, bypassing ping, it's a better approach overall, but same results in this case. Maybe the network interface itself is affected - I'll add some debug code to find out. I think we really need something at the stage just before we (re)establish a rtsp stream, just after:
```python
            # Init the video device
            log.info("Initializing the video device...")
            log.info("Device: " + str(self.config.deviceID))
```

Cybis320:
Another clue is that all 6 cameras drop offline and come back online at the same instant... so it's probably a network interface thing. They all drop at the exact moment of the switch to daytime capture and all come back together 25 minutes later.

Cybis320:
I'll experiment staggering the mode switches with `config.postprocess_delay`. Maybe, all the TCP commands at once are overwhelming?

Cybis320:
Update: staggering the mode switching with `config.postprocess_delay` added in CameraModeSwitcher solved the issue. All cameras running on all cylinders!
I did look at the system logs for the event prior to this experiment, the network interface did drop for 24 minutes at the night to day mode transition. It could a network buffer size issue or a cpu irq issue. In any case, staggering  and backing it up with a mode check on reconnects is the way to go.

RMS/Reprocess.py
dvida: Agreed, @MaadhyamRana could you implement this?

PR #429, opened 2024-09-25
Add FOV check for platepar

dvida:
This pull request adds a FOV check for the platepar in the `Reprocess.py` file. The check ensures that the FOV in the platepar is within a factor of two of the FOV in the config file. If the FOV ratio is smaller than 0.5 or greater than 2, the platepar is not used. This check helps to ensure that the platepar and config file are compatible before using the platepar for further processing.

Example of when this is needed: https://globalmeteornetwork.org/weblog/US/US003F/US003F_20240924_021434_064110_detected/

### DIFF
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -106,6 +106,26 @@ def getPlatepar(config, night_data_dir):
                 platepar.lon = config.longitude
                 platepar.elev = config.elevation
 
+        
+        # Make sure the config and the platepar FOV are within a factor of two
+        if (platepar.fov_h is not None) and (platepar.fov_v is not None):
+            
+            # Calculate the diagonal FOV for both the platepar and the config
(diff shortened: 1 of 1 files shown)


Cybis320:
I wonder if could be caused by these lines:
https://github.com/CroatianMeteorNetwork/RMS/blob/cdec45adf64585586a66399b9336161c28956af8/RMS/Astrometry/CyFunctions.pyx#L1151-L1153

https://github.com/CroatianMeteorNetwork/RMS/blob/cdec45adf64585586a66399b9336161c28956af8/RMS/Astrometry/CyFunctions.pyx#L1509-L1511

I got rid of % in my code without ill-effect and I don't seem to get these runaway issues.

Cybis320:
If we have a set of files that reliably goes rogue, we could try commenting out the above snippets to see what happens. Can we reproduce the problem?

Issue #434, opened 2024-09-29
Bug in startObservationSummaryReport  - incorrectly assuming source code location


markmac99:
Richard Bassom in the UK reports a failure on the multicam installation that seems to be due to the new functionality to create a report of station status.

At line 169, the code in startObservationSummaryReport tries to import a git repo to check the RMS version. It assumes that the the folder containing the RMS config file also contains the code. On multistation builds this is incorrect and so it fails.

The code in question seems to have been introduced into RMS about 28 days ago, and may not have been noticed by multicam users as many do not reboot frequently. 
I think we need exception handling around that section of the code and/or passing in the right folder. I'll raise an issue and perhaps someone familiar with this code could take a look? 

On 29/09/2024 09:20, Richard Bassom via groups.io wrote:

I had a problem at the start of last night with my Ubuntu Linux PC RMS system running 3 cameras. I reverted to the previous RMS release to get it running again.
The RMS log shows the following, and the capture didn't start:

2024/09/28 18:23:14-INFO-Logger-line:44 - Traceback (most recent call last):
2024/09/28 18:23:14-INFO-Logger-line:44 - File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
2024/09/28 18:23:14-INFO-Logger-line:44 - return _run_code(code, main_globals, None,
2024/09/28 18:23:14-INFO-Logger-line:44 - File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
2024/09/28 18:23:14-INFO-Logger-line:44 - exec(code, run_globals)
2024/09/28 18:23:14-INFO-Logger-line:44 - File "/home/pi/vRMS/lib/python3.8/site-packages/RMS-0.1-py3.8-linux-x86_64.egg/RMS/StartCapture.py", line 1181, in <module>
2024/09/28 18:23:14-INFO-Logger-line:44 - night_archive_dir = runCapture(config, duration=duration, nodetect=cml_args.nodetect, \
2024/09/28 18:23:14-INFO-Logger-line:44 - File "/home/pi/vRMS/lib/python3.8/site-packages/RMS-0.1-py3.8-linux-x86_64.egg/RMS/StartCapture.py", line 387, in runCapture
2024/09/28 18:23:14-INFO-Logger-line:44 - log.info(startObservationSummaryReport(config, duration, force_delete=False))
[...]

PR #435, opened 2024-09-29
Comment out lines related to git repo as a triage measure for multicam

g7gpr:
Issue #434 reports that git repo cannot be ascertained in multicam environment and prevents capture start. This PR comments out the lines of code which are causing the crash.

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -166,13 +166,13 @@ def startObservationSummaryReport(config, duration, force_delete=False):
 
     addObsParam(conn, "hardware_version", hardware_version)
 
-    repo = git.Repo(search_parent_directories=True)
-    if repo:
-        addObsParam(conn, "commit_date",
-                    datetime.datetime.fromtimestamp(repo.head.object.committed_date).strftime('%Y%m%d_%H%M%S'))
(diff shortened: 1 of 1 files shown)


g7gpr:
OK - the problem is slightly more subtle. The old version of multicam launches RMS in ~/source/Stations/STAID; which leads to all kinds of strange things. New version will launch from ~/source/RMS and then point to the correct .config file.

g7gpr:
Testing this change now, launching in ~/source/Stations/AU000A

g7gpr:
I've tested this launching in both ~/source/Stations/STATID, and also from ~/source/RMS and capture is running. I'll let the capture finish, check the results and then this is ready to merge.

g7gpr:
2 Captures completed, one launched from ~/source/RMS. and one launched with the working directory ~/source/Stations/STATID.

Cybis320:
It's just that the old startCap files are not automatically updated and need to be manually updated, no?

g7gpr:
They are not automatically updated. The files were originally distributed by emailing zip files around. Because there were different implementations, I don't think we should update anything remotely, just ask station operators to manually update their systems once everything is stable.

g7gpr:
Closing as handled by commit cdec45adf64585586a66399b9336161c28956af8

PR #441, opened 2024-09-30
Add try-except block to Observation summary

Cybis320:
Add try-except block to Observation Summary to troubleshoot reports of reprocess hanging.

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -226,10 +226,10 @@ def timestampFromNTP(addr='0.us.pool.ntp.org'):
         client.sendto(data, (addr, 123))
         data, address = client.recvfrom(1024)
     except socket.timeout:
-        logging.error("NTP request timed out")
+        print("NTP request timed out")
         return None
     except Exception as e:
-        logging.error("NTP request failed: %s", e)
+        print("NTP request failed: {}".format(e))
         return None
     if data:
         t = struct.unpack('!12I', data)[10]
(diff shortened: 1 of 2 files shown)


g7gpr:
Based on comments from op I suspect the call to get the time is hanging.

Cybis320:
This?
https://github.com/CroatianMeteorNetwork/RMS/blob/cdec45adf64585586a66399b9336161c28956af8/RMS/Formats/ObservationSummary.py#L501

g7gpr:
https://github.com/CroatianMeteorNetwork/RMS/blob/dcb771e8178bc5310aba61a1eb7d3f8033dbde4c/RMS/Formats/ObservationSummary.py#L234

Here.

g7gpr:
https://github.com/CroatianMeteorNetwork/RMS/blob/dcb771e8178bc5310aba61a1eb7d3f8033dbde4c/RMS/Formats/ObservationSummary.py#L248

This line specifically.

g7gpr:
I wonder if UDP is blocked by station's firewall, and this might cause recvfrom to hang

Cybis320:
I could be. Take a look at the last commit. This should make fetching NTP time safer.

g7gpr:
That's exactly what I had in mind. From here I can't test this on a UDP blocked network, but you should be able to check that function just by running the module.

Cybis320:
Could we just check our suspicion but just running the code vefore and after the PR with no internet connection?

g7gpr:
Tried that on a UDP blocked network. Old version of code threw an exception; which was what I had expected to happen. But is the firewall doing something which is causing it to hang rather than thrown an exception. I'll pull in this version and try again.

g7gpr:
Tested on a UDP blocked network. I couldn't get the timeout clause to activate, it jumped straight to NTP request failed; but I don't have any better ideas than this.

g7gpr:
can you change the wording from "NTP request timed out" to "Time sync check failed". I can think of situations where Chrony / NTP etc would be working fine, but this code might fail.

g7gpr (approved):

Issue #443, opened 2024-10-10
FRbinViewer shortens video


Vasiliy1992:
I found that when I save a video with a fireball, for some reason it gets shortened. Half of the video is not saved!
Although everything is fine in the preview.
If the video is split into several files when saving, is it possible to save them all? It is not difficult to glue them together in a video editor.
[I asked this question on the forum](https://globalmeteornetwork.groups.io/g/main/topic/frbinviewer_shortens_video/108836949). I was advised to duplicate it here.

PR #444, opened 2024-10-10
FRbinViewer - single video for multi line bin file fix

aitov:
Hi,  several days ago we have question on groups about short video for fireball ( [FRbinViewer shortens video](https://globalmeteornetwork.groups.io/g/main/topic/108836949) )

Last night I also got same bin file with 2 meteors for one  fits file and see that for this case we generate only part of generated frames because for ffmpeg  need correct sequence of image files but we have gaps in case multi lines.

It works with option : '-s', '--split' but it generates several video files

Proposed fix : use sequence  numbering instead real frame number in bin file.
Also I improved timestamp option ("-t", "--timestamp") - it shows real time for meteors on video.

Example of result attached.  Source bin file for testing could be get from groups link above or I could provide my if need (not sure if I can attach 40mb here)

https://github.com/user-attachments/assets/4fde2280-88d5-4f10-af71-84f20cf580fe

### DIFF
--- a/Utils/FRbinViewer.py
+++ b/Utils/FRbinViewer.py
@@ -91,10 +91,8 @@ def view(dir_path, ff_path, fr_path, config, save_frames=False, extract_format=N
             background = ff_file.maxpixel
         if append_ff_to_video:
             meteor_image = np.copy(ff_file.maxpixel)
-        timestampTitle = ""
         if add_timestamp:
-            timestampTitle = getTimestampTitle(ff_path)
-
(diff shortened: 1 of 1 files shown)


aitov:
Related issue #443

aitov:
Hi @dvida , I know it is not critical issue but I'm using it locally over 10 months from my branch, is there any chance have it in master? The multi videos are rare but feature with real time very helpfull (please see time in my example) - I can create separate PR for it if need.

PR #450, opened 2024-11-06
Increase UDP Buffer Size Configuration Script

Cybis320:
## Problem
GStreamer is reporting UDP buffer allocation warnings in the logs when `protocol: udp` is selected in `.config` :
```
GStreamer: udpsrc: warning: Could not create a buffer of requested 524288 bytes
```
This occurs because GStreamer's UDP source is requesting 512KB buffers while system defaults are often lower. Even with the warning, capture usually proceeds normally, however the smaller buffer may cause issues under high work load. 

## Solution
Added a script `update_buffers.sh` that:
- Checks current UDP buffer sizes
- Warns if buffers are below recommended sizes (512KB minimum, 1MB recommended)
- Offers to increase buffer sizes if needed
- Creates a backup of original settings
- Shows before/after comparison

## Usage
```bash
sudo ./update_buffers.sh
```

## Why
- Each camera's UDP source requests a 512KB buffer
- Default Linux buffer sizes are often too small (208KB)
- Increasing buffer size provides headroom for UDP streaming
- Helps prevent potential packet loss during traffic bursts

## Safety
- Script creates automatic backups before changes
- Prompts for confirmation before making changes
- Settings can be reverted using backup file
- Changes persist across reboots

### DIFF
--- a/RMS/Astrometry/ApplyAstrometry.py
+++ b/RMS/Astrometry/ApplyAstrometry.py
@@ -837,12 +837,9 @@ def XyHt2Geo(platepar, x, y, h):
     
     az, elev = cyTrueRaDec2ApparentAltAz(np.radians(ra_arr), np.radians(dec_arr), jd_arr, \
         np.radians(platepar.lat), np.radians(platepar.lon), platepar.refraction)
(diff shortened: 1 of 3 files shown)


Cybis320:
Malformed PR. Closing.

PR #461, opened 2024-11-11
continue framestamps.json for current day after reboot

MaadhyamRana:
The framestamps.json file would start counting from 0 after a reboot. This fixes it by initializing the starting raw frame value from an existing framestamp file for the current day, if it exists.

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -684,7 +684,7 @@ class BufferedCapture(Process):
 
 
 
-    def moveSegment(self, splitmuxsink, fragment_id):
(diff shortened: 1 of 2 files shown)


RMS/BufferedCapture.py
```diff
@@ -684,7 +684,7 @@ def handleGrayscaleConversion(self, map_info):
 
 
 
-    def moveSegment(self, splitmuxsink, fragment_id):
+    def move_segment(self, splitmuxsink, fragment_id):
```
dvida: Rename to moveSegment.
Function names in RMS are camelCase

PR #464, opened 2024-11-17
Enable network interface selection, cleaning dummy code

satmonkey:


dvida:
Thanks - the removal of the "from __future__ import print_function, unicode_literals, division, absolute_import" line will make it incompatible with py2, what this the original intent?

satmonkey:
ok, I reverted this, python2 will throw an exception now

PR #469, opened 2024-11-19
Prevent log propagation globally

Cybis320:
Prevented log propagation by setting log.propagate = False in the logger initialization and remove  log.propagate = False from BufferedCapture.py.


Cybis320:
Closing as made obsolete by recent changes to logger.

PR #483, opened 2024-12-19
Stagger camera mode switching

Cybis320:
Simultaneous camera mode switching on MCL can be unstable. This PR uses the config `capture_wait_seconds` to stagger the camera switching. On initial run, the mode switching is not delayed as capture_wait_seconds already delays start capture.


dvida:
I'd be happy to merge it, could you please resolve conflicts?

PR #490, opened 2024-12-25
Improved raw items (videos and frames) deletion

MaadhyamRana:
This PR is for clarity within the DeleteOldObservations code:

- changes the naming convention of frame_dirs_to_keep to frame_days_to_keep

- changes the naming convention of functions getRawDirs / deleteRawDirs to getRawItems / deleteRawItems
 
- improved language in comments and docstrings of the above functions

- in FramesFiles, each day consists of three processed files (frames archive, timelapse and frametime json). The code now automatically handles the frame_days_to_keep config option against this detail gracefully

### DIFF
--- a/.config
+++ b/.config
@@ -267,11 +267,11 @@ bz2_files_to_keep: 20
 ; Number of CapturedFiles folders to keep. Default is 8
 capt_dirs_to_keep: 8
 
-; Number of frame folders (days) to keep. Default is 8
-frame_dirs_to_keep: 4
+; Number of frame file days to keep. Default is 4
+frame_days_to_keep: 4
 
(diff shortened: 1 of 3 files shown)


Cybis320:
Thank you. What would happen if frame cleanup was set to 'delete' (no frame tar file generated). There would then only be two files to deal with. 

The use case for this is on stations with very small storage, we might want to immediately delete the jpg files after the mp4 creation (by making the cleanup arg a config option).

MaadhyamRana:
Should be no problem, this was tested on US000F. It's designed to look for
all three set of files but just continues if some are missing, while
deleting whatever is present

Cybis320:
Perfect. Thank you!

PR #496, opened 2025-01-08
Accurate timestamping for RMS (FT files)

MaadhyamRana:
Since the last update that changed the way raw video segments are named and stored, here are the changes this PR makes towards accurate timestamping of these segments:

1. An empty array is initialized to store the pts at the initialization of BufferedCapture.

2. ~~Every call to read() will append a new pts to the array.~~

3. Gstreamer callback moveSegment is called when a clip is ready. The segment is stored in the hierarchy under VideoFiles, while the pts array is flushed to a corresponding FT file under config.data_dir/TimeFiles (configureable), which follows a similar directory structure as the videos (Year/Day-of-Year/Hour/)

4. The timestamp array in memory is immediately cleared to store the next segment's pts.

In addition, each FT file and corresponding video segment are now named after ~~their first timestamp, and not using current time (RmsDateTime.utcnow).~~

### DIFF
--- a/.config
+++ b/.config
@@ -239,6 +239,9 @@ frame_dir: FramesFiles
 ; Name of directory within data_dir for raw video files
 video_dir: VideoFiles
 
+; Name of directory within data_dir for raw video timestamp (FT) files
+times_dir: TimeFiles
+
 ; Name of directory within data_dir for log files
 log_dir: logs
 
@@ -270,7 +273,7 @@ capt_dirs_to_keep: 8
 ; Number of frame file days to keep. Default is 4
 frame_days_to_keep: 4
 
-; Number of raw video folders (days) to keep. Default is 2
+; Number of raw video + corresponding FT file folders (days) to keep. Default is 2
 video_days_to_keep: 2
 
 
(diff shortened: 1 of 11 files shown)


MaadhyamRana:
Since the last update:

- FT files are now produced by default under TimeFiles subdirectories, not just when `config.raw_video_save` is set on.
- `config.times_days_to_keep` is set to 8 by default, indicating to keep 8 days worth of TimeFiles data when set.
- Instead of appending timestamps to buffer during every call to read, new timestamps are only appended for successful `read()` calls, and not when the camera is trying to reconnect.
- FT file is saved from buffer at the end of every 256-frame block captured.
- New video segments are now named when the previous segment is ready to be saved. This is tracked by `BufferedCapture`'s `self.last_segment_savetime`

MaadhyamRana:
Since the last update:

- Saving FT files is now a configurable option (config.save_frame_times), and is set to True by default.

- During post-processing after a night, previous day's FT files directory is archived (.tar.bz2), and original directory is removed. The archive is then added onto the upload queue.

- Further clarified documentation in DeleteOldObservations.py, and updated functions to handle keeping / deleting the above archives.

PR #500, opened 2025-01-12
Logging System Overhaul

Cybis320:
# Logging System Overhaul

This PR introduces a redesign of our logging system to address issues with multiprocessing support and log file rotation. The new implementation uses a dedicated listener process with a message queue architecture for reliable logging and implements an improved log rotation strategy.

### 1. Multiprocessing Support
- Implemented a dedicated logging listener process using `multiprocessing.Process`
- Added a message queue (`multiprocessing.Queue`) for thread-safe logging across processes
- Introduced `QueueHandler` and `QueueListener` for asynchronous logging operations

### 2. Enhanced Log Rotation System
The previous implementation had issues with log rotation due to concurrency issues in multi-process environments. The new system introduces a custom rotation handler with an improved naming scheme for better readability. It also preserves the .log file extension on the rotated log files.

#### New Naming Convention
- Live log file: `log_[STATION_ID]_[START_TIME].log`
  - Example: `log_US005A_2024-12-29_112347.log`
- Rotated log file:
  - Pattern: `log_[STATION_ID]_[INITIAL_START_TIME]-[LOG_START_TO_END].log`
  - Format: `log_XX####_YYYY-MM-DD_HHMMSS-[DD_HHMM-to-DD_HHMM].log`
  - Examples:
    ```
    log_US005A_2024-12-29_112347.log
    log_US005A_2024-12-29_112347-[30_1123-to-01_1123].log
    log_US005A_2024-12-29_112347-[29_1123-to-30_1123].log
    ```

*Note: The current naming convention assumes processes run less than one month. For longer-running processes, we could extend the format to include month information if needed.*

### DIFF
--- a/RMS/ArchiveDetections.py
+++ b/RMS/ArchiveDetections.py
@@ -3,11 +3,11 @@
 
 import os
 import sys
-import logging
 import traceback
 
 
 from RMS.Formats.FFfile import validFFName
+from RMS.Logger import getLogger
 from RMS.Misc import archiveDir
 from RMS.Routines import MaskImage
 from Utils.GenerateThumbnails import generateThumbnails
@@ -15,7 +15,7 @@ from Utils.StackFFs import stackFFs
 
 
 # Get the logger from the main module
-log = logging.getLogger("logger")
+log = getLogger("logger")
 
 
 def selectFiles(config, dir_path, ff_detected):
--- a/RMS/Astrometry/ApplyRecalibrate.py
+++ b/RMS/Astrometry/ApplyRecalibrate.py
@@ -11,7 +11,6 @@ import json
 import os
 import shutil
 import sys
-import logging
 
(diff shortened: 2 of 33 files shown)


Cybis320:
An issue with the 24hr rollover scheme on stations that reboot after processing is that the elapsed time between start of logging and reboot can be slightly over 24 hours. This results in two log files, the first 24 hr and a second very short log. 
I'm not sure if we should set a different rollover frequency (25 hr?) or rollover on a signal processing is complete or a max time is reached, whichever comes first?

dvida:
Brilliant, amazing work! I'm actually OK with the 24-hour rollover. It's consistent and easy to understand, even if it produces multiple log files.
If the rollover can be adjusted on the fly, then the alternative would be to set the rollover just before the next capture time.

PR #502, opened 2025-01-15
Grayscale fix with GStreamer

dario-zubovic:
GStreamer path in *BufferedCapture.py* included a faulty method that was using blue frame for grayscale, even though result of `read()` can return all three color frames and grayscale conversion will be handled correctly later in `captureFrames()` method.

**Not tested**, please test on RPi with hardware decoding before merging!

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1338,6 +1338,29 @@ class BufferedCapture(Process):
         finally:
             self.releaseResources()
 
+    def convertToGrayscale(frame):
+        """Converts provided frame to one channel if it's not already"""
+
+        # Convert the frame to grayscale
+        if len(frame.shape) == 3:
+
+            # If a color image is given, take the green channel
+            if frame.shape[2] == 3:
+
+                gray = frame[:, :, 1]
+
+            # If UYVY image given, take luma (Y) channel
+            elif self.config.uyvy_pixelformat and (frame.shape[2] == 2):
+                gray = frame[:, :, 1]
+
+            # Otherwise, take the first available channel
+            else:
+                gray = frame[:, :, 0]
+
+        else: # already converted
+            gray = frame
+
+        return gray
 
 
     def captureFrames(self):
@@ -1494,6 +1517,10 @@ class BufferedCapture(Process):
                     and self.video_file is None
                     and total_frames % self.config.frame_save_interval_count == 0):
 
+                    # ensure only one channel when in nightmode
+                    if (self.daytime_mode is None) or (not self.daytime_mode.value):
+                        frame = convertToGrayscale(frame)
+
                     # In case of a mode switch, the frame shape might change (color or grayscale)
                     if frame.shape != self.current_raw_frame_shape:
                         log.info("Frame shape changed, reinitializing arrays...")
@@ -1623,33 +1650,13 @@ class BufferedCapture(Process):
                 last_frame_timestamp = frame_timestamp
                 
 
-                ### Convert the frame to grayscale ###  (Not to be done in case of daytime mode)
+                # Convert the frame, clip to ROI and pass to shared arrays for further processing
+                # unless it's daytime mode
                 if (self.daytime_mode is None) or (not self.daytime_mode.value):
 
                     t1_convert = time.time()
 
-                    # Convert the frame to grayscale
-                    #gray = cv2.cvtColor(frame, cv2.COLOR_BGR2GRAY)
-
-                    # Convert the frame to grayscale
-                    if len(frame.shape) == 3:
-
-                        # If a color image is given, take the green channel
-                        if frame.shape[2] == 3:
-
-                            gray = frame[:, :, 1]
-
-                        # If UYVY image given, take luma (Y) channel
-                        elif self.config.uyvy_pixelformat and (frame.shape[2] == 2):
-                            gray = frame[:, :, 1]
-
-                        # Otherwise, take the first available channel
-                        else:
-                            gray = frame[:, :, 0]
-
-                    else:
-                        gray = frame
-
+                    frame = convertToGrayscale(frame)
 
                     # Cut the frame to the region of interest (ROI)
                     gray = gray[self.config.roi_up:self.config.roi_down, \
@@ -1657,12 +1664,6 @@ class BufferedCapture(Process):
 
                     # Track time for frame conversion
                     t_convert = time.time() - t1_convert
-
-
-                    ### ###
-
-
-
 
                     # Assign the frame to shared memory (track time to do so)
   
@@ -1673,8 +1674,6 @@ class BufferedCapture(Process):
                         self.array2[i, :gray.shape[0], :gray.shape[1]] = gray
 
                     t_assignment = time.time() - t1_assign
-
-
 
                 # Keep track of all captured frames
                 total_frames += 1
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -495,8 +495,8 @@ class BufferedCapture(Process):
                     log.info("GStreamer Buffer did not contain a frame.")
                     return False, None, None
 
-                # Handling for grayscale conversion
-                frame = self.handleGrayscaleConversion(map_info)
+                # Cast to uint8 if necessary
+                frame = np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
 
                 # Smooth raw pts and calculate actual timestamp
                 smoothed_pts = self.smoothPTS(gst_timestamp_ns)
@@ -665,23 +665,6 @@ class BufferedCapture(Process):
             return True
         
         return False
-
-    
-    def handleGrayscaleConversion(self, map_info):
-        """Handle conversion of frame to grayscale if necessary."""
-
-        # If the frame is already grayscale, return it as is
-        if len(self.frame_shape) == 2:
-            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
-
-        if (self.daytime_mode is not None) and (self.daytime_mode.value):
-            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
-
-        # Convert to grayscale by selecting a specific channel
-        bgr_frame = np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
-        gray_frame = bgr_frame[:, :, 0]  # Assuming the blue channel for grayscale
-        return gray_frame
-
 
 
     def moveSegment(self, splitmuxsink, fragment_id):
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1517,8 +1517,8 @@ class BufferedCapture(Process):
                     and self.video_file is None
                     and total_frames % self.config.frame_save_interval_count == 0):
 
-                    # ensure only one channel when in nightmode
-                    if (self.daytime_mode is None) or (not self.daytime_mode.value):
+                    # avoid storing duplicate data
+                    if self.convert_to_gray:
                         frame = convertToGrayscale(frame)
 
                     # In case of a mode switch, the frame shape might change (color or grayscale)


Cybis320:
I'll have to look at that. The idea is that the camera in night or day mode outputs three channels. in 'greyscale' night mode all three channels are the same, therefore picking the blue  or any other channel is fine. Was it causing any issues?

Cybis320:
I'll have to double-check the logic wasn't changed in recent updates. What was originally meant to happen, since RMS now handles both color and greyscale images, and because the cameras output three channels in either mode:

1. A check was added to compare the 3 channels.
2. If all channels are the same, it would pick one (doesn't matter which one) as early as possible in the pipeline and work with just one channel
3. if channels are different, only convert to greyscale before compression but after SaveFrames. 

I'll have to check that it's still happening only if all three channels are equal and not merely whether we are in night or day mode (it shouldn't happen if we are merely in night mode capture but the camera is still in color mode @MaadhyamRana).

dario-zubovic:
No need to apply such conversion before read() returns; day/night it's handled accordingly later on.

Issue is choice of a channel does matter. Most RGB bayer matrices on ccd/cmos sensors have twice as many green pixels than red or blue. It should be a preferred channel to maximize sensitivity, like it's already done in `captureFrames()` (https://github.com/CroatianMeteorNetwork/RMS/blob/5374765757774a5965a05380f71a352c73473dc5/RMS/BufferedCapture.py#L1637).

It would be appreciated if you're able to give it a test, don't have an environment set up currently.

Cybis320:
Hi Dario, I think we posted our replies at the exact same time :) Let me know if the above post answers your concern.

dario-zubovic:
My reply does address your points, reformatted it a bit now for clarity :)

Tldr:
- choice of channel is important in night mode
- no reason to do any conversion at that point in code path either way

dario-zubovic:
Oh, right, camera in nightmode will probably return 1 channel or 3 channels with same information, in which case choice of channel doesn't matter.

RMS itself should still prefer G, in case camera is outputting color information, since it;s used across different setups.

Cybis320:
The choice of channel at that point in the pipeline does not matter as all three channels are verified to be exactly identical bit for bit.
If somehow a color frame made it to the compressors, the conversion is done as it was done in the past (this is a backup method).
The log indicates the results of the bit-for-bit comparisons of all the channels:
```
2025/01/14 16:04:07-INFO-BufferedCapture-line: Video format: BGR, 1080P, color: False
```
This line indicates that the camera is in night mode. It is producing three channels (BGR), and those three are exactly the same (color: False).

You are right that it matters how the conversion to greyscale is done, but it's out of our hands and done internally by the camera.

Also, the code is currently using the green channel (but it doesn't matter):
In a BGR (Blue-Green-Red) color frame, index 1 corresponds to the Green channel. So the line `gray = frame[:, :, 1]` is correctly extracting the green channel from the BGR image.

To break it down:
- In BGR format, the channels are ordered as:
  - frame[:, :, 0] -> Blue channel
  - frame[:, :, 1] -> Green channel (what's being extracted here)
  - frame[:, :, 2] -> Red channel

But again, it doesn't matter.

RMS/BufferedCapture.py
```diff
@@ -666,23 +666,6 @@ def isGrayscale(self, frame):
...
-        # If the frame is already grayscale, return it as is
-        if len(self.frame_shape) == 2:
-            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
-
-        if (self.daytime_mode is not None) and (self.daytime_mode.value):
-            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
-
-        # Convert to grayscale by selecting a specific channel
-        bgr_frame = np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
-        gray_frame = bgr_frame[:, :, 0]  # Assuming the blue channel for grayscale
```
dario-zubovic, line 682: This is the problematic line that extracts blue (index 0). Code I quoted before is from further down the execution path where conversion is done correctly.

dario-zubovic:
You're correct that on most systems it doesn't matter much.

Even if all systems won't exhibit any issues here due to conversion done in camera, BufferedCapture in RMS shouldn't make such assumption, both for correctness and future compatibility sake. For example if we ever get sensor input directly through CSI, there probably won't be any on-camera nightmode to speak of.

Cybis320:
Yes, you are right, it does extract the first channel which is blue at that point. Also, I do see the logic was changed and it does extract the blue channel merely on capture mode change not on confirmed grayscale as was originally intended. If you don't mind I'd like to submit a separate PR reverting back to the original intend.
Thank you for noticing.

dario-zubovic:
IMO no much point in doing any checks in `read()`, regardless if channels are bit-to-bit compatible or not... OpenCV path might still return different outputs, and as such `read()` output will still need handling for return array shapes. `handleGrayscaleConversion()` is just redundant and leaves room for small slips like this one to creep in.

But if you prefer handling this differently, feel free. Will be happy to give it a review. :)

Cybis320:
The issue with that plan is that we want to save raw frames in color if available.

What I think needs to happen is to revert back to the plan:
1. check if channel are identical
2. if identical, pick the first one, if not keep all three
3. save raw frame (color or grey)
4. if three channels at this point, pick the green
5. compress

What do you think?

Cybis320:
here's the original function which was only picking the first frame (faster?) if identical:
```python
def handle_grayscale_conversion(self, map_info):
        """Handle conversion of frame to grayscale if necessary."""
        if not self.convert_to_gray:
            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)

        # Convert to grayscale by selecting a specific channel
        bgr_frame = np.ndarray(shape=self.frame_shape,
                               buffer=map_info.data, dtype=np.uint8)
        # Assuming the blue channel for grayscale
        gray_frame = bgr_frame[:, :, 0]
        return gray_frame
```

The current function doesn't appear to check first - it assumes capture night_mode = camera night mode = 3x identical ch, which is not necessarily true for all cameras:
```python
def handleGrayscaleConversion(self, map_info):
        """Handle conversion of frame to grayscale if necessary."""

        # If the frame is already grayscale, return it as is
        if len(self.frame_shape) == 2:
            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)

        if (self.daytime_mode is not None) and (self.daytime_mode.value):
            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)

        # Convert to grayscale by selecting a specific channel
        bgr_frame = np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)
        gray_frame = bgr_frame[:, :, 0]  # Assuming the blue channel for grayscale
        return gray_frame
```

Cybis320:
The OpenCV method (which is the fail-safe method, not the primary method), does indeed keep all three channels at that point and actually demonstrates the need to drop two channels. With OpenCV, the night saved frames are 3x times the size as it saves all three identical channels day or night. On the Gst path, the night saved frames are smaller. The OpenCV path was kept as simple and as close to the early RMS as possible.

> IMO no much point in doing any checks in `read()`, regardless if channels are bit-to-bit compatible or not... OpenCV path might still return different outputs, and as such `read()` output will still need handling for return array shapes. `handleGrayscaleConversion()` is just redundant and leaves room for small slips like this one to creep in.
> 
> But if you prefer handling this differently, feel free. Will be happy to give it a review. :)

dario-zubovic:
Ah, I see the issue, when raw frame saving is enabled and camera is nightmode, with this PR it will save "thick" 3x duplicate frames, instead of saving just a single channel, wasting disk space.

Regardless of this PR, that's exactly what's happening with OpenCV path, since it doesn't have any grayscale conversions in `read()`. Another bug? :)

I think the cleanest approach is:
- `read()` returns frames as they're provided by the driver (regardless if it's gstreamer or OpenCV)
- grayscale conversion for storing into shared buffers (for compression/video extraction) is handled in `captureFrames()`
- grayscale conversion for saving raw frames when in nightmode is handled in `captureFrames()`
- grayscale conversion is implemented in code just once
- grayscale conversion isn't done twice for perf reasons

Added a commit that addresses last three points. Does that work?

btw. I think there's no speed advantage of first channel versus others, it's just a fixed index offset.

RMS/BufferedCapture.py
```diff
@@ -1511,6 +1517,10 @@ def captureFrames(self):
                     and self.video_file is None
                     and total_frames % self.config.frame_save_interval_count == 0):
 
+                    # ensure only one channel when in nightmode
+                    if (self.daytime_mode is None) or (not self.daytime_mode.value):
```
dario-zubovic: First part of this expression is contestable.

When `daytime_mode` is not set (missing config value?), what's the preferred approach?
- store whatever driver has given us (change to: `if (self.daytime_mode is not None) and (not self.daytime_mode.value):`)
- always convert to grayscale (current: `if (self.daytime_mode is None) or (not self.daytime_mode.value):`)

Cybis320:
IMO, however you slice it, we need to compare the three channels before dumping two for save_frames. That's the only reliable way to be sure. Say people want to record auroras in color... the software would be in night mode, the camera would be in color.
Note the code does not convert twice currently. It converts once at two possible points in the pipeline. I think that's desirable.

Cybis320:
IMO, the only change needed is to revert back to:
```python
def handleGrayscaleConversion(self, map_info):
        """Handle conversion of frame to grayscale if necessary."""

        if not self.convert_to_gray:
            return np.ndarray(shape=self.frame_shape, buffer=map_info.data, dtype=np.uint8)

        # Convert to grayscale by selecting a specific channel
        bgr_frame = np.ndarray(shape=self.frame_shape,
                               buffer=map_info.data, dtype=np.uint8)
        # Assuming the blue channel for grayscale
        gray_frame = bgr_frame[:, :, 0]
        return gray_frame
       ```

If we are concerned about the cv2 path producing large jpgs, we could handle that too at some point (low priority).

**Comment by dario-zubovic** [comment]
Right, so the desired behaviour is to only grayscale before raw frame store if they're actually duplicate channels, since RMS might be in daymode and people are keeping cameras in color mode for other purposes.

Changed condition before raw frame store to match in the latest commit.

I see `self.convert_to_gray` is only set in `run()` (during video device initialization). Is buffered capture re-run when RMS is switching between daytime and nightime modes? Sorry, a bit out of the loop and would like to double check.

**Comment by Cybis320** [comment]
Ha. I just tested the theory that the grabbing [:, :, 0] (blue) is the fastest... it's actually the slowest!
 [:, :, 0] (red) is the fastest by ~7%!
```
Running channel access speed test...

Channel Access Speed Test Results (times in nanoseconds):
--------------------------------------------------------------------------------
 Channel   |     Mean     |    Median    |   Std Dev    |     Min      |     Max     
--------------------------------------------------------------------------------
   blue    |    181.69    |    167.00    |    47.24     |    83.00     |   1375.00   
  green    |    168.94    |    167.00    |    91.03     |    42.00     |   7833.00   
   red     |    168.60    |    167.00    |    106.68    |    42.00     |   13250.00  

Relative speed differences:
blue: +7.76% slower than fastest
green: +0.20% slower than fastest
red: +0.00% slower than fastest
```

**Comment by dario-zubovic** [comment]
Huh, interesting result. Maybe try running few dry conversions first, and doing more iterations? Min-max range and stddev seem large compared to mean.

**Comment by Cybis320** [comment]
Hm, there might be an issue there. It's checking for greyscale each time initVideoDevice is called. Which is on initial connection and on reconnect. Changing camera color mode tends to trigger a disconnect, but I'm not sure that's a reliable way to do it. Maybe there should be additional explicit triggers to re-check (like on mode changes).

> I see `self.convert_to_gray` is only set in `run()` (during video device initialization). Is buffered capture re-run when RMS is switching between daytime and nightime modes? Sorry, a bit out of the loop and would like to double check.
>

**Comment by dario-zubovic** [comment]
Ah, I suspected it might not be a good idea to rely on flag that's only calculated at initialization time. Will work out something tomorrow and report back.

**Comment by Cybis320** [comment]
Cool, thank you.
Also, I ran the test with different variations. Retrieving the blue channel is always slowest. Green and Red are about the same on average.:
```
Starting channel access speed test...
Creating array pool...
Warming up...
Running 10000000 iterations...

Raw Statistics (times in nanoseconds):
----------------------------------------------------------------------------------------------------
Channel  |    Mean    |   Median   |  Std Dev   |     Q1     |     Q3     |    Min     |    Max     | Samples 
----------------------------------------------------------------------------------------------------
  blue   |   192.06   |   167.00   |  2456.76   |   167.00   |   208.00   |   83.00    | 5365166.00 | 10000000
 green   |   185.45   |   167.00   |  7585.94   |   166.00   |   167.00   |   41.00    | 19925708.0 | 10000000
  red    |   181.91   |   167.00   |  1790.77   |   167.00   |   167.00   |   42.00    | 3583042.00 | 10000000

Filtered Statistics (outliers removed) (times in nanoseconds):
----------------------------------------------------------------------------------------------------
Channel  |    Mean    |   Median   |  Std Dev   |     Q1     |     Q3     |    Min     |    Max     | Samples 
----------------------------------------------------------------------------------------------------
  blue   |   184.20   |   167.00   |   22.14    |   167.00   |   208.00   |   125.00   |   250.00   | 9865574 
 green   |   166.65   |   167.00   |    0.48    |   166.00   |   167.00   |   166.00   |   167.00   | 7147588 
  red    |   167.00   |   167.00   |    0.00    |   167.00   |   167.00   |   167.00   |   167.00   | 5215004 

Relative speed differences (using filtered data):
blue: +10.53% slower than fastest
green: +0.00% slower than fastest
red: +0.21% slower than fastest
```

**Comment by Cybis320** [comment]
Here's another thought. The current way to check for channel identity (using cv2.split) is memory and CPU heavy, which is why we limit how often we run this check. However, after experimenting with the function, I found a much faster way to do it (220x faster):

```python
    def isGrayscale(self, frame, stride=64):
        """
        Return True if all color channels contain identical data.
        If main diagonal matches, assume grayscale (trading accuracy for speed)
        """
        if len(frame.shape) == 2:
                return True
        
        # Use slice stepping directly
        if not np.all(frame[::stride, ::stride, 0] == frame[::stride, ::stride, 1]) or \
        not np.all(frame[::stride, ::stride, 1] == frame[::stride, ::stride, 2]):
            return False
```

Performance improvements (tested on 1920x1080 frames):

- Speed: from ~1000μs to ~4.56μs (220x faster)
- Memory: from ~6.2MB (full frame copy) to a few KB (using only sampled pixels)

Still maintains 100% accuracy on my test cases

This is light enough that we could check for channel identity every time we save a raw frame (every 5 or 10 seconds) instead of restricting how often we check.

Cybis320:
I hope that's okay with you - I created draft PR #503. I did not incorporate the UYVY case because:

- UYVY conversion should happen earlier in the pipeline (cv2 handles it)
- By the time the frame reaches this function, it should already be in a standard image format
- The original UYVY handling, I think, was incorrect

I don't have a way to test it, but I assume only the cv2 pipeline could handle a UYVY stream and output a regular BGR frame. It should therefore not constitute a special case.

dario-zubovic:
Huh, timing results are still a bit peculiar, Doesn't seem like any of the channels is slower in median case, but B does seem to have by far largest variance in your timings, pulling the mean up. I'm unable to come up with believable explanation, but doesn't matter much. Hurray for green either way 😄 

I got a bit piled up with work... Will take a look at your proposals this weekend!

dvida:
Merged #503

Issue #494, opened 2025-01-03
EOFError in downloadNewMask


g7gpr:
Observed a crash caused by a network error in downloadNewMask. 

I suspect a try block would fix this problem, I'll investigate more. 

```
2025/01/01 00:37:22-INFO-DownloadMask-line:90 - files/masks exists
2025/01/01 00:37:22-INFO-DownloadMask-line:108 - Most recent flat /home/au001v/RMS_data/CapturedFiles/AU001V_20241231_115040_084674/flat.bmp
2025/01/01 00:37:22-INFO-DownloadMask-line:113 - Uploading to files/masks/ as AU001V_20241231_flat.bmp
2025/01/01 00:54:08-WARNING-DownloadMask-line:131 - Could not upload latest flat
2025/01/01 00:54:08-INFO-Logger-line:44 - Traceback (most recent call last):
2025/01/01 00:54:08-INFO-Logger-line:44 - File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
2025/01/01 00:54:08-INFO-Logger-line:44 - return _run_code(code, main_globals, None,
2025/01/01 00:54:08-INFO-Logger-line:44 - File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
2025/01/01 00:54:08-INFO-Logger-line:44 - exec(code, run_globals)
2025/01/01 00:54:08-INFO-Logger-line:44 - File "/home/au001v/source/RMS/RMS/StartCapture.py", line 941, in <module>
2025/01/01 00:54:08-INFO-Logger-line:44 - downloadNewMask(config)
2025/01/01 00:54:08-INFO-Logger-line:44 - File "/home/au001v/source/RMS/RMS/DownloadMask.py", line 140, in downloadNewMask
2025/01/01 00:54:08-INFO-Logger-line:44 - sftp.lstat(remote_mask)
2025/01/01 00:54:08-INFO-Logger-line:44 - File "/home/au001v/vRMS/lib/python3.9/site-packages/paramiko/sftp_client.py", line 511, in lstat
2025/01/01 00:54:08-INFO-Logger-line:44 - t, msg = self._request(CMD_LSTAT, path)
2025/01/01 00:54:08-INFO-Logger-line:44 - File "/home/au001v/vRMS/lib/python3.9/site-packages/paramiko/sftp_client.py", line 856, in _request
[...]

PR #507, opened 2025-01-22
Switch to Paramiko SSHClient with Timeouts & Logging to Investigate Rare Lockups

Cybis320:
We’ve had sporadic reports of lockups that are not reproducible in our test environment. The lockups are happening in UploadManager:
```
2025/01/13 14:28:25-INFO-StartCapture-line:683 - Uploading data before exiting...
2025/01/13 14:28:25-INFO-UploadManager-line:222 - Establishing SSH connection to: gmn.uwo.ca:22...
2025/01/13 14:28:26-INFO-UploadManager-line:76 - Trying ssh-agent key b'b3e09fd48cfea47fcf9d502c74436898'
2025/01/13 14:28:27-INFO-UploadManager-line:81 - ... success!

---------------- HANGS HERE INDEFINITELY --------------

2025/01/13 14:28:27-INFO-UploadManager-line:276 - Copying /home/rms/RMS_data/ArchivedFiles/US001P_20250113_003623_928590_detected.tar.bz2 (60.14MB) to files/US001P_20250113_003623_928590_detected.tar.bz2
2025/01/13 14:29:39-INFO-UploadManager-line:290 - File upload verified: files/US001P_20250113_003623_928590_detected.tar.bz2
2025/01/13 14:29:39-INFO-UploadManager-line:495 - Upload successful!
```

### This PR
- **Replaces** the low-level `Transport` usage with Paramiko’s recommended `SSHClient().connect()` approach.  
- **Adds** generous (5 minutes) connection-level timeouts (`timeout`, `banner_timeout`, `auth_timeout`) to avoid infinite waits on bad networks.  
- **Enables** a 30‑second keepalive (`transport.set_keepalive(...)`) so idle connections aren’t silently dropped on slow machines/connections.  
- **Refactors** authentication fallback—if the primary private key fails, it still falls back to the SSH agent.  
- **Introduces** more debug logging (including Paramiko version) to help us diagnose if issues persist.  
- **Increases** retry/backoff from 2 to 10 seconds, giving more time for the network to stabilize between attempts.
- **Ensures** we properly close SSH and SFTP sessions in a `try/finally` block to prevent lingering connections.

### Why?
- I haven’t been able to reproduce the reported lockups locally. This change is my best guess at a fix and prevention.  
- The expanded debug logs will make it easier to pinpoint what’s happening if a station still sees hangs.
- Properly closing resources should reduce the risk of orphaned sessions that might have caused issues.

### Sample Log:
```
025/01/22 21:31:53-INFO-DownloadMask-line:50 - Checking for new mask on the server...
2025/01/22 21:31:53-INFO-UploadManager-line:144 - Paramiko version: 3.4.1
2025/01/22 21:31:53-INFO-UploadManager-line:145 - Establishing SSH connection to: gmn.uwo.ca:22...
2025/01/22 21:31:54-INFO-UploadManager-line:161 - SSHClient connected successfully (key file).
2025/01/22 21:31:54-INFO-UploadManager-line:184 - Keepalive set to 30 seconds
2025/01/22 21:31:54-INFO-UploadManager-line:187 - Opening SFTP channel...
2025/01/22 21:31:54-INFO-UploadManager-line:189 - SFTP channel opened.
2025/01/22 21:31:55-INFO-DownloadMask-line:89 - files/masks exists
2025/01/22 21:31:55-INFO-DownloadMask-line:107 - Most recent flat /home/bolide/RMS_data/US005A/CapturedFiles/US005A_20250122_013222_354396/flat.bmp
2025/01/22 21:31:55-INFO-DownloadMask-line:112 - Uploading to files/masks/ as US005A_20250122_flat.bmp
2025/01/22 21:31:56-INFO-DownloadMask-line:122 - Not removing newly uploaded file US005A_20250122_flat.bmp
2025/01/22 21:31:56-INFO-DownloadMask-line:142 - No new mask on the server!
```

### DIFF
--- a/RMS/DownloadPlatepar.py
+++ b/RMS/DownloadPlatepar.py
@@ -85,10 +85,12 @@ def downloadNewPlatepar(config):
 
     finally:
         if sftp:
+            log.debug("Closing SFTP channel")
             sftp.close()
         if ssh:
+            log.debug("Closing SSH client connection")
             ssh.close()
-
+            
     return True
 
 
--- a/RMS/DownloadPlatepar.py
+++ b/RMS/DownloadPlatepar.py
@@ -34,7 +34,7 @@ def downloadNewPlatepar(config):
         # Connect with timeouts
         ssh, sftp = getSSHClientAndSFTP(
             config.hostname,
-            port=22,
+            port=config.host_port,
             username=config.stationID.lower(),
             key_filename=config.rsa_private_key,
             timeout=60,
@@ -90,7 +90,7 @@ def downloadNewPlatepar(config):
         if ssh:
             log.debug("Closing SSH client connection")
             ssh.close()
-            
+
     return True
 
 
--- a/RMS/DownloadPlatepar.py
+++ b/RMS/DownloadPlatepar.py
@@ -2,17 +2,18 @@
 
 from __future__ import print_function, division, absolute_import
 
+import logging
 import os
 from os.path import exists as file_exists
 
 import paramiko
 
-from RMS.Logger import getLogger
-from RMS.UploadManager import _agentAuth
+
+from RMS.UploadManager import getSSHClientAndSFTP
 from RMS.Misc import RmsDateTime
 
 # Get the logger from the main module
-log = getLogger("logger")
+log = logging.getLogger("logger")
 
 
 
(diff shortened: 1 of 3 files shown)


Cybis320:
This PR might also help address #494

RMS/DownloadPlatepar.py
```diff
@@ -25,69 +26,71 @@ def downloadNewPlatepar(config):
...
-        t = paramiko.Transport((config.hostname, config.host_port))
-        t.start_client()
-
-        # Authenticate the connection
-        auth_status = _agentAuth(t, config.stationID.lower(), config.rsa_private_key)
-        if not auth_status:
+        # Connect with timeouts
+        ssh, sftp = getSSHClientAndSFTP(
+            config.hostname,
+            port=22,
```
adalava: please use config.host_port

Cybis320: Done. Thank you.

adalava:
I really like this change, sometimes my stations "freeze" for a long time before starting because connection is silently dropped and it has to wait all retries before continuing. But I'd recommend to find out  what are the actual connection rate limit set at UWO firewall or stations may be locked out "forever".

Some time ago I figured out that connections requests to GMN's port 22 are silently ignored after 5 consecutive tries in a short period, no matter if connection was established successfully or not. I didn't determine the amount of time but you can surely reproduce it connecting 5 times in 1 minute (for instance `echo "ls" | sftp br0001@gmn.uwo.ca:files/processed).` The IP address gets banned temporarily and you can only reconnect after few minutes (something between 10 and 30 minutes). I don't know what happens if a station tries to continually in 60 seconds interval in a loop, there's a small chance that firewall will never unban that IP address. Maybe it would require to increase the timeout period dinamically between retries.

When RMS starts it already waste 2 or 3 tries checking for new platepar and mask on the server, or uploading processed files (that could be improved in the future to do all this on a single shot, or keep the SSH session alive during startup?). If user has more than one station on the same IP address or multistation scheme, it already waste all the 5 connections allowed, so would be good to check with the IT services the behavior and see if can be changed.

Cybis320:
Agreed. I'm trying to find out. Yesterday, I got locked out after testing a script that made too many consecutive requests. I'm leaving this as a draft PR until we get better clarity.

> I really like this change, sometimes my stations "freeze" for a long time before starting because connection is silently dropped and it has to wait all retries before continuing. But I'd recommend to find out what are the actual connection rate limit set at UWO firewall or stations may be locked out "forever".
> 
> Some time ago I figured out that connections requests to GMN's port 22 are silently ignored after 5 consecutive tries in a short period, no matter if connection was established successfully or not. I didn't determine the amount of time but you can surely reproduce it connecting 5 times in 1 minute (for instance `echo "ls" | sftp br0001@gmn.uwo.ca:files/processed).` The IP address gets banned temporarily and you can only reconnect after few minutes (something between 10 and 30 minutes). I don't know what happens if a station tries to continually in 60 seconds interval in a loop, there's a small chance that firewall will never unban that IP address. Maybe it would require to increase the timeout period dinamically between retries.
> 
> When RMS starts it already waste 2 or 3 tries checking for new platepar and mask on the server, or uploading processed files (that could be improved in the future to do all this on a single shot, or keep the SSH session alive during startup?). If user has more than one station on the same IP address or multistation scheme, it already waste all the 5 connections allowed, so would be good to check with the IT services the behavior and see if can be changed.

PR #512, opened 2025-01-30
Fix mask file log messages in DetectionTools

adalava:
Avoid printing 'No mask file has been found' continuously on the console when mask is all white. Also add this info to the log, that will be like this:

2025/01/30 04:46:05-INFO-DetectionTools-line:69 - Loaded mask: /root/RMS_data/config/BR001Z/mask.bmp (all white)


adalava:
There are several "print" on this file that sends the same text to both log and console. Are they still required?

PR #522, opened 2025-02-18
minor changes to space management

MaadhyamRana:
- renamed '_frametimes' suffix to '_FT' in DeleteOldObservations
- disabled quota management deletion by default

### DIFF
--- a/.config
+++ b/.config
@@ -286,8 +286,7 @@ times_days_to_keep: 8
 ; the next settings configure the archive management by space used on disk
 ; disabling quota management generates the reports, lists the files
 ; that would be deleted but prevents any action from being taken
-; true = disabled, false = enabled
-quota_management_disabled: true
+quota_management_enabled: false
 
 ; Any one of the next three parameters missing or commented out disables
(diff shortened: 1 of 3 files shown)


dvida:
Thanks - it says in the comment that the quota managements has been disabled for testing purposes, is that still the case or we're no longer testing?

MaadhyamRana:
I only made the change to reflect what's present in the .config template, below is the section from the config. quote_management_disabled is 'true' by default:

```
; the next settings configure the archive management by space used on disk
; disabling quota management generates the reports, lists the files
; that would be deleted but prevents any action from being taken
; true = disabled, false = enabled
quota_management_disabled: true
```

g7gpr:
I'd prefer it to remain disabled by default; I'm testing it on several stations, including testing the interaction with the traditional quota management schemes. I'll report back when I'm satisfied that it does exactly what was intended.

g7gpr:
However, @Cybis320 getting rid of the double negative, and using quota_management_enabled would be a good step to take.

Issue #520, opened 2025-02-10
Issue with 'prioritize_capture_over_reprocess' option and behavior change


adalava:
I just noticed that an option "prioritize_capture_over_reprocess" was introduced (in July/2024), and it has default value set to "false". I expected it to be "true" by default, since the behavior used to be "continue capturing if it's time to capture and process unprocessed folder later". Was it intentional for some reason?

Also, if RMS is restarted after 4 hours of capture, when it returns it's reprocessing that 4 hours of captures instead of continue the previous capture session. As an example, I rebooted a server responsible for 4 cameras around 03:30 UTC and 45 minutes later the four stations are still reprocessing the first part of the night and didn't start capturing. Log from one station:

```
025/02/10 03:30:39-INFO-StartCapture-line:1086 - Next start time: True UTC
2025/02/10 03:30:39-DEBUG-StartCapture-line:767 - Checking for folders containing partially-processed data
2025/02/10 03:30:40-DEBUG-StartCapture-line:792 - Checking folder: BR000M_20250126_222641_283847
2025/02/10 03:30:40-DEBUG-StartCapture-line:838 -     ... fully processed!
2025/02/10 03:30:40-DEBUG-StartCapture-line:792 - Checking folder: BR000M_20250127_222620_353620
2025/02/10 03:30:40-DEBUG-StartCapture-line:838 -     ... fully processed!
2025/02/10 03:30:40-DEBUG-StartCapture-line:792 - Checking folder: BR000M_20250206_032543_408652
2025/02/10 03:30:40-DEBUG-StartCapture-line:838 -     ... fully processed!
2025/02/10 03:30:40-DEBUG-StartCapture-line:792 - Checking folder: BR000M_20250207_222050_684633
2025/02/10 03:30:40-DEBUG-StartCapture-line:838 -     ... fully processed!
2025/02/10 03:30:40-DEBUG-StartCapture-line:792 - Checking folder: BR000M_20250208_021456_565101
2025/02/10 03:30:40-DEBUG-StartCapture-line:838 -     ... fully processed!
2025/02/10 03:30:40-DEBUG-StartCapture-line:792 - Checking folder: BR000M_20250209_234320_627108
2025/02/10 03:30:40-INFO-StartCapture-line:842 - Found partially-processed data in /root/RMS_data/BR000M/CapturedFiles/BR000M_20250209_234320_627108
[...]

PR #523, opened 2025-02-19
fix for auto reprocessing during standard capture

MaadhyamRana:
Suggested fixes for Issue #520


MaadhyamRana:
Sure, and this was also tested on station US000F.

Issue #536, opened 2025-03-14
RMS_Update.sh (prerelease) not working for one camera per user systems


g7gpr:
The line

https://github.com/CroatianMeteorNetwork/RMS/blame/41ce31f03f01590856ce9c722915af9ad6b741ad/Scripts/RMS_Update.sh#L15

added in c5fd3ab5ed1c2b08a175441e0b286836530e64ee creates a lockfile in the /tmp directory. Other lockfiles, or similar are created in $RMSBACKUPDIR, which is specific to the camera and username.

https://github.com/CroatianMeteorNetwork/RMS/blob/41ce31f03f01590856ce9c722915af9ad6b741ad/Scripts/RMS_Update.sh#L13

This seems to be breaking the update process on the one camera per username systems by not allowing the .config file to be restored after the update is completed.

g7gpr:
Resolved. Thank you.

PR #537, opened 2025-03-15
Change from global to user specific lock file

Cybis320:
Address Issue #536 by making the lock file user specific.
(Typing this on the road - did not test yet)


Cybis320:
@g7gpr, thanks for the report. I've incorporated your solution in this PR. I've tested it on a single user MCL. Let me need if it fixes the issue on the multi-user MCL.

g7gpr:
I can test your branch over the next few days, and will report back.

g7gpr:
Looks good, thank you.

Issue #549, opened 2025-04-01
DeleteOldObservations deleting all logfiles when logdays_to_keep: 0


g7gpr:
I believe that

```
; Number of days of logs to keep. Default is 30
logdays_to_keep: 0
```

Should inhibit deleting logfiles, however, this setting is deleting all the logfiles. I think a line like this

```
if config.logdays_to_keep > 0:
```

needs to be inserted somewhere.

PR #551, opened 2025-04-01
Permit log file deletion only when days_to_keep > 0 

g7gpr:
The pull request refers to Issue #549. 

When 

```
logdays_to_keep: 0
```

Deletion of logfiles should be inhibited. This pull request permits log file deletion only when logdays_to_keep > 0.

### DIFF
--- a/.config
+++ b/.config
@@ -259,7 +259,8 @@ extra_space_gb: 5
 ; Number of ArchivedFiles folders to keep. Default is 20
 arch_dirs_to_keep: 20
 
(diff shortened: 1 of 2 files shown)


dvida:
I agree with the change, chat you document the "param = 0" behaviour in the config file?

g7gpr:
It is already documented, but was not implemented. 

```
; By default the system will keep 20 ArchivedFiles folders and compressed
; versions of the folders. You can adjust these limits using the below
; parameters. A value of zero means do nothing ie keep everything forever.
```

PR #557, opened 2025-04-08
Address RMS_Update not restoring custom files

Cybis320:
This PR address an issue where RMS_Update would exit without restoring custom config and mask files if git pull failed.


Cybis320:
Added git fetch --prune -all to help keep local repo clean

PR #561, opened 2025-04-11
Only use one thread to generate timelapse

Cybis320:
Generating timelapse files using too many cpu cores while continuing to capture can cause frames to drop. 
This PR limit mp4 making to one thread.


satmonkey:
Is this an intended performance reduction or just a measure for some particular cases? I see no reason why this should be applied if no capture is running. Maybe the justification is just too short.

PR #565, opened 2025-04-16
Remove and replace `python setup.py instal`l with `pip install -e .`

Cybis320:
This PR remove and replace `python setup.py instal`l with `pip install -e .`

Change the update script to uninstall and RMS old RMS egg.

Gives clearer error messages when root config is missing or invalid.


Cybis320:
Another plus, is that RMS shouldn't crash at runtime anymore if the .config is missing from root in MCL. It will just compile with the default build values during update.

PR #569, opened 2025-04-29
Restrict PyGObject version due to higher Glib requirements  

Cybis320:
PyGObject >= 3.51.0 (September 2024 or newer) requires Glib >= 2.80 which is only available for Ubuntu 24.04.
Currently, PyGObject will fail to install on a fresh install when RMS_Update is run on all but Ubuntu 24.04 machines.

Proposing to limit PyGObject<=3.50 in requirements.txt.


dvida:
Thanks!
Is this targeted for master because it's critical and making operational systems fail?
If not, let's put it into prerelease.

Cybis320:
It's not making existing systems fails. But, new systems or re-imaged systems fail back to cv2 since September unless on Ubuntu 24.04. 

In the meantime, for new installs which don't yet have this package, manually run from within vRMS:
pip install PyOBject==3.50
Then run RMS_Update again.

PR #580, opened 2025-05-09
Unpack pmra before applying proper motion

Cybis320:
Gaia DR3 writes pmra as “mu_alpha_star” (pmra times cos dec) to keep numbers small near the poles.

This PR divides pmra by cos dec in the code when computing the new RA and make sure the resulting RA is within 0-360 range.


Cybis320:
This pretty much nails it:
Dec_vs_PMRA

PR #581, opened 2025-05-12
Update StarCatalog.py

dvida:


Cybis320:
Doesn't this risk division by zero. Also the resulting ra still risk ending up outside 0-360 range.

dvida:
I just deleted the extra spaces between binary operators, the code is the same as your PR. Did I do something wrong?

Cybis320:
Sorry, no. I didn't see you had already merged the other one. All good.

PR #584, opened 2025-05-18
Reboot camera and check connection after mode switching

Cybis320:
Camera image settings become inconsistent after DayNightColor mode changes. Specifically, when switching from night to day mode, images exhibit poor quality (crushed shadows without increased highlight details) until the camera is rebooted. Image quality is restored only after a camera reboot.

This PR implements a camera reboot after mode switching and adds connection verification to ensure proper camera operation.

The revised process flow is as follows:
1. Set camera mode switching flag when capture mode changes.
2. Probe camera connection to verify availability.
3. Apply camera mode switching commands, now including an automatic reboot via camera_settings.json.
4. Wait 5 seconds for the camera to register all commands (NEW).
5. Verify camera connection re-establishment (NEW).
6. Proceed with normal capture operations.

Note that existing prerelease testers would have to manually edit their custom `camera_settings.json` to add a reboot command. This PR only changes the `camera_settings_template.json`.

I would still recommend leaving the auto reboot as is (daily reboot) as a bricked camera may be unreachable until it auto-reboots.

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1116,12 +1116,13 @@ class BufferedCapture(Process):
 
                             try:
                                 cc.cameraControlV2(self.config, "SwitchMode", mode_name)
+
+                                # create empty sentinel file
+                                open(flag_file, "a").close()
+                                log.info("Init complete - flag written to %s", flag_file)
+
                             except Exception as e:
                                 raise RuntimeError("Failed to switch camera mode: {}".format(e))
-
-                            # create empty sentinel file
-                            open(flag_file, "a").close()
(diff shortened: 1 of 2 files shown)


Cybis320:
I narrowed down the issue. Without a reboot, brightness settings (and possibly other settings as well) revert to default values after changing DayNightColor mode.

To reproduce:
1. In color mode, set brightness to 100
2. Switch to night mode and back to color mode
3. The image is now at brightness 50 (default)
4. Set brightness to 100 - nothing happens as the camera thinks it's already at 100
5. Set brightness to 50 - nothing happens as the setting is now in agreement with the actual state of the camera
6. Set brightness to 100 - tada! Image brightens to 100

PR #589, opened 2025-05-21
Add utility to upgrade config to configTemplate format preserving attributes

Cybis320:
Adds a new utility script, MigrateConfig.py, that:
	1.	Reads existing station .config file (single or multi cam)
	2.	Merges it with the latest .configTemplate, carrying forward any non-default values.
	3.	Writes the migrated file next to the original (configNew_<CamId> by default), or in a user-supplied location.
	4.	Can optionally back-up and replace the active .config in a single --update run.
	5.	Logs every change to <CamId>_MigrateConfig.log.

# Usage:

## Dry-run migration (creates configNew_XXxxxx)
`python -m Utils.MigrateConfig.py`

## Happy? Do the swap in place (automatic backups)
`python -m Utils.MigrateConfig.py --update`

Key options
-i/--input  path to old .config
-t/--template  template file
-o/--output  write migrated file here
-u/--update  replace live config after backup
-r/--recent  also apply recently-changed template defaults

Implementation highlights
	•	Uses AuditConfig.extractConfigOptions to ignore attributes no longer recognized.
	•	Handles the legacy flag quota_management_disabled (auto-writes the inverse value to quota_management_enabled, drops the old key).

Acknowledgement

Big thanks to Steve Matheson for the work on this.

⸻

Safe to merge—no existing modules touched.

### DIFF
--- a/Utils/MigrateConfig.py
+++ b/Utils/MigrateConfig.py
@@ -9,12 +9,15 @@ Quick usage example
 -------------------
 Upgrade single or multi stations config (with backups) to latest configTemplate format preserving attributes:
 
-python -m Utils.MigrateConfig.py -u
+python -m Utils.MigrateConfig -u
 
 To first test the upgrade without applying it, run:
 
-python -m Utils.MigrateConfig.py
+python -m Utils.MigrateConfig
 
+# To incorporate recent default values (e.g. star catalog file) use the -r option:
+
+python -m Utils.MigrateConfig -r
 '''
 import sys
 import os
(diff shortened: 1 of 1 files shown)


dvida:
Awesome work!

Could you:
- add: from __future__ import print_function, division, absolute_import so it doesn't crash on Python 2
- modify the function names to conform to camelCase?

Just a reminder:
constants - CAPS
variables - snake_case
function - camelCase
classes - PascalCase

PR #593, opened 2025-05-26
Add memorial note for Global Meteor Network members in runCapture function

dvida:
Include a tribute to deceased members of the Global Meteor Network within the runCapture function.


g7gpr:
Great idea.

Issue #609, opened 2025-05-30
[prerelease] Utils.MigrateConfig removes settings from .config file


g7gpr:
Config file before migration contains the following settings

```
[FireballDetection]

; Flag for enabling/disabling fireball detection
enable_fireball_detection: true
; Subsample to 16x16 squares (default 16)
subsampling_size: 16
; Weight for stddev in thresholding for fireball extraction
k1: 7.0
; Absolute offset in thresholding for fireball extraction
j1: 10
```

and

```
[MeteorDetection]

; Minimum number of stars required in order to run the detection
ff_min_stars = 20

; Binning (only supported for videos, images, and vid files, but no FF files!)
; -------
; Bin images before detection (has to be a factor of 2, e.g. 2, 4, 8, etc.). The X, Y coordinates in
; detections will be rescaled to the original size. 1 = do not bin.
detection_binning_factor: 1
; The image can be binned by either averaging ('avg') or summing ('sum') pixel intensities in the bin window.
; Note that the output image data type is uint16, so be careful of integer overflows when using "sum" method!
detection_binning_method: avg

; Thresholding and KHT parameters
; -------------------------------
; Weight for stddev in thresholding for faint meteor detection
k1: 3.5
; Absolute levels above average in thresholding for faint meteor detection
j1: 12
; Maximum ratio of white to all pixels on a thresholded image (used to avoid searching on very messed up
; images)
```

after migration 

```
[FireballDetection]

; Flag for enabling/disabling fireball detection
enable_fireball_detection: true

; Subsample to 16x16 squares (default 16)
subsampling_size: 16

; Weight for stddev in thresholding for fireball extraction
k1: 7.0

; Absolute offset in thresholding for fireball extraction
j1: 10
```

but k1 is missing from the next section

```
[MeteorDetection]

; Minimum number of stars required in order to run the detection
ff_min_stars: 20

; Binning (only supported for videos, images, and vid files, but no FF files!)
; -------
; Bin images before detection (has to be a factor of 2, e.g. 2, 4, 8, etc.).
; The X, Y coordinates in detections will be rescaled to the original size.
; 1 = do not bin.
detection_binning_factor: 1

; The image can be binned by either averaging ('avg') or summing ('sum') pixel
; intensities in the bin window. Note that the output image data type is uint16,
; so be careful of integer overflows when using "sum" method!
detection_binning_method: avg

; Thresholding and KHT parameters
; ----
[...]

PR #612, opened 2025-05-31
MigrateConfig - flush file buffers

g7gpr:
Flush file buffers to prevent settings getting lost, refer to issue #609


dvida:
Stellar work, this one's a classic.

Issue #362, opened 2024-08-17
StartCaptue option to suppress checking GMN for revised platepar & mask


peschman:
I would like to request an added StartCapture option, -t, which suppresses GMN checking for revised .config, mask or platepar. This option would only be used for short test captures on a .config that is already set to upload: false.

Or, even better, if upload_enabled: false, then suppress checking for revised files, since this would not require a new cmd line option.

Without this capability a test with -d 0.03 spends more time timing out, while waiting for new files, than it does actually running the test capture and processing.

Thanks,
Peter E.

peschman:
Sorry about the subject typo: StartCapture is what it should have said.

g7gpr:
Interesting thought. I don't like the idea of skipping the checks for updated files if upload is false, because at least two operators, and multiple stations do not use the standard RMS upload mechanism, and have upload_enabled set to false. I think you are referencing a special case where your private key does not match the public key of the account you are trying to connect to.

peschman:
Given the problems that might be caused by relying on upload_enabled: false, I think the -t or other char to indicate running in test mode would be the best solution. This would allow test runs to complete without having to wait for comms timeouts when looking for new mask or platepar on GMN.

Cybis320:
How about adding a `download_enable` flag?

g7gpr:
Why is that better than the original idea?

Cybis320:
Sorry, which one exactly?

g7gpr:
Just adding a -t or --test flag to disable server connections so you can quickly start and run capture on a station with no keys.

Cybis320:
No, I think that's an excellent idea. 

Additionally, we currently disable upload by setting upload_enabled to false when station id is XX0001. We could use the same logic for download. 

Also, we have a download mask permissive flag to disable downloading mask but there's no equivalent for platepar. Is there a use case for only disabling mask but not pp download. If not a generic download_enabled option could disable both.

g7gpr:
I need to think about disabling download when XX0001 is stationID. It sounds like a good idea.

g7gpr:
I've implemented it, refer to above PR.

peschman:
I do a lot of testing using secondary or tertiary streams of IMX-291, where the primary stream uploads to GMN, but the testing streams, lasting all night, or all day and night, do not upload.

In this case it would be best if authentication was also skipped if .config is set to 
upload_enabled: false

[...]

PR #615, opened 2025-06-02
Skip authentication

g7gpr:
To assist with testing, a new command line switch is added

```
python -m RMS.StartCapture -s
```

or

```
python -m RMS.StartCapture --skip_authentication
```

will run capture, however any activities requiring authentication will be skipped.

This is in response to issue #362.

Additionally stationID XX0001 will also cause any activites requiring authentication to be skipped.

### DIFF
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -50,7 +50,7 @@ log = getLogger("logger")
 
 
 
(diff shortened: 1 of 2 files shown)


g7gpr:
Workaround found. This PR no longer required.

Issue #626, opened 2025-06-07
SkyMap - add more functions


g7gpr:
Add features to FOVSkyMap to

* show the pointing on the plot in degrees
* show the field of view on the plot in degrees
* allow output filename and path to be specified
* optionally rotated the text plotted on the chart to be easier to read, although having the text aligned with the camera is better for working on paper

PR #627, opened 2025-06-07
Add features to FOVSkyMap

g7gpr:
This pr is related to issue #626 

Original functionality

```
python -m Utils.FOVSkyMap ~/ExampleStationData/
```

Produces the file in 

```
~/source/RMS/fov_sky_map.png
```

fov_sky_map

```
python -m Utils.FOVSkyMap ~/ExampleStationData/ -o "~/RMS_data/walnut.png"
```

Saves the same file to the specified path.

Switch -r rotates the text to be aligned with the camerawalnut
walnut

Adding the -l switch prevents the text being upside down

walnut

Which also works when -n (north up is specified)

walnut

-p reduces the text one size, and shows the pointing.

walnut

finally -f shows the fov, the complete command being

```
python -m Utils.FOVSkyMap -rlnpf ~/ExampleStationData/ -o "~/RMS_data/walnut.png"
```
walnut

Although, I'd prefer not to flip the text

```
python -m Utils.FOVSkyMap -rnpf ~/ExampleStationData/ -o "~/RMS_data/walnut.png"
```
walnut


Cybis320:
I love this, thank you! When -o is not specified, it used to save in the same dir as the input dir, now it saves in the RMS root directory. It's fine either way (although I prefer to save in the Stations dir).

Here's my station (ignore the doubling of the north camera)
fov_sky_map

g7gpr:
That is a very well organised station; here is my equivalent with 6x6mm + 2x4mm

image

g7gpr:
@cybis320 "When -o is not specified, it used to save in the same dir as the input dir,"

I think this should be addressed. Folks might be using this in a script, and if the behaviour has changed, it might upset things. I'll look into reproducing the original behaviour.

Issue #639, opened 2025-06-10
Unable to run RMS.Reprocess due to circular dependency


markmac99:
since the last release i am unable to run RMS.Reprocess on older data. I get an error that getLogger can't be imported, as shown below. This is happening on all three of my stations that are up to date with the master branch.

It looks like a circular dependency issue since RMS.Logger imports RMS.Misc which imports RMS.Logger. This might affect other standalone modules too. 
``` bash
(vRMS) pi@uk001l:~/source/RMS $ python -m RMS.Reprocess /home/pi/RMS_data/CapturedFiles/UK001L_20250609_210925_307794/
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/pi/source/RMS/RMS/Reprocess.py", line 15, in <module>
    from RMS.ArchiveDetections import archiveDetections, archiveFieldsums, archiveFrameTimelapse
  File "/home/pi/source/RMS/RMS/ArchiveDetections.py", line 10, in <module>
    from RMS.Logger import getLogger
  File "/home/pi/source/RMS/RMS/Logger.py", line 15, in <module>
    import RMS.Misc
  File "/home/pi/source/RMS/RMS/Misc.py", line 35, in <module>
    from RMS.Logger import getLogger
ImportError: cannot import name 'getLogger' from 'RMS.Logger' (/home/pi/source/RMS/RMS/Logger.py)
```

PR #640, opened 2025-06-10
Rollback changes ref circular dependencies

markmac99:
This reverts a4bd454 "removed duplicated code while preventing circular imports" which unfortunately broke RMS.Reprocess and possibly some other standalone modules (see Issue #639 )  My google search did not find any workaround for this, so for now I think we'll have to live with the duplication until we cease support for Python 2.7 and can replace mkdirP and the utc-handling classes with generic Python 3.x code throughout the entire codebase.

### DIFF
--- a/RMS/Logger.py
+++ b/RMS/Logger.py
@@ -12,7 +12,6 @@ import threading
 import atexit
 import time
 
-import RMS.Misc
 
 try:
     from logging.handlers import QueueHandler  # Python 3.2+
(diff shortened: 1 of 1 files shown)


Cybis320:
Thank you for flagging this. Your solutions works but another option is to move the getLogger() function just above import RMS.Misc in Logger.py. This will also avoid the circular import without the need to duplicate code.

markmac99:
That does indeed work, though its not the python style (see PEP8: imports should be at the top of the file). It also relies on getLogger() not using any capabilities in RMS.Misc and on nobody unknowingly reorganizing the code or adding further functionality before the import. 
Personally i think it'd be simpler to accept duplication of these three functions for now. Once we drop Python 2.7 support we won't need mkdirP or the other functions and can refactor more simply.

The other alternative would be to move the functions into Logger.py. I didn't check every file but most of the code that i did check already imports Logger.

markmac99:
ps - the other reason i'm not comfortable moving the import is that it'll only solve the problem for that one function. If somewhere else in the codebase, we need to import other functionality from Logger, then the problem could re-emerge and it'll be quite confusing because importing getLogger will work but importing shutdownLogging or initLogging might generate an error.  I think this is one of the reasons for the recommendation in the PEP.

Cybis320:
Ok, yes, let's add it back until we drop py2. Ty.

Cybis320 (approved):
I agree that this the most reliable solution despite the duplications.

PR #645, opened 2025-06-11
set pre-init logger to INFO

markmac99:
Set logging to INFO rather than DEBUG to reduce excessive logging from paramiko, matplot etc.  Hopefully no em-dashes in the file this time.

PR #647, opened 2025-06-11
Remove ad-hoc logger filters and only allow logging from RMS modules

Cybis320:
This PR filters out any log messages emanating from modules outside the RMS codebase.

### DIFF
--- a/RMS/Astrometry/ApplyAstrometry.py
+++ b/RMS/Astrometry/ApplyAstrometry.py
@@ -829,6 +829,102 @@ def applyPlateparToCentroids(ff_name, fps, meteor_meas, platepar, add_calstatus=
     return meteor_picks
 
 
+def applyPlateparToRaDecCentroids(ff_name, fps, meteor_meas, platepar, add_calstatus=False):
+    """ Given the meteor centroids in RA and Dec, compute X,Y image coordiantes and Alt/Az, as well as the
+        photometry. 
+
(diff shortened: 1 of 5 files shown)


Cybis320:
@dvida, prerelease should first be synched with master to show the actual diffs created by this PR.

markmac99:
I can't support this PR as it stands, as it suppresses ALL third-party logging, including any logging generated by external scripts.

Cybis320:
The external is script is using RMS logger, but is located outside the RMS repo, right? Hm, that would indeed be blocked.

markmac99:
I think i can see how to work around this by checking and adding the external script path to the test in the filter - just testing it now.

markmac99:
and yes, we never put any of our scripts in the RMS source tree, otherwise things get messy!  :)

Cybis320:
OK, added the external script path specified in the config to the white list and incorporated your PR #645. Would this work?

markmac99:
Tested and it does capture logging from external scripts.  Thanks for sorting that out.

The logging still goes into the RMS log however. It seems that the new RMS logger functionality overrules all attempts to configure a separate logger!  I'll continue to investigate that.

markmac99:
NB: Please can this be reviewed and rolled out as soon as possible as i am concerned about the potentially sensitive debug detail currently being leaked into the RMS logs on UKMON stations.

PR #653, opened 2025-06-15
Refresh current time before computing time to wait

Cybis320:
Currently, there is an error when computing the time_to_wait before switching mode if capture_wait_seconds is not zero (used to stagger switching in a multi-camera setup). The time_to_wait uses a current_time that is no longer current by the time of the calculation. 

This PR uses a fresh current time.


g7gpr:
Does this cause all reprocessing to start at the same time?

PR #658, opened 2025-06-17
Proposed changes to logging levels

markmac99:
This update changes logging level of a few messages so that more is reported at the INFO level and a some mainly duplicated mesasges are dropped to DEBUG. It also updates the camera reboot message to clarify that its the camera rebooting.

Specific changes in logging are
* in Compression
  * change compression time to INFO so it can be monitored for issues
  * change saving time to INFO so it can be monitored for issues
* in DetectStarsAndMeteors, change "at least x stars" message to INFO so it can be monitored for issues
* in VideoExtraction change "nothing found" message to INFO so it can be monitored for issues

* in FrameInterface, add logging as previously it was using print statements so logs lost entirely!

* in ExtractStars change "extracted stars" to DEBUG as its already reported at INFO elsewhere
* in DetectionTools change "loaded mask" to DEBUG as its not necessary info - if the mask can't be loaded this is separately logged at INFO level and we want to log problems rather than nominal activity
* in BufferedCapture 
  *  change raw frames message to DEBUG as its already reported at INFO elsewhere
  * change frame age and FT file creation to DEBUG  

I also suppressed the fireball slope message.

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1373,10 +1373,8 @@ class BufferedCapture(Process):
                     self.pipeline.set_state(Gst.State.NULL)
                     self.pipeline = None
 
-                if abs(self.last_calculated_fps - self.config.fps) > 0.0005 and self.last_calculated_fps_n > 25*60*60:
-                    log.info('Config file fps appears to be inaccurate. Consider updating the config file!')
 
-                log.info("Last calculated FPS: {:.6f} at frame {}, config FPS: {}, resets: {}, startup status: {}"
+                log.debug("Last calculated FPS: {:.6f} at frame {}, config FPS: {}, resets: {}, startup status: {}"
                          .format(self.last_calculated_fps, self.last_calculated_fps_n, self.config.fps, self.reset_count, self.startup_flag))
 
                 log.info('GStreamer Video device released!')
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1906,7 +1906,7 @@ class BufferedCapture(Process):
                     
                     # For GStreamer, show elapsed time since frame capture to assess sink fill level
                     else:
-                        log.debug("Block's max frame age: {:.3f} seconds. Run's dropped frames: {}"
+                        log.info("Block's max frame age: {:.3f} seconds. Run's dropped frames: {}"
                                  .format(max_frame_age_seconds, self.dropped_frames.value))
 
                 last_frame_timestamp = frame_timestamp
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -49,12 +49,12 @@ import Utils.CameraControl as cc
 # Get the logger from the main module
 log = getLogger("logger")
 
-try:
-    # py3
-    from urllib.parse import urlparse
-except ImportError:
+if sys.version_info[0] < 3:
     # py2
     from urlparse import urlparse
+else:
+    # py3
+    from urllib.parse import urlparse
 
 
 GST_IMPORTED = False
@@ -1906,7 +1906,7 @@ class BufferedCapture(Process):
                     
                     # For GStreamer, show elapsed time since frame capture to assess sink fill level
                     else:
-                        log.info("Block's max frame age: {:.3f} seconds. Run's dropped frames: {}"
+                        log.debug("Block's max frame age: {:.3f} seconds. Run's dropped frames: {}"
                                  .format(max_frame_age_seconds, self.dropped_frames.value))
 
                 last_frame_timestamp = frame_timestamp
@@ -1995,7 +1995,7 @@ class BufferedCapture(Process):
                 else:
                     self.start_time2.value = first_frame_timestamp
 
-                log.info('New block of raw frames available for compression with starting time: {:s}'
+                log.debug('New block of raw frames available for compression with starting time: {:s}'
                          .format(str(first_frame_timestamp)))
 
             
@@ -2020,7 +2020,7 @@ class BufferedCapture(Process):
 
                 mkdirP(ft_subpath)
                 FTfile.write(ft, ft_subpath, ft_filename)
-                log.info("Created FT file {} for block starting at {}".format(os.path.join(ft_subpath, ft_filename), first_frame_timestamp))
+                log.debug("Created FT file {} for block starting at {}".format(os.path.join(ft_subpath, ft_filename), first_frame_timestamp))
 
                 # For Testing: 
                 # Print first and last 10 timestamps, array length, average time difference and time difference from last block
--- a/RMS/Compression.py
+++ b/RMS/Compression.py
@@ -317,14 +317,14 @@ class Compressor(multiprocessing.Process):
             # Cut out the compressed frames to the proper size
             compressed = compressed[:, :self.config.height, :self.config.width]
             
-            log.debug("Compression time: {:.3f} s".format(time.time() - t))
+            log.info("Compression time: {:.3f} s".format(time.time() - t))
             t = time.time()
             
             # Save the compressed image
             filename_millis, filename_micros = self.saveFF(compressed, startTime, n*256)
             n += 1
             
-            log.debug("Saving time: {:.3f} s".format(time.time() - t))
+            log.info("Saving time: {:.3f} s".format(time.time() - t))
 
 
             # Save a live.jpg file to the data directory
--- a/RMS/Compression.py
+++ b/RMS/Compression.py
@@ -354,7 +354,7 @@ class Compressor(multiprocessing.Process):
 
                 # Add the file to the detector queue
                 self.detector.addJob([self.data_dir, filename, self.config])
-                log.info('Added file for detection: {:s}'.format(filename))
+                log.debug('Added file for detection: {:s}'.format(filename))
 
 
 
--- a/RMS/DetectStarsAndMeteors.py
+++ b/RMS/DetectStarsAndMeteors.py
@@ -234,7 +234,7 @@ def detectStarsAndMeteors(ff_directory, ff_name, config, flat_struct=None, dark=
     # Run meteor detection if there are enough stars on the image
     if len(star_list[1]) >= config.ff_min_stars:
 
-        log.debug('At least ' + str(config.ff_min_stars) + ' stars, detecting meteors...')
+        log.info('At least ' + str(config.ff_min_stars) + ' stars, detecting meteors...')
 
         # Run the detection
         meteor_list = detectMeteors(img_handle, config, flat_struct=flat_struct, dark=dark, mask=mask)
--- a/RMS/DetectionTools.py
+++ b/RMS/DetectionTools.py
@@ -71,7 +71,7 @@ def loadImageCalibration(dir_path, config, dtype=None, byteswap=False):
     if mask_path:
         mask = MaskImage.loadMask(mask_path)
 
-        log.info('Loaded mask: {:s}'.format(mask_path))
+        log.debug('Loaded mask: {:s}'.format(mask_path))
 
         # If the mask is all white, set it to None
         if (mask is not None) and np.all(mask.img == 255):
--- a/RMS/EventMonitor.py
+++ b/RMS/EventMonitor.py
@@ -2143,9 +2143,9 @@ class EventMonitor(multiprocessing.Process):
                 next_check_start_time_str = next_check_start_time.replace(microsecond=0).strftime('%H:%M:%S')
                 log.info('Next EventMonitor run : {} UTC; {:3.1f} minutes from now'.format(next_check_start_time_str, int(self.check_interval)))
                 if time_left_before_start_minutes < 120:
-                    log.info('Next Capture start    : {} UTC; {:3.1f} minutes from now'.format(str(start_time.strftime('%H:%M:%S')),time_left_before_start_minutes))
+                    log.debug('Next Capture start    : {} UTC; {:3.1f} minutes from now'.format(str(start_time.strftime('%H:%M:%S')),time_left_before_start_minutes))
                 else:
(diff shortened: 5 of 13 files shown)


Cybis320:
Can we revert frame age to info?

Cybis320:
We don't want to clean it up so much as to force people interested in knowing whether their system is working well to go back to debug level. That would render the whole idea of split console/log file level idea much less useful. 

Personally, I want to glance at the screen and know whether it's been dropping frames and why. It would be a bummer if I had to turn on debug to see that. 

I see a low dropped frame number - I'm happy. I see a high dropped frames number - I then look at frame age or mean interval. 

We can talk about better ways to do it but I think we need something at info level.

satmonkey:
Can you also look at the mass messages showing the progress like:
"Uploading to ..."
"Stacking ..."
I can see they are hardcoded print commands, but flashing all terminal
content out, when they happen.
My preference would be to change the newline character to "" as a
parameter, so only the last print output is shown, e.g.:
print('Stacking: ', ff_name, end="")

M.

út 17. 6. 2025 v 14:03 odesílatel Mark McIntyre ***@***.***>
napsal:

markmac99:
@Cybis320 
"Personally, I want to glance at the screen and know whether it's been dropping frames and why. It would be a bummer if I had to turn on debug to see that."
I understand you want to see it, but you and i are not most users  :) The majority don't have a screen attached and dont monitor RMS in realtime (i don't either, in fact). Hopefully they  look at the daily reports which show dropped frames, delays etc, which is more user-friendly than a log file. Though in most cases i think they probably don't even do that unless someone like you or i notices a problem and asks them to check! 

Can i suggest we investigate whether we can report only when its a problem rather than reporting nominal behaviour. 

Would be good to bottom out the changes we want to make if we can. 

Ideally, aside from general progress, i think we should try to only report when problems occur. This would make it easy for users to implement some simple checks eg grep the log for WARNING - I'd love to be able to run a tool alongside RMS that monitors for warnings and alerts me via my OpenHab server. Also that way in Denis' proposed ncurses version we could highlight warnings in red or something so they'd be clear to users.

markmac99:
@satmonkey I can look into this.

markmac99:
@satmonkey, The stacking messages come from StackFFs and StackImgs which do not currently have any proper logging. I'll add it and figure out hte required level of logging.  I think this will also deal with any line-end issues as the logger will properly format the messages.

markmac99:
@satmonkey the "Uploading to..." message is already logged at INFO level and so should appear in any logfiles. 
```
(vRMS) pi@uk000f:~/RMS_data/logs $ grep Uploading log_UK000F_20250616_033738.log
2025/06/16 03:37:40-INFO-DownloadMask-line:110 - Uploading to files/masks/ as UK000F_20250615_flat.bmp
2025/06/16 21:14:20-INFO-DownloadMask-line:110 - Uploading to files/masks/ as UK000F_20250615_flat.bmp
```

markmac99:
@satmonkey oh, i think you mean the messages in uploadmanager . I believe these were added on purpose fairly recently so that there was an onscreen progress report on upload speed and status. 
I assume you're suggesting that we should not print a newline so that each status report overwrites the previous. This isn't logged, only printed onscreen. I can change the line ending i guess.

markmac99:
@satmonkey  ah, no i can't do this yet. The end= keyword is python-3-only and i don't want to have to wrap the print statements in a test for the version. Once we move to py3 only i can revisit it.  (only got a few weeks to wait!)

satmonkey:
@markmac99 - ok, that would be fine, I can survive a couple of  weeks more:)

út 17. 6. 2025 v 16:19 odesílatel Mark McIntyre ***@***.***>
napsal:

dvida:
Regarding dropped frames, I think it's super useful to report this, but only if there is a problem.
My proposal is:
- Log all frame timings stats to DEBUG
- If there are any dropped frames or the frame age is too old (I don't have a good feeling about what that might be), let's log that as WARNING and show it to the user.

This is how the old CAMS code worked and it was super useful. You definitely want to know if something is up so you can address the problem.

dvida:
I see that we have two threads going, I'm only going to reply to this one to avoid any confusion.

In terms of reporting a warning if there are frame drops, it should only be printed when a new frame drop occurs, say something like "x dropped frames in the last n seconds, N total dropped frames".

markmac99:
@dvida ok, will look into that. 

@satmonkey  i reliased i can actually do what you asked, will implement and test.

markmac99:
@dvida @Cybis320 my examination of the logs on my four cameras (two running cv2, two running gst) indicate that cv2 cameras may drop one or two frames every few minutes, and even the gst pipeline drops a few every so often, 

However i also noticed that a reasonable diagnostic is max block age / max block interval. On the gst cameras, whenever this goes above 2 i start to see more dripped frames. On the cv2 cameras, mean is 1.0 and max around 2.5. Frames often get dropped if its higher though its not consistent. 

We could warn for gst if max block age > 2 and for cv2 if max > 3 or in either case if the change in dropped frames < 10 ? 

I can pull a few other camera's logs and do more analysis if you prefer.

dvida:
This one is dropping a lot of frames: https://globalmeteornetwork.org/weblog/HR/HR000V/HR000V_20250616_192916_456752_detected/
I'm attaching its log for reference.
[log_HR000V_20250613_150444.log.20250616_1510.txt](https://github.com/user-attachments/files/20778723/log_HR000V_20250613_150444.log.20250616_1510.txt)

Interestingly, it sometimes reports a small frame age but it still drops.

markmac99:
Looks like that is a Pi3 rev 1.3 which dates back to 2018, so its probably at the limit of what RMS can run on now.

dvida:
It is, but it would still be good to report that it has issues.

markmac99:
I agree.  If we can determine some thresholds for the checks i'm happy to put something in place. Not sure what tho!

dvida:
@Cybis320 Do you have a good idea of what thresholds we should use?

Cybis320:
@dvida 
I don't have a simple answer. In a sense, it's like asking when should we display the fuel gauge? I think we are confusing the fuel gauge from the low-level alert light. One should always be displayed, and the other only when the level is low.

I'm reposting this from the other thread, which should answer your earlier question about dropping frame despite small frame age.

We use different methods to assess video stream health for the cv2 and gst pipelines because they collect timestamps differently.

**GST Pipeline:**
```
INFO: Block's max frame age: 0.342 seconds. Run's dropped frames: 0
```
**CV2 Pipeline:**
```
INFO: Block interval: mean 4.059, max 116.431 (normalized). Dropped frames: 24198
```
These metrics are crucial indicators of the client's ability to handle the video stream and whether stream buffers are reaching capacity.

If I had to choose a single log message to display in the console, this would be it. Dropping frames with low frame age/block interval suggests camera connection issues, while dropping with high values indicates the computer can't keep up with processing.

How the metrics work:

**GST Method - Frame Age:**

- Measures the age of retrieved frames (current time - frame timestamp)
- High values = buffer filling up = processing bottleneck

**CV2 Method - Block Interval:**

- Since we can't access frame timestamps directly, we calculate the mean interval between frames in 256-frame blocks
- Values are normalized to 1/fps (so 1.0 = real-time processing)
Mean > 1.0 = client falling behind
- The 'max' value shows the worst-case interval in the block (less critical but still informative)

**Dropped frames**
- The cumulative total number of missing frames since the start of the run

If we wanted to *add* a CAUTION level message (fuel gauge alert light) in addition to the fuel gauge, I would say for gst when the age is above 3 seconds or so (although there is probably a more precise way to determine the percentage of buffer being used), for cv2 when the mean is above 1.1 or something of that nature.

And a WARNING when the frame drop count is increasing. Frame drop with small frame age or interval means either a camera or network issue, so we could make the WARNING message smart by detecting whether we are dropping with small age/interval or dropping with large age/interval.

markmac99:
A better analogy would be to suggest that we display the mpg and tyre pressure continuously on the dash board, as well as having a warning light. We don't normally do that in cars because its not necessary for driving.  My car *can* show you the fuel consumption rate and tyre pressure, but you don't need to know it 99% of the time and so you have to click a few buttons to see it 

I keep coming back to this. I don't think we should be routinely printing nominal operation of technical components. Its fine to log that we're starting or completing a process o we can track what stage we're at and that RMS isn't stuck. But for technical details to be useful, we need to focus the logging on when things are wrong not when they're normal. 

Look at it this way: if we're struggling to decide on what is good and bad, how will most users who are considerably less technical than us, know whether the values are good, bad or merely soso?  Technical messages need to be useful to the audience. 

Bear in mind that the debug can be gathered by a one-off operation of setting the flag in the config. But 99% of users do not monitor in realtime, and probably only look at the logs once their system has failed.  In such a case they can easily enable debug and restart to capture any errors. 

I'm going to stop arguing my case now because i've reached the limits of my ability to explain why this is not the right approach.

markmac99:
Here's the log from UK0006 from last night, set to INFO via the config file, and including Luc's request to reinstate the Block messages but not Denis request to add logging to the Flux calcs. Log size is around 2MB which is a lot smaller than the previoust 4-5MB, so thats good. 

There are a couple of places i messed up the log message formatting, will fix those and retest on UK0006 tonight. We're forecast clear skies for several weeks so should be time to test! 

A couple of questions. Firstly, i saw some gstreamer c++ errors in the log at the time of end-capture, is this normal or indicative of an issue? 

Secondly, is the threshold test on the FPS check a bit tight? I'd have thought 24.981 was close enough given the quality of the hardware. 
```
2025/06/18 02:58:08-INFO-BufferedCapture-line:2041 - Releasing video device...
2025/06/18 02:58:08-INFO-Logger-line:270 - rtspsrc ../gst/rtsp/gstrtspsrc.c:6685:gst_rtsp_src_receive_response: receive interrupted
2025/06/18 02:58:08-INFO-Logger-line:270 - rtspsrc ../gst/rtsp/gstrtspsrc.c:6783:gst_rtspsrc_try_send: receive interrupted
2025/06/18 02:58:08-INFO-Logger-line:270 - rtspsrc ../gst/rtsp/gstrtspsrc.c:9238:gst_rtspsrc_pause: PAUSE interrupted

2025/06/18 02:58:10-INFO-BufferedCapture-line:1377 - Config file fps appears to be inaccurate. Consider updating the config file!
2025/06/18 02:58:10-INFO-BufferedCapture-line:1379 - Last calculated FPS: 24.981179 at frame 232192, config FPS: 25.0, resets: 4, startup status: False

```
[log_UK0006_20250617_195758.log](https://github.com/user-attachments/files/20794052/log_UK0006_20250617_195758.log)
[log_UK0006_20250617_195758.log](https://github.com/user-attachments/files/20794052/log_UK0006_20250617_195758.log)

markmac99:
As an aside, i was thinking i'd go through the code to see where we're notifying the user that something has gone wrong, and i'd bump those to "warning". This would mean one could grep the log for WARNING and quickly find potential problems. I appreciate this might be too much change for one PR tho, so i will wait till this one has been merged and then work in a new branch.

Cybis320:
Mark, those GStreamer messages don't seem to be associated with any issues and are just annoying. They are generated during normal shutdown. I haven't really looked into how to turn that off. We could reduce the Gst debug level even further - not sure what we would lose.

Regarding FPS messages, it's on my list of things to do. It’s way too sensitive right now and can absolutely be moved to debug. The average FPS is available other places anyway. While we are there, I would even completely remove 'Config file fps appears to be inaccurate. Consider updating the config file!' as it's not a true statement. 'Last calculated FPS: 24.981179 at frame 232192, config FPS: 25.0, resets: 4, startup status: False' should be debug level.

I appreciate keeping frame age and dropped frame info level even if you think I'm nuts for wanting this ;)

I'm sorry I don't have time to work on making them more self-explanatory or universally relevant at this time (my family is treating me enough as a stranger as it is right now).

markmac99:
ps - logging added to Flux.py, but not sure what you'd like to have logged @dvida  ?  Just two statements being logged right now! 
```
2025/06/18 11:40:28-INFO-Reprocess-line:347 - Preparing flux files...                                                   2025/06/18 11:40:28-INFO-Flux-line:1851 - Loaded collection areas from: flux_col_areas_UK0006_sp-20_htmin-60.0_htmax-130.0_dht-2.0_elemin-20.0.json
2025/06/18 11:40:28-INFO-Flux-line:1873 - Raw collection area at height of 100 km: 47555.84 km^2                        2025/06/18 11:40:28-INFO-Flux-line:3677 - Detecting clouds...
2025/06/18 11:40:28-INFO-Reprocess-line:364 - Plotting field sums...                                    
```

Cybis320:
Regarding Upload manager, most in there should be debug. I think we should keep a single info level just for awareness. In general, I think it's poor form to have no visible activity in the console while RMS is actually doing work, but the following are definitely debug stuff:
```
2025/06/17 19:57:58-INFO-UploadManager-line:148 - Paramiko version: 3.5.0
2025/06/17 19:57:58-INFO-UploadManager-line:149 - Establishing SSH connection to: gmn.uwo.ca:22...
2025/06/17 19:57:59-INFO-UploadManager-line:167 - SSHClient connected successfully (key file).
2025/06/17 19:57:59-INFO-UploadManager-line:172 - Keepalive set to 30 seconds
2025/06/17 19:58:00-INFO-UploadManager-line:229 - SFTP connection established.
```
Maybe keeping 'SFTP connection established' INFO is sufficient to maintain situational awareness which I think is an important concept to preserve at INFO level.

markmac99:
@Cybis320 thanks for the info.

markmac99:
Here's last night's log on UK0006. Note that i deleted the ftpdetect files from the previous night's CapturedFiles folder and then rebooted at about 1400 to force RMS to reprocess that data so i could get a view of how much logging it created. 
Actual data capture started at 2115 at line 13313 out of 35333.  
Aside: the GST pipeline failed twice during the night, though it recovered. Not sure whats causing that 

[log_UK0006_20250618_135806.log](https://github.com/user-attachments/files/20818632/log_UK0006_20250618_135806.log)

PR #661, opened 2025-06-19
Calculate fits count from ephemeris

g7gpr:
Use the ephemeris computations to work out how many fits files should be in a captured directory. This allows late starts and early finishes to be detected


g7gpr:
This needs a lot of testing.

PR #663, opened 2025-06-20
Fix next_rise error

Cybis320:
A copy-paste error in a recent change inadvertently replaced next_rise with next_set. This PR corrects the error.


dvida:
Sounds we merge this directly into master if it's causing issues?

g7gpr:
Could this be why stations have run in night mode all day? Not git blaming anyone, but trying to avoid chasing a bug that has been fixed.

Cybis320:
Yes, this caused prerelease to not switch back to day.

Cybis320:
Should have affected only prerelease.

g7gpr:
Apologies - yes, prerelease.

PR #664, opened 2025-06-24
Fix missing SITE_DIR bug in scripts not calling the logger

dvida:
RMS.Astrometry.ApplyRecalibrate didn't run as a standalone script as it failed to find the SITE_DIR global variable. I completely removed the use of globals, but left them initialized in case they're used outside of RMS.


dvida:
@Cybis320 Could you test this PR just to check that it's not immediately crashing? Otherwise, it should be good to go.

Cybis320:
No immediate issue, seems to run fine so far.

PR #671, opened 2025-06-30
Avoid crash when no fits in CapturedFiles

g7gpr:
If capture fails and there are no fits files, then RMS crashes.

This PR addresses issue #672


g7gpr:
Closing this, too many things going on at once.

PR #679, opened 2025-07-16
Fix UploadManager timing synchronization between parent and child processes

Cybis320:
Fixed a multiprocessing bug where timing variables (`last_runtime` and `next_runtime`) were not shared between parent and child processes, causing upload delays to be ignored and preventing proper coordination during shutdown.

  ## Problem
  - UploadManager inherits from `multiprocessing.Process`
  - When `upload_manager.start()` is called, it creates a new process with separate memory
  - Parent process sets timing via `addFiles()` and `delayNextUpload()`
  - Child process runs the upload loop but couldn't see parent's timing changes
  - This caused:
    - Upload delays to be ignored
    - Files not uploading before program exit
    - Coordination issues during shutdown

  ## Changes
  - Changed timing variables to use `multiprocessing.Value('d', 0.0)` for shared memory
  - Used raw Unix timestamps throughout (via `time.time()`)
  - Removed unused imports

### DIFF
--- a/RMS/UploadManager.py
+++ b/RMS/UploadManager.py
@@ -7,10 +7,7 @@ import os
 import ctypes
 import multiprocessing
 import time
-import datetime
 import logging
-
-import binascii
 import paramiko
 
 # Suppress Paramiko internal errors before they appear in logs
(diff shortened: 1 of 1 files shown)


dvida:
This is also another one I'd like to merge into prerelease sooner than later. Should it be tested better before merging or is it good to go?

Cybis320:
It should be tested on a non-continuous capture station first. (Runs fine in continuous mode).

Cybis320:
@dvida testing complete on my end. Good to go.

dvida:
Awesome. I'd you could fix the merge conflicts, I'll merge it.

Cybis320:
> Awesome. I'd you could fix the merge conflicts, I'll merge it.

done

PR #680, opened 2025-07-16
Improve logging file name handling

dvida:
Enhance the logging mechanism by updating the file naming convention to include timestamps that reflect the start time of each logging period, instead of RMS start time.

This change prevents renaming of old files during rollover and introduces a more structured log file naming approach. 

Additionally, adjust rollover intervals based on whether RMS reboots or not, to prevent fractional logs.

### DIFF
--- a/RMS/Logger.py
+++ b/RMS/Logger.py
@@ -276,37 +276,82 @@ def gstDebugLogger(category, level, file, function, line, obj, message, user_dat
 ##############################################################################
 
 class CustomHandler(logging.handlers.TimedRotatingFileHandler):
-    """ Custom handler for rotating log files.
-    
-    The live file: log_XX0000_20241229_112347.log
-    On rollover: log_XX0000_20241229_112347-[29_1123-to-30_1123].log
     """
-    def __init__(self, station_id, start_time_str, *args, **kwargs):
+    Custom handler for rotating log files where the new file's name
+    reflects the start time of the new logging period, without renaming old files.
+    
+    On rollover, it closes the current log file and creates a new one with a
+    filename timestamped to the beginning of the new logging interval.
+    """
+    def __init__(self, station_id, log_file_prefix, filename, when='H', interval=24, utc=True, **kwargs):
+        """
+        Initializes the handler.
+        
+        Args:
+            station_id (str): The station ID to embed in the filename.
+            log_file_prefix (str): A prefix for the log filename.
+            filename (str): The initial full path to the log file.
+            when, interval, utc, **kwargs: Standard TimedRotatingFileHandler arguments.
+        """
         self.station_id = station_id
-        self.start_time_str = start_time_str
-        super(CustomHandler, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


Cybis320:
Looks perfect. Will test.
Now testing on a MCL with no reboot.

Cybis320:
Should we remove the dead code? (TimedRotatingFileHandler, TimedRotatingFileHandler suffix generation). And, since we're overriding the core rollover functionality, should we inherit from just FileHandler instead of TimedRotatingFileHandle?

Regarding the 25 hr, what is the longest expected delta between the shortest and longest consecutive run+reprocess? Say, a cloudy night with no star followed by a clear night with lots of meteors on the slowest device at high latitude in the winter. Should we increase it to say 30hr?

dvida:
We're still using the internal self.atRotate flag, so that part should be left as is.
I can't find much about the suffix generation, what do you mean exactly?

nzstevem:
A few comments in the mix for consideration ...

I can see that originally the design relied on the standard reboot at the end of each capture session. This event naturally created a new log that nicely bookended a complete session. But the event monitor, linux systems that hardly ever reboot, and now continuous capture has undermined that strategy. The reboot is no longer the appropriate event to trigger a log rollover. 

1. I'm understanding log_yyyymmdd_hhmmss_tttttt.log is the name format - representing the log start time in utc - great!
2. I assume on log start it picks up the most recent log and continues appending to it
3. Can I suggest when the logger starts and it does 2. above that it puts an entry in the log to say it has done this - say "logging resumed at timestamp". This would then be inserted after a reboot, and the message text could usefully include the linux systems uptime at that moment.
4. I don't understand how you plan to coordinate the roll over with meteor capture start - a method I have seen used before in these situations  - not elegant but pragmatic - is for the logger to recognize a particular message it is being asked to log and use this to trigger the new log file - in this case it would be the start capture message. 
5. Normally the log would roll daily when it spots the start capture message, but to cater for the situation when this doesn't happen implement a max log duration check in the logger. Default to say 48 hours, but pick up the actual value from .config.
6. When a new log file is created suggest the logger write a header line, in standard format (timestamp - type - text), with some context information - roll over reason (start capture or max duration), perhaps the software version, camera id, system uptime ...??

In the end my main concern is 1. above.   Nice.

dvida:
> I assume on log start it picks up the most recent log and continues appending to it

Nope, it always open a new one, this way we can keep track of runs.

> Can I suggest when the logger starts and it does 2. above that it puts an entry in the log to say it has done this - say "logging resumed at timestamp". This would then be inserted after a reboot, and the message text could usefully include the linux systems uptime at that moment.

Great idea, we're planning to do just that.

> I don't understand how you plan to coordinate the roll over with meteor capture start - a method I have seen used before in these situations - not elegant but pragmatic - is for the logger to recognize a particular message it is being asked to log and use this to trigger the new log file - in this case it would be the start capture message.

For now, we'll just have it roll over after 24 hours to have smaller logs, but we can add more functionality later.

PR #684, opened 2025-07-20
Safeguard releaseResources and add diagnostics

Cybis320:
Fix hang on backend mismatch.  Cleanup now checks the actual object (hasattr(device, "release") vs set_state) instead of trusting the video_device_type flag, so a GstAppSink is never routed through cap.release().
- Timeout-protected OpenCV release: cap.release() runs in a daemon thread with a 2 s join; if it stalls we drop the FD and log a warning rather than block forever.
- Unified teardown: always sets device to None and handles Gst to NULL, making multiple calls to releaseResources() idempotent.
- Extra DEBUG logs: added around state transitions, cleanup branches, and timeout paths to pinpoint future stalls.

Python has no safe built-in way to abort a running C call in another thread, so there is a bit of trial and error on my part to see what works and what doesn't. The number of debug log messages may seem overkill, but until we have a good handle on this, I think it's necessary.

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1327,45 +1327,99 @@ class BufferedCapture(Process):
 
         return False
 
+
     def releaseResources(self):
         """Tear everything down in the right order so no FD survives."""
+        
+        log.debug("releaseResources: Starting")
+
+        def _timedCall(fn, timeout_s=2):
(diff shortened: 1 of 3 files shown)


Cybis320:
Battle tested on NZ cams. Good to merge.

PR #687, opened 2025-07-22
Restore initLogging function for external script compatibility

Cybis320:
A recent update refactored the logging system to use a LoggingManager class, removing the standalone initLogging function. However, external scripts that integrate with RMS may still depend on calling initLogging() directly.

This PR reintroduces initLogging as a backward-compatibility wrapper.


Cybis320:
@dvida, I haven't test this as I don't run external scripts.

Cybis320:
Tested with external script. No issues noted.

PR #689, opened 2025-07-24
Update RunExternalScript fix to support multicam

markmac99:
RMS.RunExternalScript, which can be used standalone to rerun the external script for historic or current data, currently doesnt support multicam because it doesn't accept the -c argument. This PR fixes it.


markmac99:
This is similar to the fix proposed by @adalava  in 2022, but works from the current codebase and preserves the current behaviour if the -c param is not supplied.

Cybis320 (approved):
Great!

PR #713, opened 2025-08-17
Add a bz2 file with some .config, platepars and mask file

g7gpr:
Add a bz2 file containing some sample station configuration information so that data is available for testing.

These stations are located at Perth Observatory, which is included in Stellarium and the NASA Horizons database, so the data is publicly available.

The bz2 file is arranged in the same structure as used by the mult-cam linux project. 

Whilst putting lots of data in git repos can be fraught, this is 116kB, so I feel this is safe.


dvida:
This is great - could you just add a little readme file next to the archive so one can quickly understand what's going on here.

g7gpr:
Yes, good thought. Should I add a fits file for each station as well? That's the kind of thing that can come in handy for testing purposes.

dvida:
Yes please, great idea.

g7gpr:
Done

PR #716, opened 2025-08-21
Update ConfigReader.py to fix undefined variable

markmac99:
Move definition of log_level_mapping outside the if statement so that its available for use in the next if statement. Otherwise if console_log_level is not in the config file but log_file_log_level is, then the mapping is an undefined symbol.


dvida:
I merged the master into prerelease, is this still needed? I hope I did it the right way.
The PR is showing a conflict.

markmac99:
This fix seems to be in both master and prerelease now. I can't see the conflict so i cant comment on it i'm afraid.

dvida:
Sounds good, should I close the PR?

markmac99:
Closed as the changes are now in master and prerelease.

PR #718, opened 2025-08-27
add improved ML filter model

9helix:


RMS/ConfigReader.py
```diff
@@ -602,10 +602,10 @@ def __init__(self):
         self.min_patch_intensity_multiplier = 0.0
 
         # Filtering by machine learning
-        self.ml_filter = 0.85
+        self.ml_filter = 0.5
```
dvida, line 605: The issue here is that all config files have a value of 0.85, so this default number will be ignored. I recommend changing the way the ml_filter number is read from the config file.
Scale it so that 0.85 = 0.5. In essence, read the number from the config file and then just apply y = x*0.5/0.85.

9helix: Could we do this also in ConfigReader?
```
    if parser.has_option(section, "ml_filter"):
        config.ml_filter = parser.getfloat(section, "ml_filter") * 0.5/0.85
```

dvida: That's exactly what I had in mind!

PR #734, opened 2025-10-20
Make sure shebang is first line

markmac99:
If the shebang isn't teh first line then the kernel will run a .sh file with sh rather than the intended bash. This causes many errors as the script is not sh-compatible.

### DIFF
--- a/Scripts/MultiCamLinux/add_GStation.sh
+++ b/Scripts/MultiCamLinux/add_GStation.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # This software is part of the Linux port of RMS
 # Copyright (C) 2023  Ed Harman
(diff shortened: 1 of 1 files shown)


markmac99:
This is currently causing errors when migrating to pi5, because without the correctly-placed shebang, teh pi5 kernel runs the script using sh, not bash.
Bug reported by an end user.
========================================
/home/rms/source/RMS/Scripts/MultiCamLinux/Migrate.sh: 33: [[: not found
/home/rms/source/RMS/Scripts/MultiCamLinux/Migrate.sh: 37: cd: can't cd to migrate
/home/rms/source/RMS/Scripts/MultiCamLinux/Migrate.sh: 40: [[: not found

PR #735, opened 2025-10-28
Fix: Reset stream counters and backend on mode transitions

Cybis320:
When running capture continuously without rebooting, the dropped frames counter is not being reset. After a day or two, this causes capture to switch and remain in cv2 mode until reboot.
   - Resets dropped frames counter and reset_count at day/night transitions to provide accurate per-session metrics
   - Resets media_backend_override at mode transitions to allow GStreamer retry
   - Fixes infinite reconnection loop by properly updating last_daytime_mode before breaking


dvida:
Awesome - is this solid and tested? Good for merging?

PR #737, opened 2025-10-30
Update device URL in configuration

markmac99:
Correct default ip address in config file
This should not actually be an issue as the address baked-into the current image is correct and any self-builder would need to check it anyway. However if RMS_update crashes then this incorrect version can get left behind and sometimes stashed by mistake. 
This was reported by Chuck, who noted that several of his managed cameras had the wrong address.  However, oddly, it seems they had the correct camera ID and other parameters. Does RMS now have some code to attempt to patch teh config file with missing or "incorrect" values? That would be high risk.

### DIFF
--- a/.config
+++ b/.config
@@ -93,7 +93,7 @@ reboot_lock_file: .reboot_lock
 [Capture]
 
 ; device URL
-device: rtsp://192.168.1.25:554/user=admin&password=&channel=1&stream=0.sdp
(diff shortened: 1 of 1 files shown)


dvida:
Agreed. This used to be the default IP but I lost track of when it changed. The config resets should go away once prerelease is rolled out.

Issue #742, opened 2025-11-04
FRbinViewer fails to create mp4s if bin file dimensions are not even


markmac99:
Ventsi Bo reports that FRbinViewer fails to create an mp4 from a bin file with the following error:
``` bash
[libx264 @ 0x1b38290] width not divisible by 2 (593x593)
```
This is an ffmpeg error, and not strictly RMS.  

I've checked this on Windows 11 with ffmpeg 6.1.1.16 and it does NOT exhibit the problem. However, both Ubuntu 24.04.3 and Raspian Buster do exhibit the problem. My ubuntu is running ffmpeg 6.1.1.3, buster is version 4.x.  There are newer versions of ffmpeg but they're not yet available from the Ubuntu repositories and have many additional dependencies on newer versions of libavcodec and so forth. 

The problem can be worked around by making sure the corresponding FF file is present in the folder containing the FR file. This forces ffmpeg to resize the video to 1280x720 which is a multiple of two. 

So one fix would be to check for the existence of the FF file in the CapturedFiles dated folder as well as the current working directory. 

As we do not pass dimensions into ffmpeg this can only be fully fixed by amending the code that creates the region-of-interest bin files so that it rounds the dimensions to a multiple of two.  

I'll implement and test both options, and i'll also check if we can somehow force ffmpeg to round the dimensions.

AlexanderAitov:
Hi @markmac99 , 
I can reproduce it on Mac OS with ffmpeg 7.1.1, probably it is codec restriction.
As one of possible solution (for video without FF) we can just use commented code which uses image size from config:
`#background = np.zeros((config.height, config.width), np.uint8) ` and do not calculate anything

markmac99:
> As one of possible solution (for video without FF) we can just use commented code which uses image size from config: `#background = np.zeros((config.height, config.width), np.uint8) ` and do not calculate anything

Thanks for the tip. That would work but it'd change the behaviour as all videos would come out full size instead of RoI sized and i am not keen on introducing behaviour-altering changes since they could impact other users of RMS in unexpected ways. 

However maybe i can interrogate the .bin file to determine its dimensions and apply a correctly-sized background.

markmac99:
@AlexanderAitov  your tip was in the right direction - all i needed to do was to check img_size a few lines later and make sure it was even. Works on Ubuntu, Buster and Windows. 
``` bash
        # Make the image square
        img_size = max(y_size, x_size)

[...]

PR #743, opened 2025-11-05
Fix for issue #742 in master

markmac99:
Adjust image size to be divisible by two to ensure compatability with various platforms. 
FFMPEG uses a codec that on linux and macos requires the dimensions to be even.
Thanks to Alex Aitov for pointing me in the right direction for a quick fix. 

This PR applies the fix to Master. there's a second PR to apply it in prerelease as the codebases are significantly different


markmac99:
closes #742

markmac99:
As we're close to merging prerelease i will close this.

PR #744, opened 2025-11-05
Fix for issue #742 in FRbinViewer in prerelease

markmac99:
Fix for issue reported by Ventsi Bo where FRbinViewer fails to create mp4s on Linux platforms, if the dimensions of the RoI bin file are not even.  
This PR pulls the fix into Prerelease. There is a separate PR for the same fix in Master, as the two codebases (master and prerelease) are significantly different.

### DIFF
--- a/Utils/FRbinViewer.py
+++ b/Utils/FRbinViewer.py
@@ -308,6 +308,7 @@ def view(dir_path, ff_path, fr_path, config, save_frames=False, extract_format=N
         # ffmpeg requires dimensions to be divisible by two on some platforms see issue 742
         if img_size % 2: 
             img_size +=1
+            
         background = np.zeros((img_size, img_size), np.uint8)
         add_timestamp = False
         add_shower_name = False


dvida:
Thanks! Has this been tested?
I'll also just merge this one, as I'll be merging prerelease into master soon.

markmac99:
Its been tested on UK001L which is running prerelease on a Buster 32 Pi4 build.  I tested the same fix on Ubuntu 24 running master. 

As noted in anotehr thread i'm still concerned that logging isn't working properly in Prerelease.  Notably, the logger is taking control of all logging from post-processing scripts, even those that create their own logger. This is resulting in the UKMON logs being appended to the RMS logfile, and my custom post-proc script's logs being empty. This bug appeared a while back in Master but got fixed and has reappeared. Sorry to bring it up again but i don't think prerelease is ready and it will break some of UKMONs processes if the logging isn't present.

PR #747, opened 2025-11-09
Update MultiCamLinux scripts

Cybis320:
Update GRMSUpdater.sh and StartCapture.sh MultiCamLinux scripts to support Ubuntu 24.04.

New Arguments:

  -`--term <terminal> `- Specify preferred terminal emulator
  - Options: lxterminal (default), gnome-terminal, tmux
  - Example: --term gnome-terminal

  -`--force` - Force update/restart even if RMS is already up to date
  - Without this flag, script checks if updates are needed and exits early if not (by default, no StartCapture restart if no update)

### DIFF
--- a/Scripts/MultiCamLinux/GRMSUpdater.sh
+++ b/Scripts/MultiCamLinux/GRMSUpdater.sh
@@ -16,6 +16,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+# Version 2.0 major refactoring: added early update check, graceful shutdown, strict error handling, 
+# user-mode execution, comprehensive terminal support, and regex helper function
+#
 # Version 1.9 changed parsing of the current username and display number to handle various display managers
 #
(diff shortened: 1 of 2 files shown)


dvida:
I see a lot of changes, has it been tested and should we get someone to test it?

peschman:
I am testing with success from alpha1 branch on two systems:
Ubuntu v24: US0001, 02, 0E
Ubuntu v20: US0005, 0M

markmac99:
@Cybis320  will these scripts work in "headless" mode when there is no X display available?

I start RMS from cron, as my Linux box is headless, so i have the following in the crontab:
``` bash
@reboot sleep 20 && $HOME/source/RMS/Scripts/MultiCamLinux/StartCapture.sh UK0006 2> $HOME/RMS_data/UK0006/logs/console.log
@reboot sleep 20 && $HOME/source/RMS/Scripts/MultiCamLinux/StartCapture.sh UK000F 2> $HOME/RMS_data/UK000F/logs/console.log
```
GRMSUpdater seems a bit complex -  Is it really needed? I just run RMS_Update.sh at 1500 local time on a Sunday, and then immediately reboot the server.

PR #748, opened 2025-11-12
use a unique logger name in logging, and also add -e parameter to Reprocess

markmac99:
When creating loggers, RMS was using a generic name "logger". This generic name, common in example code, was likely to clash with 3rd party and end-user-developed code which might also use the same generic name. This could cause RMS to "steal" messages from other modules or indeed to suppress them due to the filter that RMS puts on its log streams.   This arises because calls to logger.getLogger("logger") in 3rd party code would connect to the RMS logger, if it existed. 

By using a unique name "rmslogger" we avoid the risk of 3rd party logs being incorrectly either captured or filtered out and lost. 

Tested on UK001L which runs two post-processing scripts one after the other. Before the change, the ukmon postproc scripts' logs were partly-captured into RMS's log and partly lost, while the subsequent site-specific script's logs were entirely lost.
After the change, both downstream scripts logged correctly to their own log files (though RMS still captured some messages from the ukmon script)

Note: UK001L is a single-station Pi4 but it is configured in the multicam layout.

I also sneaked in a small change to RMS.Reprocess to allow the passing of a parameter -e. When this is set, the external script will be run after reprocessing. At present, Reprocess does not run the external scripts so they have to be run manually. Default behaviour is unchanged, but this would allow me to reprocess and automatically re-upload to ukmon etc.

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -880,7 +880,7 @@ def parseDFNStation(config, parser):
     config.catalog_mag_limit = 4.5
 
     config.star_catalog_path = 'Catalogs'
-    config.star_catalog_file = 'BSC5'
+    config.star_catalog_file = 'GMN_StarCatalog'
     config.platepar_name = 'platepar_cmn2010.cal'
     config.deinterlace_order = -2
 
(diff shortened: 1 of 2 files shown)


dvida:
Great job, thank you! Let's test it in the field for a few days. We'll have to wait until the Leonids are done, they seem to be showing some unusual activity!

PR #750, opened 2025-11-17
Fix timelapse generation for night-only and polar stations

Cybis320:
Fixes two issues with timelapse generation that affected night-only cameras and polar stations:

  1. **Gap detection for night-only cameras**: Splits recording sessions when there are time gaps > 2 hours between consecutive frames
  2. **Polar station cutoff fix**: Uses current time instead of midnight UTC as the processing cutoff for polar regions


Cybis320:
@dvida, this seems to be working well now. This can be merged IMO.

PR #751, opened 2025-11-25
fix: Improve platepar loading to prevent the default hardcoded platep…

dvida:
Sometimes the platepar does not get properly loaded during the recalibration procedure. This improvement passes a platepar to the recalibration function instead of the recalibration function having to find it on its own.


Cybis320:
Yep, I think that addresses the root cause.

PR #756, opened 2025-12-01
Fix infinite auto-reprocessing loop and QueuedPool memory leak

Cybis320:
On StartCapture launch, if any CapturedFiles directory contains a meteor-less night (empty FTPdetectinfo file), auto-reprocess (if enabled) would reprocess these folders repeatedly on every  startup until DeleteOldObservations eventually cycles them out.

This happens because the FTPdetectinfo file for those nights is last touched before the platepar is saved. Since one of the auto-reprocess triggers compares the mtime of these two files (to detect manual platepar replacement), the platepar always appears newer, triggering unnecessary reprocessing.

Fix: Touch the FTPdetectinfo file after recalibration completes, ensuring its timestamp is updated even when no meteors are detected.

Additionally, after auto-reprocessing completes, StartCapture would not fully release QueuedPool memory until the end of the subsequent normal reprocess. The multiprocessing Manager and its queues would persist, holding references to cached results.

Fix: Add a cleanup() method to QueuedPool that explicitly shuts down the Manager and clears queues. Call it in Reprocess after detection is complete.

### DIFF
--- a/RMS/ExtractStars.py
+++ b/RMS/ExtractStars.py
@@ -873,6 +873,9 @@ def extractStarsAndSave(config, ff_dir):
     # Delete QueuedPool backed up files
     if workpool is not None:
         workpool.deleteBackupFiles()
+        # Ensure manager/queues are torn down when using QueuedPool
+        if hasattr(workpool, "cleanup"):
+            workpool.cleanup()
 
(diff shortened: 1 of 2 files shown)


Cybis320:
Closing. Will remove the time check instead.

PR #761, opened 2025-12-02
Add pandas to requirements.txt

Cybis320:
Utils/KalmanFilter.py requires pandas

### DIFF
--- a/Utils/Astra.py
+++ b/Utils/Astra.py
@@ -26,8 +26,10 @@ from RMS.Routines import Image
 
 try:
     from pyswarms.single.global_best import GlobalBestPSO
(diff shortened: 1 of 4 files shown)


Cybis320:
Actually, made pandas and pyswarms optional and update messages communicating their use.

markmac99 (commented):
Suggest something is also added to the README file to explain what ASTRA is and what the Kalman Filter module is for. These sections could then also explain about the two additional requirements.

That said, pandas is generally useful and all my RMS environments have it installed anyway !

dvida:
removed pandas from requirements in PR #766

PR #762, opened 2025-12-02
Replace FFT with a Nearest Neighbor Ransac approach and refactor SkyFit

Cybis320:
Replace deprecated imreg_dft with scikit-image.


RMS/Astrometry/FFTalign.py
```diff
@@ -141,38 +232,101 @@ def findStarsTransform(config, reference_list, moved_list, img_size=256, dot_rad
...
+            log.warning('FFT registration error: The scale correction is too high ({:.3f})!'.format(scale))
+            return 0.0, 1.0, 0.0, 0.0
+
+        # Apply rotation correction to moved image to find translation
+        # Use skimage.transform.rotate which handles center rotation properly
+        img_mov_corrected = rotate(img_mov, angle, resize=False, mode='constant', cval=0, order=1)
+
+        # Handle scale correction if significant
+        if abs(scale - 1.0) > 0.001:
+            from scipy.ndimage import zoom
```
markmac99: Recommend moving the import to the top of the file with other imports, to follow PEP-08 best practice and prevent confusion / multiple inclusion of the same module.

RMS/Astrometry/FFTalign.py
```diff
@@ -331,8 +485,8 @@ def alignPlatepar(config, platepar, calstars_time, calstars_coords, scale_update
 
     # Compute the new reference RA and Dec
     _, ra_centre_new, dec_centre_new, _ = ApplyAstrometry.xyToRaDecPP([jd2Date(platepar_aligned.JD)], \
-        [platepar_aligned.X_res/2 - 2*translation_x], \
-        [platepar_aligned.Y_res/2 - 2*translation_y], [1], platepar_aligned, \
+        [platepar_aligned.X_res/2 + translation_x], \
```
markmac99: The original code had  "+ 2*translation_x" - is removal of the 2 deliberate? I don't know the code well enough to be sure.

markmac99 (commented):
I'll have to assume the maths is correct as its not an area i'm familiar with. I left a couple of comments in the review.

RMS/Astrometry/FFTalign.py
Cybis320: It is intentional. I think that was an error in the original code, but I'm not yet 100% sure. It's still very much a work in progress.
Something is still not quite right.

Cybis320: the removal of the '2' was deliberate, but not the change of sign... fixed. Now things line up fine.

Cybis320:
Ok, I think it works. It's slightly more accurate (although this a coarse tool so it doesn't matter) and it's slightly faster.

dvida:
Would could also just use RANSAC to reduce complexity: https://scikit-image.org/docs/0.25.x/auto_examples/transform/plot_matching.html
Apparently there are also USAC and MAGSAC that's included in opencv (>v4.0) which are apparently better, but I never used them myself.

Cybis320:
> Would could also just use RANSAC to reduce complexity: https://scikit-image.org/docs/0.25.x/auto_examples/transform/plot_matching.html Apparently there are also USAC and MAGSAC that's included in opencv (>v4.0) which are apparently better, but I never used them myself.

Great suggestion!! It's so much better. 185x faster, perfectly accurate, and 180 less lines of code.

markmac99:
@Cybis320  I commented this on the latest commit, but ImageIO version 2.33 isn't available for Python 3.7 which is the standard on Buster builds. The last version thats compatible with python 3.7 is 2.31.2.  See https://pypi.org/project/ImageIO/#history for more. 
```
  Could not find a version that satisfies the requirement imageio>=2.33 (from -r requirements.txt (line 16)) (from versions: 1.1-linux32, 1.1-linux64, 1.1-osx64, 1.1-win32, 1.1-win64, 1.2-linux32, 1.2-linux64, 1.2-osx64, 1.2-win32, 1.2-win64, 1.3-linux32, 1.3-linux64, 1.3-osx64, 1.3-win32, 1.3-win64, 0.2.1, 0.2.2, 0.2.3, 0.3.0, 0.3.1, 0.3.2, 0.4, 0.4.1, 0.5, 0.5.1, 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.1.2, 2.2.0, 2.3.0, 2.4.0, 2.4.1, 2.5.0, 2.6.0, 2.6.1, 2.8.0, 2.9.0, 2.10.1, 2.10.2, 2.10.3, 2.10.4, 2.10.5, 2.11.0, 2.11.1, 2.12.0, 2.13.0, 2.13.1, 2.13.2, 2.13.3, 2.13.4, 2.13.5, 2.14.0, 2.14.1, 2.15.0, 2.16.0, 2.16.1, 2.16.2, 2.17.0, 2.18.0, 2.19.0, 2.19.1, 2.19.2, 2.19.3, 2.19.5, 2.20.0, 2.21.0, 2.21.1, 2.21.2, 2.21.3, 2.22.0, 2.22.1, 2.22.2, 2.22.3, 2.22.4, 2.23.0, 2.24.0, 2.25.0, 2.25.1, 2.26.0, 2.26.1, 2.27.0, 2.28.0, 2.28.1, 2.29.0, 2.30.0, 2.31.0, 2.31.1, 2.31.2)
No matching distribution found for imageio>=2.33 (from -r requirements.txt (line 16))
(testsk) pi@uk001l:~/source/testing/testsk $ python -V
Python 3.7.3
(testsk) pi@uk001l:~/source/testing/testsk $ cat /etc/os-release
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
```

markmac99:
Latest update resolves the imageio issue, however frustratingly we have a new problem on Python 3.7 - the latest build of bcrypt is only available as a tar.gz file, and to build it requires Rust which is not installed / available for Buster 32bit. 
```
Building wheels for collected packages: bcrypt
  Building wheel for bcrypt (pyproject.toml) ... error
  error: subprocess-exited-with-error
 This package requires Rust >=1.64.0.
```
To work around that i think we need to install the last-good wheel version on python 3.7
```
bcrypt==3.2.0 ; python_version <='3.7'
bcrypt ; python_version >='3.8' 
```
in the requirements.  Sigh.

Cybis320:
Isn't that a paramiko dependency? We didn't touch that here or in prerelease. Are you running into issues with a new install on Buster?

markmac99:
Quite possibly. I was testing by trying to install the requirements on Buster in a brand new virtualenv, so that i can identify any potential incompatabilities. Existing stations might be ok, as they'd have bcrypt 3.2.0 installed already. So, it should only be a problem if someone has a buster-based system and has to rebuild it to buster.  Maybe we just make a note in the README file explaining the potential issue.

Cybis320:
I was trying a simple sudo apt update on buster and got some sort of end-of-life message. I didn’t investigate further. On Dec 11, 2025, at 3:21 PM, Mark McIntyre ***@***.***> wrote:﻿markmac99 left a comment (CroatianMeteorNetwork/RMS#762)
Quite possibly. I was testing by trying to install the requirements on Buster in a brand new virtualenv, so that i can identify any potential incompatabilities. Existing stations might be ok, as they'd have bcrypt 3.2.0 installed already. So, it should only be a problem if someone has a buster-based system and has to rebuild it to buster.  Maybe we just make a note in the README file explaining the potential issue.

Cybis320:
It turns out FFTalign was making the pointing worse in every case. Within ~0.9° of tolerance, the 10 px latching logic would snap the pointing back to where it belongs, but outside that window, FFT would amplify the error.

I replaced it with a nearest-neighbor approach using RANSAC-like outlier rejection (NNalign). This method can recover pointing that has drifted up to 2–3°. It can also reliably creates match pairs between detected and catalog stars. Combined with the accurate initial pointing from astrometry.net, this means fully automated platepar creation is now possible.

No additional packages required (no imreg_dft, scikit-image, or anything else.)

See demo:
https://youtu.be/q-ODjRWmUpI

markmac99:
> Combined with the accurate initial pointing from astrometry.net, this means fully automated platepar creation is now possible.

Sounds interesting. I have never been able to get the astrometry.net process to platesolve an RMS image though!

dvida:
The astrometry.net online service is hit or miss, but the local install of the astrometry package (linux only) works great.

RMS/Formats/Platepar.py
```diff
@@ -501,6 +510,149 @@ def _calcSkyResidualsAstro(params, platepar, jd, catalog_stars, img_stars):
...
+            jd: [float] Julian date of the image.
+            img_stars: [ndarray] Detected stars as Nx3 array (x, y, intensity).
+            catalog_stars: [ndarray] Catalog stars in FOV as Mx3 array (ra, dec, mag).
+
+        Keyword arguments:
+            fixed_scale: [bool] Keep scale fixed. True by default (for camera drift correction).
+
+        """
+
+        from RMS.Astrometry.ApplyAstrometry import raDecToXYPP
```
dvida, line 573: all imports should be at the top of the script

RMS/Formats/Platepar.py
```diff
@@ -501,6 +510,149 @@ def _calcSkyResidualsAstro(params, platepar, jd, catalog_stars, img_stars):
...
+        # update these to match the observation JD, changing RA_d won't have the expected
+        # effect on the catalog star projections.
+        #
+        # When we change JD/Ho, we must also adjust RA_d to maintain the same pointing direction.
+        # The relationship is: new_RA_d = old_RA_d + delta_Ho (sidereal rotation)
+        old_Ho = pp_work.Ho
+
+        pp_work.JD = jd
+        T = (jd - 2451545.0) / 36525.0
+        new_Ho = (
```
dvida: Isn't there a way to call this from a function instead of duplicating the code?

RMS/Formats/Platepar.py
```diff
@@ -501,6 +510,149 @@ def _calcSkyResidualsAstro(params, platepar, jd, catalog_stars, img_stars):
...
+        # Update fitted parameters
+        self.RA_d, self.dec_d, self.pos_angle_ref = res.x[:3]
+        if not fixed_scale:
+            self.F_scale = abs(res.x[3])
+
+        # Update JD and Ho to match the observation time
+        # This ensures the fitted RA_d is consistent with the new reference time
+        self.JD = jd
+        T = (jd - 2451545.0) / 36525.0
+        self.Ho = (
```
dvida: same here

Utils/SkyFit2.py
```diff
@@ -6039,9 +6155,223 @@ def filterCatalogStarsInsideFOV(self, catalog_stars, remove_under_horizon=True,
...
+    def tryQuickAlignment(self):
+        """ Try to align platepar using existing pointing as starting point.
+
+        Uses alignPlatepar() to refine pointing, then checks if the fit is good enough
+        by counting matched stars. If successful, performs full NN-based astrometry fit.
+
+        Returns:
+            success: [bool] True if quick alignment succeeded and we can skip astrometry.net
+        """
+        from RMS.Formats.FFfile import convertFRNameToFF
```
dvida: move imports to the top of the scrip

markmac99 (commented):
I've found out why astrometry.net doesn't work in SkyFit2 - Imgur is not available in the UK I believe because the company were found to be hosting indecent images of minors, and decided to block UK access to try to evade penalties. However, its possible to directly submit an image to astrometry.net so maybe we don't need imgur.  Happy to investigate

dvida:
Oh shoot, it never occurred to me that it could be a problem. Please do go ahead

markmac99:
I've pushed a fix which allows for direct submission to nova.astrometry.net in the case where imgur isn't available for commercial or legal reasons. To get this working I had to update the astrometry client as the method of submitting files has changed. I also changed SkyFit2 to always use the x/y star points if available. This seems to work really well with the new NN code.

RMS/Astrometry/AstrometryNetNova.py
markmac99, line 1: I updated the astrometry.net client code here using the latest version in their github repo. This allows files to be directly submitted to nova.

RMS/Astrometry/AstrometryNetNova.py
```diff
@@ -62,7 +62,7 @@
 
 def printDebug(*args):
     if DEBUG:
-        printDebug(*args)
```
markmac99, line 65: printDebug was recursive !!! I suspect this was a search/replace error.  I fixed it in a commit to the branch

RMS/Astrometry/AstrometryNetNova.py
```diff
@@ -353,8 +326,13 @@ def _printWebLink(stat, first_status=None):
         img_data = file_handle.getvalue()
 
         # Upload the image to imgur
-        image_url = imgurUpload('skyfit_image.jpg', image_data=img_data)
-
+        try:
```
markmac99: If the imgur upload fails, which it will if imgur is not available locally or for legal reasons, then I changed the code to instead submit data directly to nova.astrometry.net

markmac99 (approved):
Can't comment much on the NN method except to say i have tested it in Skyfit with well-calibrated, badly calibrated and totally uncalibrated data, and it seems to do an excellent job .

dvida:
It's pretty awesome, yup! A few minor tweaks are needed, e.g. if the fit is good after the 5odd step, the asymmetry correction should be turned on and it should finish the fit with the odd7 model. It also needs a bit more stringent outlier rejection procedure for both astrometry and photometry. Once these are done, that's it.

dvida:
I keep testing it and I'm pleased with the astrometry performance. However, when I add a few more stars after the initial picking, the photometry is off by ~2 mags. There might be an issue with how the photometry is being done in the automated procedure.

Cybis320:
> I keep testing it and I'm pleased with the astrometry performance. However, when I add a few more stars after the initial picking, the photometry is off by ~2 mags. There might be an issue with how the photometry is being done in the automated procedure.

Good catch. There was an off-by-one error on an index which caused the photometric calibration to use the wrong field from CALSTARS. Hopefully, it's fixed now.

dvida:
Awesome, it works now! The only outstanding thing is robust outlier rejection. Here's the photometry on a test data set, there are 1-2 obvious outliers that should be automatically removed:

image

The fit stddev is 0.33, so 3 sigma outlier rejection will get rid of the worst one. I suggest doing 3 iterations of outlier rejection for the stddev to "stabilize" on a good set of data. After I manually remove the worst ones following the suggested procedure, I get a really good final fit:

image

The same holds for astrometry (the SNR plot might also need a bit of attention to fix the x-axis ticks):

image

And finally, after the initial odd3 fit, make it turn on asymmetry correction and then fit all the way to odd7. After the odd5 star pairing, it doesn't need to re-pair for odd7, it should just apply a higher order fit to take care of the small discrepancies. That will be it then.

Cybis320:
All good points.

Regarding the last one, currently the last pass is done with whatever the user selected before hitting auto-fit - do we want to ignore that and force asymmetry and odd7, or leave to the user? I'm leaning towards leave it to the user, no?

dvida:
I suggest optimizing for the default behaviour (asymmetry + odd7) and removing any assumptions about correct usage by casual users. This would make the approach genuinely one-shot. In the rare cases where an alternative model is appropriate, advanced users can explicitly specify it.

markmac99:
> and removing any assumptions about correct usage by casual users.

+1 on this. I have come across quite a few platepars where the user had left it to odd3 because that was the old default value, and they've never revisited it, or were unaware of the recommendation to use odd7.

Cybis320:
Good points. How about changing the defaults and keeping the last pass as selected? This way if the user only clicks on auto-fit, the last pass will be with sensible default values.

dvida:
For sure - how about you add a second button on the auto fit window? One will respect the set model, and the other will use the recommended values (odd7 + asymmetry)? And then highlight the "Recommended" option so the users always select that if they don't know what they're doing.

Cybis320:
What are the recommended options? This? 
Screenshot 2026-01-04 at 10 25 35 AM

dvida:
Also enable equal aspect and that should be it:
image

Modern digital sensors have an equal aspect ratio, while the older analog sensors had non-square pixels (fun).

dvida:
I just tried the latest version - holy s*** it works great. This will save us a ton of time.
The only outstanding issue seems to be that the star SNR is not being tracked, they're all set to -1.
image

Otherwise, the fit is perfect:
image

Cybis320:
> I just tried the latest version - holy s*** it works great. This will save us a ton of time. The only outstanding issue seems to be that the star SNR is not being tracked, they're all set to -1. 

Was the CALSTARS generated by the code from this branch?

dvida:
It's an old CALSTARS file with no SNR values, so I guess that's the reason. We can tolerate that, unless the SNR is crucial in the fitting process.

Cybis320:
> It's an old CALSTARS file with no SNR values, so I guess that's the reason. We can tolerate that, unless the SNR is crucial in the fitting process.

It's only used for display, so I think we are good. In another branch (test-coordinate-transforms) I have added a tab to regenerate CALSTARS with user settings - that way there is no need to edit the config file and delete the CALSTARS, one can experiment directly in SkyFit.

dvida:
Awesome. Any other outstanding things you're working on?

Cybis320:
> Awesome. Any other outstanding things you're working on?

On this particular PR, yes. I've not tested how well it handles during reprocessing. And it still needs a sanity check code.

dvida:
Got it, I will temper my excitement about this feature for now.

Cybis320:
Because local astrometry.net is not available on Windows, and the public nova.astrometry.net can be unreliable (slow, timeouts, rate limits), I've set up a dedicated astrometry.net server for GMN at astro.contrailcast.com.  
Here's the full fallback flow:
```
astrometryNetSolve() called
  │
  ├─► LOCAL AVAILABLE?
  │   │
  │   ├─► YES: Try local astrometry.net
  │   │   │
  │   │   ├─► SUCCESS → return result
  │   │   │
  │   │   └─► FAIL (exception) → continue to remote...
  │   │
  │   └─► NO: Skip to remote
  │
  ├─► REMOTE: Try PRIMARY (contrailcast)
  │   │
  │   ├─► Has coordinates?
  │   │   │
  │   │   ├─► YES: Try coordinate-only solve (XYLS upload)
  │   │   │   │
  │   │   │   ├─► SUCCESS → return result
  │   │   │   │
  │   │   │   └─► FAIL → Has image?
  │   │   │       │
  │   │   │       ├─► YES: Try image upload (direct to contrailcast)
  │   │   │       │   │
  │   │   │       │   ├─► SUCCESS → return result
  │   │   │       │   │
  │   │   │       │   └─► FAIL → continue to fallback server
  │   │   │       │
  │   │   │       └─► NO: continue to fallback server
  │   │   │
  │   │   └─► NO: Try image upload (direct)
  │   │       │
  │   │       ├─► SUCCESS → return result
  │   │       │
  │   │       └─► FAIL → continue to fallback server
  │   │
  │
  └─► REMOTE: Try FALLBACK (nova.astrometry.net)
      │
      ├─► Has coordinates?
      │   │
      │   ├─► YES: Try coordinate-only solve (XYLS upload)
      │   │   │
      │   │   ├─► SUCCESS → return result
      │   │   │
      │   │   └─► FAIL → Has image?
      │   │       │
      │   │       ├─► YES: Try image upload
      │   │       │   │
      │   │       │   ├─► Try imgur upload first
      │   │       │   │   │
      │   │       │   │   ├─► SUCCESS → nova fetches from imgur URL
      │   │       │   │   │
      │   │       │   │   └─► FAIL → direct upload to nova
      │   │       │   │
      │   │       │   ├─► SUCCESS → return result
      │   │       │   │
      │   │       │   └─► FAIL → return None
      │   │       │
      │   │       └─► NO: return None
      │   │
      │   └─► NO: Try image upload (imgur → nova)
      │       │
      │       ├─► SUCCESS → return result
      │       │
      │       └─► FAIL → return None
      │
      └─► return None (all attempts exhausted)

  Summary of upload methods by server:
  ┌──────────────┬─────────────┬────────────────────────────────┐
  │    Server    │ Coordinates │             Image              │
  ├──────────────┼─────────────┼────────────────────────────────┤
  │ Local        │ XYLS        │ Direct file                    │
  ├──────────────┼─────────────┼────────────────────────────────┤
  │ Contrailcast │ XYLS        │ Direct upload                  │
  ├──────────────┼─────────────┼────────────────────────────────┤
  │ Nova         │ XYLS        │ Imgur URL → fallback to direct │
  └──────────────┴─────────────┴────────────────────────────────┘
```
  Worst-case timing (all failures until final success):
  - Local fail: immediate
  - Contrailcast coord: ~75 sec timeout
  - Contrailcast image: ~6 sec fail
  - Nova coord: ~75 sec timeout (it's currently not reachable so I just guessed)
  - Nova image (imgur): ~6 sec (also a guess)
  - Total worst case: ~2.7 min

Cybis320:
for testing remote when local astrometry.net is available:
```
RMS_DISABLE_LOCAL_ASTROMETRY=1 python -m Utils.SkyFit2
```

RMS/Formats/Platepar.py
Cybis320: done

RMS/Formats/Platepar.py
Cybis320: fixed

g7gpr:
Humbly request that Folder/FF_Images button opens RMS_data, or RMS_data/CapturedFiles/  rather than ~/

Auto fit works brilliantly for 4mm lenses, for 6mm (au000d, au000l) and 16mm (au000k) astrometry.net fails to find a solution.

dvida:
The mag 12 version of the catalog also needs to be updated with the new fields:
```
The full catalog (LM+12.0) is beind downloaded from the GMN server... 
Downloading: 100.00% - Done!
  File "/home/dvida/source/RMS/Utils/SkyFit2.py", line 6832, in keyPressEvent
    self.autoFitAstrometryNet()
  File "/home/dvida/source/RMS/Utils/SkyFit2.py", line 8844, in autoFitAstrometryNet
    self.balanceCatalogMagnitude()
  File "/home/dvida/source/RMS/Utils/SkyFit2.py", line 5145, in balanceCatalogMagnitude
    temp_catalog = self.loadCatalogStars(current_mag_limit)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dvida/source/RMS/Utils/SkyFit2.py", line 9375, in loadCatalogStars
    catalog_results = StarCatalog.readStarCatalog(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dvida/source/RMS/RMS/Formats/StarCatalog.py", line 501, in readStarCatalog
    return loadGMNStarCatalog(
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dvida/source/RMS/RMS/Decorators.py", line 79, in __call__
    value = self.func(*args, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dvida/source/RMS/RMS/Formats/StarCatalog.py", line 365, in loadGMNStarCatalog
    raise ValueError("Unknown field(s) in additional_fields:" + ', '.join(unknown))
ValueError: Unknown field(s) in additional_fields:common_name, bayer_name
```

Cybis320:
> Humbly request that Folder/FF_Images button opens RMS_data, or RMS_data/CapturedFiles/ rather than ~/
> 
> Auto fit works brilliantly for 4mm lenses, for 6mm (au000d, au000l) and 16mm (au000k) astrometry.net fails to find a solution.

I think it may have been pointed out somewhere else, but the astrometry solver needs a rough FOV size clue to narrow down its search. It gets this clue from the config FOV parameters. Hopefully, that will solve the 16 mm lens issue.

Regarding the open folder button, it should probably open ~/RMS_data if available, which is the least common denominator between single-cam and multi-cam.

g7gpr:
From command line solver can operate without any clues. 
```
(vRMS) david@lap-deb-01:~/source/RMS$ solve-field  ~/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.bmp --overwrite
Reading input file 1 of 1: "/home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.bmp"...
bmptopnm: Windows (v1) BMP, 1280x720x8
bmptopnm: WRITING PGM IMAGE
Read file /tmp/tmp.ppm.FmSZsL: 1280 x 720 pixels x 1 color(s); maxval 255
Using 8-bit output
Extracting sources...
simplexy: found 2636 sources.
Solving...
Reading file "/home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.axy"...
Field 1 did not solve (index index-4119.fits, field objects 1-10).
Field 1 did not solve (index index-4118.fits, field objects 1-10).
Field 1 did not solve (index index-4117.fits, field objects 1-10).
Field 1 did not solve (index index-4116.fits, field objects 1-10).
Field 1 did not solve (index index-4115.fits, field objects 1-10).
Field 1 did not solve (index index-4119.fits, field objects 11-20).
Field 1 did not solve (index index-4118.fits, field objects 11-20).
Field 1 did not solve (index index-4117.fits, field objects 11-20).
Field 1 did not solve (index index-4116.fits, field objects 11-20).
  log-odds ratio 170.59 (1.21939e+74), 40 match, 0 conflict, 103 distractors, 60 index.
  RA,Dec = (59.9679,-59.5352), pixel scale 55.7195 arcsec/pix.
  Hit/miss:   Hit/miss: +++--+-+-++--+++++-+--+-+---++-++----+-++-+---+--------+-+------+++-+---------+----+-----------+-+--
Field 1: solved with index index-4115.fits.
Field 1 solved: writing to file /home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.solved to indicate this.
Field: /home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.bmp
Field center: (RA,Dec) = (59.979198, -59.530061) deg.
Field center: (RA H:M:S, Dec D:M:S) = (03:59:55.008, -59:31:48.220).
Field size: 19.6944 x 11.15 degrees
Field rotation angle: up is -114.795 degrees E of N
Field parity: neg
Creating new FITS file "/home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.new"...
Creating index object overlay plot...
Creating annotation plot...
Your field contains:
  The constellation Reticulum (Ret)
  The star κ Dor
  The star α Ret
  The star β Ret
  The star γ Ret
  The star δ Ret
  The star ε Ret
  The star η Ret
  The star θ Ret
  The star ι Ret
  The star κ Ret
  NGC 1787

```

Cybis320:
Interesting, can you compare the output with SkyFit auto fit and the wrong FOV clues in config?

> From command line solver can operate without any clues.
> 
> ```
> (vRMS) david@lap-deb-01:~/source/RMS$ solve-field  ~/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.bmp --overwrite
> Reading input file 1 of 1: "/home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.bmp"...
> bmptopnm: Windows (v1) BMP, 1280x720x8
> bmptopnm: WRITING PGM IMAGE
> Read file /tmp/tmp.ppm.FmSZsL: 1280 x 720 pixels x 1 color(s); maxval 255
> Using 8-bit output
> Extracting sources...
> simplexy: found 2636 sources.
> Solving...
> Reading file "/home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.axy"...
> Field 1 did not solve (index index-4119.fits, field objects 1-10).
> Field 1 did not solve (index index-4118.fits, field objects 1-10).
> Field 1 did not solve (index index-4117.fits, field objects 1-10).
> Field 1 did not solve (index index-4116.fits, field objects 1-10).
> Field 1 did not solve (index index-4115.fits, field objects 1-10).
> Field 1 did not solve (index index-4119.fits, field objects 11-20).
> Field 1 did not solve (index index-4118.fits, field objects 11-20).
> Field 1 did not solve (index index-4117.fits, field objects 11-20).
> Field 1 did not solve (index index-4116.fits, field objects 11-20).
>   log-odds ratio 170.59 (1.21939e+74), 40 match, 0 conflict, 103 distractors, 60 index.
>   RA,Dec = (59.9679,-59.5352), pixel scale 55.7195 arcsec/pix.
>   Hit/miss:   Hit/miss: +++--+-+-++--+++++-+--+-+---++-++----+-++-+---+--------+-+------+++-+---------+----+-----------+-+--
> Field 1: solved with index index-4115.fits.
> Field 1 solved: writing to file /home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.solved to indicate this.
> Field: /home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.bmp
> Field center: (RA,Dec) = (59.979198, -59.530061) deg.
> Field center: (RA H:M:S, Dec D:M:S) = (03:59:55.008, -59:31:48.220).
> Field size: 19.6944 x 11.15 degrees
> Field rotation angle: up is -114.795 degrees E of N
> Field parity: neg
> Creating new FITS file "/home/david/RMS_data/ArchivedFiles/AU000K_20260127_120427_648492/FF_AU000K_20260127_145833_615_0258304.new"...
> Creating index object overlay plot...
> Creating annotation plot...
> Your field contains:
>   The constellation Reticulum (Ret)
>   The star κ Dor
>   The star α Ret
>   The star β Ret
>   The star γ Ret
>   The star δ Ret
>   The star ε Ret
>   The star η Ret
>   The star θ Ret
>   The star ι Ret
>   The star κ Ret
>   NGC 1787
> ```

g7gpr:
PR #802 provides a solution for this, which seems robust. Testing an auto fit platepar on au000g.

Cybis320:
Added fallback to wide search when quick narrow atrometry.net fails and warn user config FOV is wrong.
Screenshot 2026-01-28 at 11 53 30 PM

Cybis320:
Also, we should test not just SkyFit but Reprocess as well and compare results to master.

g7gpr:
> Also, we should test not just SkyFit but Reprocess as well and compare results to master.

I'm probably well placed to do this, as I have one user account per station. Therefore I can run two different versions of the code with the same camera address. I'll set that up tonight.

g7gpr:
au000e is now pointed at the same camera as au000g. g is on this branch, e is on master.

g7gpr:
```
(vRMS) gmn@walnut:~/source/RMS$ sudo su au000g -c "cd ~/source/RMS; git log -1"
commit 1c454e7e03fb82c234b03811d8a87816fde995ea (HEAD -> feature-scikit-image, origin/feature-scikit-image)
Author: Luc Busquin <133058544+Cybis320@users.noreply.github.com>
Date:   Thu Jan 29 00:07:51 2026 +0100

    Change star name filter to magnitude instead of count
(vRMS) gmn@walnut:~/source/RMS$ sudo su au000e -c "cd ~/source/RMS; git log -1"
commit 8ec0c1eb05aa030294f8c73609c9c29860b2f204 (HEAD -> master, origin/master, origin/HEAD)
Merge: 12b44493 ab747f62
Author: Denis Vida <denis.vida@gmail.com>
Date:   Sun Jan 18 20:33:07 2026 -0500

    Merge pull request #797 from CroatianMeteorNetwork/markmac99-typo-fix
    
    Fix logger name typo in Logger.py
```
AU000E_20260129_120305_205395_calib_report_astrometry
AU000G_20260129_120305_107642_calib_report_astrometry

g7gpr:
These are all 6mm lenses, but I think the results are clear enough! 

For au000g, manually created platepar astrometry report
AU000G_20260125_120544_636814_calib_report_astrometry

Automatically created platepar astrometry report
AU000G_20260129_120305_107642_calib_report_astrometry

Manually created platepar photometry report
AU000G_20260125_120544_636814_calib_report_photometry

Automatically created platpar photometry report
AU000G_20260129_120305_107642_calib_report_photometry

dvida:
Can you try doing it on the same image? Half a mag of difference in the photometric calibration should be investigated, unless the clouds interfered with the first fit.

g7gpr:
Same commit, manual platepar vs automatic platepar?

dvida:
Right, we want to see that the automated photometry fit compares well to the manual fit on the same data. There was previously a bug that caused issues with the automated photometry, I want to confirm that everything is 100% ok now.

Cybis320:
> These are all 6mm lenses, but I think the results are clear enough!
> 
> For au000g, manually created platepar astrometry report AU000G_20260125_120544_636814_calib_report_astrometry
> 
> Automatically created platepar astrometry report AU000G_20260129_120305_107642_calib_report_astrometry
> 
> Manually created platepar photometry report AU000G_20260125_120544_636814_calib_report_photometry
> 
> Automatically created platpar photometry report AU000G_20260129_120305_107642_calib_report_photometry

It looks like the manual photometry fit also fitted the vignetting coefficient vs the auto used the fixed value, so that goes some ways into explaining the difference.

g7gpr:
I'm very time poor for the next few days, it seems like @Cybis320 is all over this. Can I leave this to @Cybis320 to look into?

Please do consider that the difference may be caused by poor platepar creation by the operator.

dvida:
Centroiding somehow got messed up, I click on the fireball but the pick gets significantly offset. Any ideas?
image
image

Cybis320:
> Centroiding somehow got messed up, I click on the fireball but the pick gets significantly offset. Any ideas?

I'm not seeing the issue on my side. Could you post a full screenshot with the distortion params?

dvida:
The distortion params should't affect the centroids as that is only done in the x/y space. I'll try to take a look on my end.

Cybis320:
I tried to replicate on a Mac and Ubuntu with and without display scaling, in manual reduction mode or not, the centroids are always perfect. Are you measuring FF files or something else?

dvida:
On master, using an FF file in the manual reduction mode. I found a meteor segment.
I intentionally made the aperture large and clicked off the bright segment, the centroid snaps in the right place no matter where I click:
image
image
image

On the new branch:
image
image
image

Cybis320:
Right, I see the behavior you describe from master using this branch even with a wide aperture. I really want to reproduce though.

Cybis320:
And you seeing the same on regular stars or just on meteors - is it a moving object issue somehow?

dvida:
You can take any FF file with a meteor to test. I only noticed it in manual reduction, haven't tried it in stars. This was perhaps not noticed because most often star centroids snap to the extracted positions (green circles).

Cybis320:
Ok, found the issue when in manual reduction in frame picking mode and picking fast moving objects. It should be fixed now.

dvida:
It works, thanks!

g7gpr:
Can you add a check that the loaded .config file stationID matches the stationID of the FITS files?

Also, it would be great if you could choose a place to save the generated platepar.

Cybis320:
Good idea regarding stationid.

You can already specify where to save the pp with 'save default platepar...', no?

g7gpr:
Oh - yes you can; sorry. It's a funny set of words; I presumed that save default platepar meant save a template platepar with no calibration information.

Cybis320:
I think the 'save default platepar ...' is not very intuitive. In general, we have several potential input sources and output locations, and the whole flow is not very intuitive. 

The .config, mask.bmp, flat.bmp, and platepar could be sourced/saved in the FF Files folder, `~/source/RMS`, `~/source/Stations/<stationID>`, or a custom location. The UI is not handling this very elegantly at the moment.

dvida:
Agreed, there should definitely be a box that shows up that asks the user to confirm the path or choose another path.

g7gpr:
Changing, "save default platepar" to "save as..." would make it a lot more straightforward.

Cybis320:
OK, I created a centralized 'File manager...' dialog which offers one identical interface for all code paths.
Screenshot 2026-02-19 at 10 48 20 AM

Also, am I the only one annoyed by the help overlay being displayed by default? Should it not be hidden by default?

dvida:
Looks great! Can you also add the "dark" as one of the options?

I'm fine with hiding the hotkey guide by default.

Cybis320:
> Looks great! Can you also add the "dark" as one of the options?
> 
> I'm fine with hiding the hotkey guide by default.

Done.

PR #763, opened 2025-12-05
Simplify installer: use system packages instead of building from source

Cybis320:
- aligns the RMS installation method between RMS_Update.sh and RMS_Installer.sh.
  - Remove packages needed to build OpenCV and the OpenCV build script
  - Install GStreamer, PyGObject, OpenCV, and PyQt5 via system packages
  - Conditionally install ai-edge-litert==2.1.0rc1 on python_version >= '3.13' (Trixie)

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -40,7 +40,7 @@ import json
 
 
 from RMS.Misc import obfuscatePassword
-from RMS.Routines.GstreamerCapture import GstVideoFile
+from RMS.Routines.GstreamerCapture import GstVideoFile, getStructureValue
 from RMS.Formats.ObservationSummary import getObsDBConn, addObsParam
 from RMS.RawFrameSave import RawFrameSaver
 from RMS.Misc import RmsDateTime, mkdirP, UTCFromTimestamp
@@ -1366,8 +1366,8 @@ class BufferedCapture(Process):
                         raise ValueError("Could not determine frame shape.")
                     
                     # Extract width, height, and format, and create frame
-                    width = structure.get_value('width')
-                    height = structure.get_value('height')
+                    width = getStructureValue(structure, 'width')
+                    height = getStructureValue(structure, 'height')
 
                     if self.config.gst_colorspace == 'GRAY8':
                         self.frame_shape = (height, width)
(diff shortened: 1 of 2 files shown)


dvida:
That's nice and elegant - can you confirm that the lib version comes with gstreamer, ffmpeg, and v4l2 support? You should be able to inspect using cv2.getBuildInformation()

Cybis320:
Confirmed. The OpenCV 4.10.0 build in the vRMS venv includes all three:
GStreamer | YES (1.26.0)
FFMPEG    | YES (avcodec 61.19.101, avformat 61.7.100, avutil 59.39.100, swscale 8.3.100) |
V4L2      | YES (linux/videodev2.h)
 
This was tested on a completely virgin Trixie install with only RMS_Installer.sh having been run.

dvida:
Brilliant! I presume this needs a bit more testing before we roll it out?

satmonkey:
We do not have the Trixie RMS image yet, is this cv2 backend support also in the Raspbian Bookworm?

satmonkey:
If the same backends are not in the previous Raspbian distros, then the only way how to remove the opencv4 install script, is to officially decomission those RMS distros as not supported any longer. If this is not the case, I suggest to keep this install script as part of the repo.

Cybis320:
The cv2 backend with V4L2 should still be available. There should be no impact on that.

If a machine currently has access to OpenCV in its virtual environment, it won't be removed by running RMS_Update.

If one runs RMS_Installer, which is typically done on a new machine, or a new image, but could also be done on an existing machine, then OpenCV will be picked up from the OS. 

It should be a seamless transition. At least, that's what I think will happen- but it needs to be confirmed.

Cybis320:
A couple tests we should run on the oldest distro we want to support:
 - backup your files
 - switch to this branch (update-installer)
 - run RMS_Update. Test whether the cv2 backend still works
 - run RMS_Installer. Test cv2 backend
 
Any volonteer? :)

Cybis320:
Tested the RMS_Update-only route on Buster, Trixie, and Ubuntu - no issues with either gst or cv2 backends.

Trixie initially had an issue with GStreamer 1.26, which introduced some breaking changes. This is fixed separately in PR #764.

Note on Buster support: it's been moved to the archive since reaching end-of-life, so apt update currently fails out of the box. This means RMS_Installer also fails, so I couldn't test that route. But if someone's doing a fresh install anyway, they should probably just start from Trixie.

Issue #769, opened 2025-12-20
EventMonitor retrieval delayed on MultiCam linux


g7gpr:
EventMonitor looks for platepar file in current captured directory, or ~/source/RMS

If capture is in progress on a system where multiple stations share a single user account, no valid platepar is found, and event retrieval occurs after the next station restart. This has been observed on other occasions, but never explained. For stations that rarely restart this means that events are lost. 

I do not think it is a good idea to copy the active platepar into the captured directory at the start of capture, as the operator may update the platepar during capture. I think that the current behaviour, copying across at the end, is the correct way to handle this.

g7gpr:
Merged

PR #770, opened 2025-12-20
Fix/em/platepar multicam

g7gpr:
This is a fix for issue #769 

Rather than using rms_root_dir as the path for the platepar file, uses config_file_path, as the platepar file is always in the same directory as the config file on all types of stations.

### DIFF
--- a/RMS/EventMonitor.py
+++ b/RMS/EventMonitor.py
@@ -4813,7 +4813,7 @@ if __name__ == "__main__":
     arg_parser = argparse.ArgumentParser(description="""Check a web page for trajectories, and upload relevant data. \
         """, formatter_class=argparse.RawTextHelpFormatter)
 
-    arg_parser.add_argument('-c', '--config', metavar='CONFIG_PATH', type=str,
+    arg_parser.add_argument('-c', '--config', nargs=1, metavar='CONFIG_PATH', type=str,
                             help="Path to a config file which will be used instead of the default one.")
 
     arg_parser.add_argument('-o', '--oneshot', dest='one_shot', default=False, action="store_true",
@@ -4833,10 +4833,12 @@ if __name__ == "__main__":
 
     # Load the config file
     if cml_args.config is None:
-        syscon = cr.parse(os.path.join(os.getcwd(),".config"))
+        syscon = cr.loadConfigFromDirectory(".config", os.getcwd())
     else:
-        syscon = cr.parse(os.path.expanduser(cml_args.config))
+        syscon = cr.loadConfigFromDirectory(cml_args.config, os.getcwd())
     # Set the web page to monitor
+
+    print(f"Loaded config for {syscon.stationID}")
 
     if testIndividuals():
         log.info("Individual function test success")


g7gpr:
The preference remains to collect the platepar from the captured files directory, however if capture is still in progress, then the comfig_file_path is used as the path to the platepar.

RMS/EventMonitor.py
```diff
@@ -4836,7 +4835,7 @@ def testIndividuals(logging = True):
     if cml_args.config is None:
         syscon = cr.parse(os.path.join(os.getcwd(),".config"))
     else:
-        syscon = cr.parse(os.path.expanduser(cml_args.config[0]))
+        syscon = cr.parse(os.path.expanduser(cml_args.config))
```
markmac99: It'd be more consistent to use the standard cr.loadConfigFromDirectory() rather than parsing the file directly. This function ignores the 2nd argument if a full path to the config file is supplied, and handles errors consistently with the rest of RMS. 
```
syscon = cr.loadConfigFromDirectory(cml_args.config, 'notused')
```
Also i'm pretty sure arg_parser returns a list of strings (with one entry if only one is supplied), but again calling the std function would resolve this since it handles the parsed arg as needed. 
```
$ cat test.py
import argparse
arg_parser = argparse.ArgumentParser()
arg_parser.add_argument('-c', '--config', nargs=1, metavar='CONFIG_PATH', type=str)
cml_args = arg_parser.parse_args()
print(cml_args.config)
$ python C:\temp\test.py -c some/path/to/file
['some/path/to/file']
```
```

File: RMS/EventMonitor.py [hunk 2]
```diff
@@ -1511,7 +1511,7 @@ def getPlateparFilePath(self, event):
             if len(platepar_file_list) > 0:
                 platepar_file = platepar_file_list[0]
             else:
-                platepar_file = os.path.join(self.syscon.rms_root_dir, self.syscon.platepar_name)
+                platepar_file = os.path.join(self.syscon.config_file_path, self.syscon.platepar_name)
```
markmac99, line 1514: looks good

RMS/EventMonitor.py
```diff
@@ -3029,8 +3029,7 @@ def dictMagsRaDec(config, r, d, e_jd=0, l_jd=np.inf, max_number_of_images=300, w
     """
 
     full_path_to_archived = os.path.expanduser(os.path.join(str(config.data_dir), str(config.archived_dir)))
-    full_path_to_default_platepar = os.path.join(getRmsRootDir(), config.platepar_name)
-
+    full_path_to_default_platepar = os.path.join(config.config_file_path, config.platepar_name)
```
markmac99, line 3032: looks good

RMS/EventMonitor.py
g7gpr: I'll look into that. All those methods have me rather confused.

g7gpr:
Got some work to do, so putting back to draft.

RMS/EventMonitor.py
g7gpr: If nargs are specified, returns a list, otherwise a string.

g7gpr:
I believe this is ready to merge. Tested on several stations with good results.

PR #773, opened 2025-12-22
Add disabling external script logging.

dvida:
Added an option to disable all logging done in the external script.

### DIFF
--- a/.config
+++ b/.config
@@ -66,10 +66,9 @@ camera_group_name: none
 ; three arguments: captured_night_dir, archived_night_dir, config 
 external_script_run: false
 
-; Toggle logging of the external script. If true, start/stop messages are logged.
-; Also, if false, the script process will have all log handlers removed to
-; prevent interference with its own logging.
-external_script_log: true
+; Toggle logging of the external script. If true, logging  will be captured by RMS from the external script. 
+; If false, the script process will have all log handlers removed to prevent interference with its own logging.
+external_script_log: false
 
 
 ; Run the external script after auto reprocess.
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -276,7 +276,7 @@ class Config:
 
 
         self.external_script_run = False
-        self.external_script_log = True
+        self.external_script_log = False
         self.auto_reprocess_external_script_run = False
         self.external_script_path = None
         self.external_function_name = "rmsExternal"


dvida:
@g7gpr great job! Let's wait for @markmac99 to confirm that this works well on his UKMON setup.

RMS/RunExternalScript.py
```diff
@@ -55,13 +74,26 @@ def runExternalScript(captured_night_dir, archived_night_dir, config):
...
 
-        log.info('Starting function "{}" from external script "{}"'.format(externalFunction, module))
-
         # Call the external function in a separate process, protecting the main process from potential crashes
-        p = multiprocessing.Process(target=externalFunction, args=(captured_night_dir, archived_night_dir, config))
+        
+        # If logging is disabled, create a wrapper function which removes all log handlers in the external process
+        if not config.external_script_log:
+            # Use the wrapper function
+            log.info('Starting function "{}" from external script "{}" with logging inhibited'.format(externalFunction, module))
```
g7gpr, line 84: This line was included for development purposes. I still think it is useful, but it does not line up with the text description in the .config file, I suggest remove it.

RMS/RunExternalScript.py
```diff
@@ -16,6 +16,25 @@
...
+                root.removeHandler(handler)
+                handler.close()
+
+    # Call the external function
+    func(captured_night_dir, archived_night_dir, config)
+
+
+
 def runExternalScript(captured_night_dir, archived_night_dir, config):
     """ Run the external script. It's results won't be returned to the main program, the script will just be run as a separate process.
```
g7gpr: Nit pick, no apostrophe in possessive "Its", not "It's"

RMS/RunExternalScript.py
dvida: I'm not sure I understand, do you mean the log.info line?

g7gpr: Yes - line 82. Having the start of the script logged when logging is disabled does not line up with the comment line in the .config file. 

```
; Toggle logging of the external script. If true, start/stop messages are logged.
```

I understand "if true" is not the same as "if and only if", but informally, I don't think logging start if false is quite right

dvida: Got it, feel free to update the comment.

markmac99: I think the log message is useful, as it lets you know the extenal script was started, but perhaps the comment in the config file should be updated 
```
Toggle logging of the external script. If true, no logging  will be captured by RMS from the external script. 
```

markmac99 (approved):
All seems good. I've tested it on a spare Pi, with logging disabled, only the message about logging inhibited appeared:
```
2026/01/02 16:51:37-INFO-RunExternalScript-line:84 - Starting function "<function rmsExternal at 0x7f8a9c2ca0>" from external script "<module 'ukmonPostProc' from '/home/rms/source/ukmon-pitools-UK0006/ukmonPostProc.py'>" with logging inhibited
```
with the flag set to true, all the logging went into the RMS log as before.

Obviously i'd prefer the default to be false, since this would match RMS's original behaviour from before the recent updates to logging. But this does give me a mechanism to inhibit unwanted logging into the RMS log. I'll ask my ukmon colleagues to edit their config files as needed, or possibly enhance our toolset to add the parameter to the config file, if missing.

RMS/RunExternalScript.py
markmac99: editor autocorrect i should think - mine does love changing its to it's... :)

dvida:
Got it, I changed the defaults to false. It just makes sense and it will prevent external scripts from possiblty flooding the RMS log.

PR #781, opened 2025-12-26
Substitute default values in .config file from a template file

g7gpr:
If a .config file contains a value set as 'default' substitute in the value taken from the .config file in ~/source/RMS/RMS/ConfigurationDefaults/.config

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -778,7 +778,16 @@ def removeInlineComments(cfgparser, delimiter):
         [cfgparser.set(section, item[0], item[1].split(delimiter)[0].strip()) for item in cfgparser.items(section)]
 
 
-def getDefaultValue(section_key, attribute_key):
+def getDefaultValue(attribute_key):
+    """
+    Get default value for an attribute from the configuration file containing default values.
+
+    Arguments:
+        attribute_key: Key for the configuration item to be substituted
+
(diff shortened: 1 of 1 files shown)


markmac99:
I'm just wondering how this is different to having the default set in RMS/ConfigReader.py, and what use-case we envisage. 

Firstly, we can already roll out a new default of any config setting by updating ConfigReader. So for new parameters that are missing in the user's config file, these would anyway be defaulted via ConfigReader. It won't allow us to override existing parameters for which the user has specified a value (eg the catalogue file), because presumably we still have to honour the users' settings.

g7gpr:
I agree with you. I just wanted to see if I could modify object properties like this. 

What would be more useful, is to implement different properties based on machine available resources.

Even so, scope for breaking perfectly good stations is large, and what benefit flows from it?

dvida:
It's an interesting proposition. However, a config template is already created by the FirstRun script once the repo is reset and updated. That way, we don't have to maintain a separate config file. Second, any "reset" feature could easily be rolled into MigrateConfig. It could check if the value has been set to 'default' and voila, it will update it to the default value from configreader. But as Mark indicates, an easier way to do this would be comment out the config entry and RMS will automatically pull the default value.

g7gpr:
Yep, that makes sense.

Issue #795, opened 2026-01-18
Observation summary reports commit of remote branch, not the upstream branch


g7gpr:
In https://globalmeteornetwork.groups.io/g/coordinators/message/668 it was reported that the ObservationSummary does not report the upstream branch, but rather the source of the latest commit in the upstream branch. 

Whilst this was as intended, it is only helpful for this purpose. Generally operators wish to know the upstream branch, and not the source of the latest commit on the upstream branch. 

Change ObservationSummary to report the upstream branch.

g7gpr:
Code changes made, closing.

PR #796, opened 2026-01-18
Report upstream branch

g7gpr:
This PR relates to issue #795 

It is intended to report the upstream branch, which is more useful than the upstream source of the current commit. 

At the moment, it adds a key, this is just while testing is in progress. Ultimately the original key "upstream_branch" will be used and report the correct information.


g7gpr:
On test on au001a

g7gpr:
OK, let's go. I'll do some more work here.

Issue #798, opened 2026-01-19
Allow user to choose the colour used for sporadics in the ShowerAssociation plot


markmac99:
Currently its set to 'gray' at line 726 in Utils/ShowerAssociation.py

Suggest we allow this to be configured via the [colors] section of the config file, something like
sporadic_color: gray

markmac99:
done in early 2026, closing ticket

PR #799, opened 2026-01-21
add support to change colour of sporadics on the radiants map

markmac99:
This addresses issue #798 by adding an option to the config file [colors] section to set the colour of sporadics on the radiant plots. Colour can be any of the matplotlib CSS colours specified on this page
https://matplotlib.org/stable/gallery/color/named_colors.html

If an invalid colour is chosen or no colour is specified either in the config or on the commandline, it defaults to gray. 

When called standalone, you can add an extra parameter --sporadic_color to specify the desired colour. If you don't specify, it'll be read from the config file, and if there's no setting there, then the default gray is used. 

Example standalone call:
```
python -m Utils.ShowerAssociation -c .\UK0006\.config .\UK0006 -x --sporadic_color blueviolet
```
Attached examples with colours green, white and blueviolet. White is actually a pretty good choice to make the sporadics show up. 
UK0006_20251213_163400_129262_radiants_green
UK0006_20251213_163400_129262_radiants_white
UK0006_20251213_163400_129262_radiants_blueviolet

### DIFF
--- a/Utils/ShowerAssociation.py
+++ b/Utils/ShowerAssociation.py
@@ -11,6 +11,8 @@ import sys
 import matplotlib.gridspec as gridspec
 import matplotlib.pyplot as plt
 import numpy as np
+import matplotlib.colors as mcolors
+
 from RMS.Astrometry.Conversions import (EARTH_CONSTANTS, datetime2JD,
                                         jd2Date, raDec2AltAz, raDec2Vector,
                                         vector2RaDec, AEH2Range)
@@ -1153,6 +1155,9 @@ if __name__ == "__main__":
     sporadic_color = cml_args.sporadic_color
     if sporadic_color is None:
         sporadic_color = config.sporadic_color
+    if sporadic_color not in mcolors.CSS4_COLORS:
+        sporadic_color = 'gray'
+
        
 
     # Perform shower association


markmac99:
Currently running overnight on UK001L, though it is solid cloud and rain here....

Cybis320:
Great idea. Is there any reason to keep gray as the default at all? White is pretty nice, I think, as it stands out and is guaranteed to be unique.

markmac99:
> Great idea. Is there any reason to keep gray as the default at all? White is pretty nice, I think, as it stands out and is guaranteed to be unique.

Purely for backwards compatability since thas the current colour. I agree white would be nicer, and originally i defaulted to that but i am a believer in not making unexpected changes to interfaces, even human ones!

dvida:
Amazing! The sporadic colour used to be brighter, but then there is a complete mess if there are a lot of false positives. Grey is good, unless we make the colour scaling dynamic with the number of sporadics (few = white, lots = grey).

markmac99:
> The sporadic colour used to be brighter, but then there is a complete mess if there are a lot of false positives. 

It might be that with improvements to the detector, this is less of a problem. Would be interesting to check on a night with many clouds/moonlight/satellites. 

>Grey is good, unless we make the colour scaling dynamic with the number of sporadics (few = white, lots = grey).

Yea, thats why i thought it best to leave it to the end-user to choose!

Issue #800, opened 2026-01-22
New video segment created in 1970


g7gpr:
```
2026/01/22 09:33:24-INFO-BufferedCapture-line:891 - Created new video segment #0 at: /home/au000a/RMS_data/VideoFiles/1970/19700101-001/19700101-001_00/AU000A_19700101_000000_233171_video.mkv
```

Not sure if this is important, it was only the first file created.

dvida:
Fixed by #801

PR #801, opened 2026-01-24
Set start_timestamp after PAUSED but before PLAYING

Cybis320:
Address issue #800 by setting start_timestamp after PAUSED but before PLAYING.


dvida:
Awesome, is this ready for merging?

PR #802, opened 2026-01-28
Reduce reliance on config FoV_W for auto fit

g7gpr:
This has solved for all the lenses that I can test.


Cybis320:
This is more robust but slower. We could first try the narrow range around the hint and fall back to this if it fails?

dvida:
Agreed - try with the default range first and then if it fails, expand the search space.

g7gpr:
OK, I'll have to leave it for others to implement this, if it needs to be implemented.

PR #803, opened 2026-01-28
Quieten continuous capture logging

g7gpr:
Logging from continuous capture can obscure other log messages. 

This PR adds a config option

```
; Quieten the logging from continous capture
quieten_continuous_capture_logging: false
```

When true it stops only the following messages being logged.

```
2026/01/28 09:40:54-INFO-BufferedCapture-line:2192 - Buffer fill: 7.7%. Dropped frames: 0 (last 10 min), 0 this session
2026/01/28 09:40:54-INFO-BufferedCapture-line:1931 - Grabbing a new block of 256 frames...
2026/01/28 09:40:55-INFO-RawFrameSave-line:99 - Saving block of 10 raw frames to disk (day mode)
```

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -483,7 +483,7 @@ class Config:
         # Switch camera settings between day/night modes, for when continuous_capture is enabled
         self.switch_camera_modes = False
 
-        # Silence most of the logging from continuous capture
+        # Quieten the logging from continuous capture
         self.quieten_continuous_capture_logging = False
 
         ##### Upload


Cybis320:
Not a fan of giving the option to not log this critical stat as it would make troubleshooting much harder. Maybe optionally move them to debug? But I would personally keep them at info as this is convenient to quickly assess the health of the pipeline during capture.

dvida:
I also agree that we should keep them, as they are vital to assess what's going on, especially when the camera is being initially set up. I very often rely on the "grabbing a new block" message to check that things are actually working and not hanging.

g7gpr:
Agree with these comments; it was something another operator asked for. 

I've advised him that 

```
tail -F ~/RMS_data/*.log | grep -v -e BufferedCapture -e RawFrameSave
```

probably will let him do what he wants to do.

PR #809, opened 2026-01-31
Install cython

g7gpr:
On a new install of trixie, cython was missing, preventing a build of imreg_dft. 

This PR adds cython.

### DIFF
--- a/Scripts/RMS_Installer.sh
+++ b/Scripts/RMS_Installer.sh
@@ -60,7 +60,7 @@ sudo systemctl start ssh
 sudo systemctl enable ssh
 
 # Clone repositories
-git clone https://github.com/g7gpr/RMS.git
+git clone https://github.com/CroatianMeteorNetwork/RMS.git
 git clone https://github.com/CroatianMeteorNetwork/cmn_binviewer.git
 
 # Set up Python virtual environment with access to system packages
@@ -69,7 +69,6 @@ python3 -m venv --system-site-packages vRMS
 source ~/vRMS/bin/activate
 
 # Install Python packages
-python -m pip install psutil
 python -m pip install --upgrade pip setuptools wheel
 python -m pip install -r ~/source/RMS/requirements.txt
 


g7gpr:
Closing this; too many strange things going on.

Cybis320:
OK, I'm not sure if this is related to the issue you encountered, but pip 26.0 is out, and it's not liking this line in requirements:
```
imreg_dft @ git+https://github.com/matejak/imreg_dft@master#egg=imreg_dft>'2.0.0'
```
Pip has been ignoring the `#egg=imreg_dft>'2.0.0'` for a while now, but with 26, it's throwing a syntax error.
I think just changing it to this should do:
```
imreg_dft @ git+https://github.com/matejak/imreg_dft@master
```
Although, with the upcoming change from FFTAlign to NNAlign, imreg_dft is dropped from requirements. Also, FFTAlign never worked as far as I can tell, and imreg_dft has always been dead weight. So, I'm not sure if we should release an urgent fix to master or wait for NNAlign to be rolled out. 

RMS_Update in master fails with pip 26 and maybe we can't wait...

g7gpr:
Indeed, it is all related.

g7gpr:
If RMS_Update is failing, I think this does require a response. I'll raise it as an issue; but unfortunately I don't have the time this weekend to address it.

g7gpr:
I've reopened this as I'm confident that it is a real issue, and this is one possible solution. I'm doing a rebuild on a trixie system tonight, so I'll keep this updated.

g7gpr:
The scripts have to point to my fork, so I can do testing, I'll put this back to CroatianMeteorNetwork before I take this out of draft

Cybis320:
Cython is already in requirements.txt. The issue with pip 26 is that the install fails early due to the new syntax requirement, before Cython gets installed. Fixing the syntax should be sufficient, no?

With this fix, we'd have two `pip install cython` - one new one in RMS_Installer and one from `pip install -r ~/source/RMS/requirements.txt` which already contains cython.

Also, why is psutil required? It's not a dependency of RMS.

Note that this wasn't an issue with Trixie until pip 26 was introduced last week.

g7gpr:
psutil is required for my fork of RMS so I can get a clean shutdown of all the subprocesses when RMS exits; I'll take it out when I take this out of draft, at the same time as I change the install script to point to CroatianMeteorNetwork. However, I'll take out the pip install cython now.

Cybis320:
Looks good to me! thanks!

PR #814, opened 2026-02-06
Multiple external scripts

g7gpr:
This PR adds the ability to run multiple external scripts concurrently, and reboot is prevented until all scripts have terminated.

Scripts can be listed in the .config file as a comma separated list, 

```
external_script_path: ~/source/RMS/Tests/RsyncUploader.py,
                      ~/source/RMS/Tests/ExternalTest2Minutes.py, 
                      ~/source/RMS/Tests/ExternalTest4Minutes.py,
                      ~/source/RMS/Tests/ExternalTest8Minutes.py, 
                      ~/source/RMS/Tests/ExternalTest8Minutes.py,
                      ~/source/RMS/Tests/ExternalTest16Minutes.py, 
                      ~/source/RMS/Tests/TrackStackExternalScript.py
```

These scripts have been created to test the process tracking. 

When the scripts are launched, the pid of the new process is added to a dictionary. Once all the scripts have been launched, this dictionary is returned to the main function.

 A new function is added to the RunExternalScript module.

```
def runningExternalScripts(external_script_process_dict):
    
    Given a dictionary of external processes, are any still running.

    Arguments:
        external_script_process_dict: [dict] Dictionary of external script processes.

    Return:
        [bool]: True if any external scripts are still running.
```

This augments the existing .reboot_lock facility, and ensures that reboot is prevented until all the scripts have terminated.

### DIFF
--- a/RMS/RunExternalScript.py
+++ b/RMS/RunExternalScript.py
@@ -84,7 +84,7 @@ def runExternalScript(captured_night_dir, archived_night_dir, config):
         if not os.path.isfile(external_script_path):
             log.error('The script {:s} does not exist!'.format(external_script_path))
             continue
-        time.sleep(1)
+
 
         try:
 


dvida:
Hi Dave,
The purpose of calling a custom python script is for the users to adjust it according to their needs, which encompases calling multiple scripts. I'd rather not maintain additional features in RMS that users could deploy themselves for their needs.

Issue #822, opened 2026-02-12
Speeding up the processing after capture has finished


g7gpr:
A few of the tasks at the end of processing take most of the time. 
Making thumbnails, stacking capture images and the timelapse could be built as the observation session progresses.

Is this worth exploring, or is it too much work and complexity for too little gain?

dvida:
Most of these are IO limited as RMS needs to read all images to make those plots. It's hard to get around that. I'd say leave it as is.

g7gpr:
The idea was to build the thumbnails and stacks in memory, frame by frame, as a capture progressed, but happy to leave this.

dvida:
That's a pretty good idea, I didn't see it that way. Give it a shot!

g7gpr:
This is not going smoothly. Shelving it for now.

PR #826, opened 2026-02-13
Start work on generating thumbnails during capture

g7gpr:
This PR is in response to issue #822, and is a test of creating thumbnails, and possibly timelapse on the fly. 

It creates a class Thumbnail, to which images are added. The images are placed on a canvas in the same way as the original GenerateThumbanails process. The key difference is that the time taken to produce and save the Thumbnail at the end of capture is reduced by at least one order of magnitude. 

This is an example of a thumbnail produced 'on the fly'

AU000A_20260204_115819_413028_CAPTURED_on_the_fly_thumbs

And by the conventional technique, which reads all the files in at the end.

AU000A_20260204_115819_413028_CAPTURED_thumbs

### DIFF
--- a/RMS/Compression.py
+++ b/RMS/Compression.py
@@ -38,7 +38,6 @@ from RMS.Routines.Image import saveImage
 import pyximport
 pyximport.install(setup_args={'include_dirs':[np.get_include()]})
 from RMS.CompressionCy import compressFrames
-from Utils.GenerateThumbnails import Thumbnail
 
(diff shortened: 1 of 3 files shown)


g7gpr:
It works as a concept, but integration with the capture process is either going to require reloading the images after they are saved - which is adding IO load at the worst possible time, or going to need a lot of work. Neither of those options are very appealing right now, so I'm putting this to one side.

PR #827, opened 2026-02-13
Support camera firmware operations in CameraControl

Cybis320:
This PR adds the ability to retrieve the camera firmware version and build time, report it in Observation Summary, and support updating the camera firmware from Linux.

```
(vRMS) ops@procamsff02:~/source/RMS$ python -m Utils.CameraControl GetDeviceInformation -c ~/source/Stations/US05D1/.config
Loading config file: /home/ops/source/Stations/US05D1/.config
Camera settings file: ./camera_settings.json
Device Model    : IPC_GK7205V200_G3S
Hardware        : IPC_GK7205V200_G3S
Firmware        : V5.00.R02.000739AG.10010.140400.0020000
Build Time      : 2023-02-02 19:24:54
Serial No       : #############
```

The motivation is that the same camera model can behave very differently with different firmware version (or even with the same reported camera firmware but different build time... G3S).

New firmware for the G3S is in the works and currently updating the firmware would require using a Windows utiliy.

To update FW with this PR:
```
python -m Utils.CameraControl UpgradeFirmware /path/to/firmware.bin -c /path/to/.config 
```


Cybis320:
Changed the minimum ExposureParam from 100 microseconds to 10 microseconds, as some camera and firmware combination are able to capture 10 microseconds exposures, which increases the dynamic range during day capture.

Issue #807, opened 2026-01-30
Meteor capture never restarts after a failed reboot if reboot after processing is selected


g7gpr:
Originally if 

```
reboot_after_processing: true
```

was selected, but reboot did not occur for some reason, perhaps meteor capture crashed, meteor capture would start again on the next evening. Now if reboot is selected, meteor capture seems to run in a one-shot mode. So meteor capture stops, but continuous capture keeps running forever. Because meteor capture never restarted, and meteor capture is responsible for calling 

```
shutdown -r now
```

A reboot never occurs. 

I think a few people have seen this, but perhaps this is the mechanism for it.

Cybis320:
PR #833 addresses issue #805 and #807. However, it seems to me `reboot_after_processing: true` can only work when StartCapture is started from the desktop or on passwordless sudo systems.
                                                                                
When started from cron, screen, or headless, `sudo shutdown -r now` fails silently. The reboot loop then fires `shutdown -r now` 240 times in rapid succession (no sleep between attempts when reboot_go is True), and previously would re-enter the loop indefinitely on the next iteration of the outer loop.                                               

With this fix, after exhausting the 240 reboot attempts, ran_once is reset so continuous capture resumes normally instead of getting stuck.

g7gpr:
Love your work!

g7gpr:
I believe this is resolved now.

Issue #805, opened 2026-01-30
[prerelease] Interrupt from RawFrameSaver erroneously logged as Ctrl-C from console


g7gpr:
```
2026/01/30 06:55:47-INFO-BufferedCapture-line:1176 - Pipeline successfully created and started
2026/01/30 06:55:47-INFO-BufferedCapture-line:1357 - GStreamer pipeline created!
2026/01/30 06:55:48-INFO-BufferedCapture-line:1397 - Video format: BGR, 720P, color: True
2026/01/30 06:55:51-INFO-BufferedCapture-line:1931 - Grabbing a new block of 256 frames...
2026/01/30 06:55:51-INFO-BufferedCapture-line:1996 - Frame shape/mode changed, reinitializing arrays...
2026/01/30 06:55:51-DEBUG-BufferedCapture-line:1659 - Creating shared arrays with shape: (10, 720, 1280, 3)
2026/01/30 06:55:56-INFO-BufferedCapture-line:2024 - Successfully reinitialized raw frame handling
2026/01/30 06:56:00-INFO-BufferedCapture-line:535 - GStreamer pipeline did not emit a sample.
2026/01/30 06:56:00-INFO-BufferedCapture-line:1945 - Frame grabbing failed, video device is probably disconnected!
2026/01/30 06:56:00-DEBUG-BufferedCapture-line:1460 - releaseResources: Starting
2026/01/30 06:56:00-DEBUG-BufferedCapture-line:1475 - releaseResources: Calling releaseRawArrays()
2026/01/30 06:56:00-DEBUG-RawFrameSave-line:168 - Raw frame saver exit flag set
2026/01/30 06:56:32-INFO-RawFrameSave-line:177 - Flushing 2 tail-end raw frames before shutdown
2026/01/30 06:56:32-INFO-RawFrameSave-line:98 - Saving block of 2 raw frames to disk (day mode)
2026/01/30 06:56:33-DEBUG-RawFrameSave-line:155 - Frame saved: AU000D_20260130_065549_999_d.jpg
2026/01/30 06:56:34-DEBUG-RawFrameSave-line:155 - Frame saved: AU000D_20260130_065554_999_d.jpg
2026/01/30 06:56:34-DEBUG-RawFrameSave-line:188 - Freeing frame buffers in raw frame saver...
2026/01/30 06:56:39-WARNING-BufferedCapture-line:1607 - RawFrameSaver still busy. Sending interrupt signal...
[...]

PR #833, opened 2026-02-20
Fix continuous capture stuck after failed reboot (#807)

Cybis320:
Address issue #807 by resuming capture instead of looping reboot attempts indefinitely when reboot_after_processing is enabled in continuous capture mode. Resets ran_once after exhausting reboot retries so the outer loop proceeds to restart capture normally.


Cybis320:
Also addresses issue #805 by replacing SIGINT with SIGTERM for child process cleanup to prevent spurious STOP_CAPTURE

PR #838, opened 2026-03-06
Update requirements.txt for bcrypt and cryptography

Cybis320:
When rebuilding a new vRMS venv on Buster, pip install fails because newer versions of bcrypt and cryptography require a Rust compiler, which is not available on most Pi setups. This PR caps bcrypt and cryptography versions for Python 3.7 compatibility.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,9 @@
 gitpython>=2.1.8
 # Use versions 2.4.0 due to compatibility issues with OpenSSL on Pi3/Jessie
 paramiko>=2.9.0; python_version>='3.6'
+# Cap bcrypt and cryptography for Python 3.7 (no prebuilt armv7l wheels for newer versions)
+bcrypt<4.0; python_version<'3.8'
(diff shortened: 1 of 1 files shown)


markmac99:
Rust is available for bookworm and later, where python 3.13 or later would be the default i think. Could we therefore pin based on the python version, so that newer builds are not potentially using outdated or insecure encryption protocols?

dvida:
Great idea - that's should already be the case :)
image

markmac99:
> Great idea - that's should already be the case :) 

Perfect - i failed to notice that !

markmac99:
I guess we're assuming that some other requirement will pull in bcrypt and cryptography and the pin will only be applied if python <3.8. 
I guess thats ok, though personally i would make the requirement explicit, something like
bcrypt<4.0 ; python_version<3.8
bcrypt; python_version>=3.8

Admittedly I suspect it won't matter as something will for sure import the cryptography libs. :)

Issue #845, opened 2026-03-13
Automatically build and load missing .so files


g7gpr:
Some operators have faced kht symbol errors 

```
AttributeError: python: undefined symbol: kht_wrapper
```
when the kht*.so file is missing, leading to many nights of good capture, but no detections. 

https://globalmeteornetwork.groups.io/g/coordinators/topic/118287234?p=Created,,,20,1,0,0

RMS could attempt to rebuild the .so on the fly, and then continue execution. If the .so file is missing, then there is no risk of namespace pollution.

PR #846, opened 2026-03-13
Detect and fix missing kht files

g7gpr:
This PR relates to issue #845, and attempts to build missing .so files for the kernel Hough transform.

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -1671,7 +1671,7 @@ def parseMeteorDetection(config, parser):
         kht.kht_wrapper.restype = ctypes.c_size_t
         print("KHT loaded successfully")
 
-    # If loading KHT library fails get the OSError subclass
+    # If loading KHT library fails get the exception
     except Exception as e:
         # If the file exists remove it and record that namespace may be corrupted
 


RMS/ConfigReader.py
```diff
@@ -1637,9 +1643,58 @@ def parseMeteorDetection(config, parser):
...
+        kht.kht_wrapper.restype = ctypes.c_size_t
+        print("KHT loaded successfully")
+
+    # If loading KHT library fails get the exception
+    except Exception as e:
+        # If the file exists remove it and record that namespace may be corrupted
+
+        if os.path.exists(kht_lib_path):
+            if os.path.isfile(kht_lib_path):
+                print("Warning - kht file could not be loaded, but existed - namespace may be corrupted")
```
markmac99, line 1680: Can we think of phrasing for this message which might be more informative for non-linux / non-programmers? I'm not sure most will know what a namespace is or how to repair it.  
Maybe one option would be to have an FAQ page on the GMN wiki, and reference the FAQ in the error message.

RMS/ConfigReader.py
```diff
@@ -1637,9 +1643,58 @@ def parseMeteorDetection(config, parser):
...
+                                    ctypes.c_double,
+                                    ctypes.c_double]
+        kht.kht_wrapper.restype = ctypes.c_size_t
+        print("KHT loaded successfully")
+
+    # If loading KHT library fails get the exception
+    except Exception as e:
+        # If the file exists remove it and record that namespace may be corrupted
+
+        if os.path.exists(kht_lib_path):
```
markmac99, line 1678: I don't tihnk we need to test both path and file existence - if the path doesn't exist then os.path.isfile returns false:
```
>>> os.path.isfile('/nosuchpath/frobozz')
False
```

RMS/ConfigReader.py
```diff
@@ -1637,9 +1643,58 @@ def parseMeteorDetection(config, parser):
...
+    kht_lib_path = findBinaryPath(config, config.kht_build_dir, config.kht_binary_name,
         config.kht_binary_extension)
 
+    if kht_lib_path is None:
+        try:
+            pythonSetup()
+            kht_lib_path = findBinaryPath(config, config.kht_build_dir, config.kht_binary_name,
+                                          config.kht_binary_extension)
+        except Exception as e:
+            print(str(e).encode("ascii", "replace").decode("ascii"))
```
markmac99, line 1655: I dont think we decode error messages anywhere else, is this definitely necessary? I'm not sure some odd symbols onscreen will matter. 

Also can we use log,warning() rather than print(), because the former will defnitely go  into the RMS log whereas the latter might be 'lost' in the console logging

RMS/Misc.py
```diff
@@ -1168,4 +1169,73 @@ def sphericalDomainWrapping(ra_min, ra_max, dec_min, dec_max,
...
+        return
+
+    try:
+        # We hold the lock here
+        log_text = "Executing python setup.py"
+
+        log.info(log_text)
+
+        current_working_directory = os.path.normpath(os.path.expanduser(getRmsRootDir()))
+        source_rms = os.path.normpath(os.path.expanduser("~/source/RMS"))
```
markmac99, line 1203: Obviously this will break for anyone who installed RMS in a nonstandard location (i've seen some installed in /usr/local or in /opt ). Hard to work around, but we could maybe have a 'hidden' config option? Not sure of the best option. :(

RMS/Misc.py
```diff
@@ -1168,4 +1169,73 @@ def sphericalDomainWrapping(ra_min, ra_max, dec_min, dec_max,
     query = []
     query.append([ra_min, ra_max, ra_range_min, ra_range_max])
     query.append([dec_min, dec_max, dec_range_min, dec_range_max])
-    return nDimensionDomainSplit(query)
\ No newline at end of file
+    return nDimensionDomainSplit(query)
+
+
+def pythonSetup():
```
markmac99, line 1175: I assume the purpose of this is to force-recompile the C++ module. It should be possible to separate out the logic that does that - i had a look at setup.py a while back and think it can be done. This would avoid having to fully rebuild everything.

markmac99 (commented):
Looks good. A few comments inline. One thought is that by restructuring setup.py we could separate out the part that rebuilds kht_wrapper and make it less intrusive to rebuild just that. A while back i experimented with splitting setup.py into two scripts, one to build kht_wrapper and one to do everything else.

RMS/ConfigReader.py
g7gpr: Good point - I'll make this less obscure.

RMS/ConfigReader.py
g7gpr: Is it possible to instantiate log during config read? log doesn't exist here, but I can look into it. 

Yes, in this case I think decode is necessary, because the logger module seems to sometimes pick up console print, and non-ascii crashes the logger. Playing around around with System V ABI, and ELF formats could lead to anything appearing in the exception.

RMS/Misc.py
g7gpr: Neither am I. My own feeling is that if someone is canny enough to install RMS in somewhere none default, they can sort out binary compatibilities for themselves.

RMS/Misc.py
g7gpr: I've tried - and rebuilding the .so only fixed it sometimes - building everything generally fixed it. I'll do more work. I think my brand of kht wrapper issues might be unique - most people just seem to be missing the file. I seem to be unique in my ability to build them with a mismatched interface.

RMS/ConfigReader.py
markmac99: okay, seems sensible then. Makes me wonder if we need to decode any messages printed elsewhere!

RMS/Misc.py
markmac99: maybe another one for an FAQ on the wiki.

RMS/Misc.py
markmac99: haha! if the current approach works, then i guess stick with it. If/when i find time to finish rebuilding the Hough Transform as a standalone python module it might be simpler to solve.

RMS/ConfigReader.py
markmac99: Good point about the logger. Obviously it needs the log_dir from the config file. I'm just concerned that not all console prints get captured, and so the failure might go unnoticed.

RMS/Misc.py
g7gpr: Definitely. I suspect you will come up with a new solution the day before I get this packaged up nicely.

RMS/ConfigReader.py
g7gpr: Sitting back and reflecting, this needs re-engineering. Testing an binary interface compatibility in a config parser, is a bit odd. Probably a better place to make this test is at the start of capture - where log has already been instantiated.

dvida:
Hi all,
I don't want to derail this work in any way, but I'm working on an alternative to the KHT line detector that uses RANSAC and is in pure python. It's still very much a work in progress, but we should have an alternative to use instead of KHT. It's still in the development phase and needs a lot of testing. I also improved the thresholding which may require updating the thresholding parameters (not fun).

g7gpr:
I'm happy for this work to be derailed. Given that I have a different failure mechanism to everyone else, and I don't understand the failure mechanism that other people are experiencing, there is risk that this fix will introduce a non-terminating restart loop, which is probably worse than just recording fits files without detections. At least these can be reprocessed in the future. 

I've had a crack at it - which is the main thing.

markmac99:
> but I'm working on an alternative to the KHT line detector that uses RANSAC and is in pure python.

good to know. I will probably plug away at converting the KHT code as its an interesting learning exercise and you never know, someone might find a use for it as a module :)

PR #848, opened 2026-03-16
Flag tracebacks in observation summary and verify extensions after build

Cybis320:
- Scan session log files for tracebacks during finalization and add `traceback_count` to the observation summary when > 0
 - Verify all compiled extensions from setup.py are importable after RMS_Update build, fail the update if any are missing

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -1340,9 +1340,7 @@ def finalizeObservationSummary(config, night_data_dir, platepar=None):
         addObsParam(obs_db_conn, "repository_lag_remote_days", "Not determined")
 
     # Scan log files for tracebacks during this session
-    traceback_count = countTracebacksInLogs(config)
-    if traceback_count > 0:
-        addObsParam(obs_db_conn, "traceback_count", traceback_count)
+    addObsParam(obs_db_conn, "traceback_count", countTracebacksInLogs(config))
 
     obs_db_conn.close()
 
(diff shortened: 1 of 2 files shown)


g7gpr (changes requested):
Can you break the traceback counting work into it's own function? A big rewrite is in progress on ObservationSummary, splitting this into its own function will make the merge easier. Please add the traceback_count even when it is zero, otherwise folks will complain that it is 'missing'.

g7gpr (commented):
I think this discovers if the extensions can be imported, but not if they include the specific names in the namespace. 

If a problem is found, what does this do? Just send a message to the console?

Cybis320:
Right, if running RMS_Update attended, the user sees the error and can investigate. If unattended, the tracebacks are logged and the weblog can be changed to display a specific alert. Retries don't make sense here - if it fails to build, there is something wrong with the host.

g7gpr:
OK - I understand the intended functionality now; and it's safe, and I'll approve the merge and deal with the conflicts down the track. I don't agree the retries don't make sense. Something was wrong with the host at that moment; but 30 seconds later after the file system has had chance to flush buffers and calm down, maybe it would work?

g7gpr (approved):

g7gpr (approved):

dvida:
This one is still set as draft, is it ready now?

g7gpr:
Does anyone object if I take this work over and roll it into the updates I am doing for the next prelease? I'll try to maintain as much git blame as I can, but this will be a difficult merge as the code has moved on so much.

Cybis320:
> Does anyone object if I take this work over and roll it into the updates I am doing for the next prelease? I'll try to maintain as much git blame as I can, but this will be a difficult merge as the code has moved on so much.

No objection on my part. Thank you, David!

PR #851, opened 2026-03-16
Fix Qt error

Cybis320:
QMessageBox was incorrectly referenced from QtGui instead of QtWidgets in the qmessagebox() helper, causing an AttributeError whenever SkyFit2 tried to display a warning or error dialog.


markmac99:
Tested under windows, all seems to work well - thanks

markmac99 (commented):
all seems good

PR #852, opened 2026-03-17
Config secion: posibility to use old ML model

sanlupkim:
On CSI camera (non standard setup, no x264 compression) old ML model (meteorml32) has better results - up to 30% meteors are correctly detected.


dvida:
Thank you - could you please make the PR with respect to the "prerelease" branch? We never merge directly into master.

PR #853, opened 2026-03-18
Fix AutoPlatepar CALSTARS requirement and misleading success message

Cybis320:
- Auto-generate CALSTARS file when missing (matching SkyFit2 behavior)
 - Return `None` instead of unvalidated intermediate platepar when star matching fails, preventing misleading "SUCCESS" after "ERROR"


Cybis320:
I've replaced the brute-force matchStars in CyFunctions.pyx with a scipy.spatial.cKDTree implementation. The new version lives in RMS/Astrometry/MatchStars.py and is a drop-in replacement with the same signature and return format.

dvida:
Is the new approach really faster? Building KD trees takes time and by definition querying is faster only when done multiple times. Here, the KD tree is built for each run.

Cybis320:
> Is the new approach really faster? Building KD trees takes time and by definition querying is faster only when done multiple times. Here, the KD tree is built for each run.
 ```
  ┌───────────────────────┬──────────┬──────────┬─────────────────────┐
  │       Scenario        │  Cython  │ KD-tree  │       Winner        │
  ├───────────────────────┼──────────┼──────────┼─────────────────────┤
  │ Small (20×50)         │ 0.002 ms │ 0.035 ms │ Cython 18x faster   │
  ├───────────────────────┼──────────┼──────────┼─────────────────────┤
  │ Typical (200×500)     │ 0.147 ms │ 0.120 ms │ ~tie                │
  ├───────────────────────┼──────────┼──────────┼─────────────────────┤
  │ Large (500×2000)      │ 1.54 ms  │ 0.46 ms  │ KD-tree 3.3x faster │
  ├───────────────────────┼──────────┼──────────┼─────────────────────┤
  │ Dave 360 (2000×10000) │ 28.9 ms  │ 2.9 ms   │ KD-tree 10x faster  │
  └───────────────────────┴──────────┴──────────┴─────────────────────┘
```
Cython wins for very small inputs (20×50), that's 20 detected stars x 50 catalog stars, but those are 0.002ms - nobody will notice, and that's not a typical scenario. At typical RMS sizes it's a wash, and it scales much better for larger inputs.

matchStars works in pixel space, so the pixel positions change every time and the KD-tree has to be rebuilt for each frame.

PR #855, opened 2026-03-19
Fix memory leaks: zombie Compressor processes and inherited shared buffers

Cybis320:
Targets the OOM / swap-spiral failures seen on memory-tight stations. 
  
  **Capture memory (BufferedCapture, Compression, QueuedPool, ConfigReader)**
  - New `gst_queue_size` config (default 100) drives the GStreamer queue/appsink max-buffers, previously hardcoded to 100. Lower it to shrink per-camera buffering on tight boxes. The buffer-fill % calc now uses it too.
  - Daytime release: the shared compression buffers (~0.5 GB/array) are returned to the OS while idle in daylight (MADV_REMOVE, with DONTNEED fallback) instead of lingering in swap until night.
  - MADV_DONTFORK on those buffers so GStreamer/RawFrameSaver grandchildren don't carry them.
  - Leak fixes: QueuedPool now owns its Manager and exposes shutdownManager() (called after getResults() in StartCapture/DetectStarsAndMeteors/ExtractStars); Compressor reaps unjoined procs and os._exit()s its run-exit paths so lingering Manager proxy threads don't block shutdown.

  **MultiCam updater (GRMSUpdater.sh)**
  - New `--reboot` (always) / `--reboot-if-needed` (only when pending OS updates) flags; default unchanged. Stops stations cleanly before rebooting.
  - `--reboot-if-needed` uses /var/run/reboot-required, falling back to running-vs-installed kernel.
  - Requires passwordless sudo shutdown or the reboot is skipped with a warning. Stock Raspberry Pi OS already grants the default user blanket NOPASSWD sudo, so no setup is needed there. On hardened setups, grant just the shutdown command:
  ```
echo 'my_username ALL=(ALL) NOPASSWD: /usr/sbin/shutdown' | sudo tee /etc/sudoers.d/rms-reboot > /dev/null
  sudo chmod 0440 /etc/sudoers.d/rms-reboot
```

For instance, this cron job checks for an RMS update daily, updates and restarts RMS if needed, and additionally reboots the machine if a reboot is required to complete an OS update: 
```
0 19 * * * /home/<username>/source/RMS/Scripts/MultiCamLinux/GRMSUpdater.sh  --reboot-if-needed
```


dvida:
Reviewed this thoroughly — the hard parts are all correct: MADV_DONTFORK ordering (set in `BufferedCapture.run()` before `initVideoDevice()` and the `RawFrameSaver` fork; and `RawFrameSaver` maps the *separate* `shared_raw_array`, not `array1/array2`, so it's unaffected), the night→day MADV_REMOVE timing (after `releaseResources()`, buffers idle, Compressor parked in its wait loop), the child-only `os._exit(0)` in `Compressor.run()` (runs after FF writes + `detector.addJob()` + `run_exited.set()`; parent collects via the shared Manager queue), and `QueuedPool.shutdownManager()` ordering (after `getResults()` in all three sites, Compressor stopped first, fresh pool per `runCapture`, None-guarded against double-shutdown). `gst_queue_size` clamping and the buffer-fill recalc are consistent.

I pushed one fix directly to the branch (afd3d96e):

**Daytime buffer-release logged success unconditionally.** Unlike the DONTFORK block (which gates on `ret == 0`), the MADV_REMOVE block didn't check the syscall result and logged `"Released … back to OS"` even when both `MADV_REMOVE` and the `MADV_DONTNEED` fallback failed — a false success with no diagnostic. The commit:
- adds a module-level `madviseArray(arr, advice)` helper (resolves `libc.madvise` once instead of re-binding it at three call sites),
- rounds the start address **down to a page boundary** and extends the length, since `arr.ctypes.data` is a view into a `multiprocessing.Array` suballocation that isn't guaranteed page-aligned — `madvise` would otherwise return EINVAL (verified: a deliberately misaligned view now succeeds),
- returns `(ok, errno)` so the daytime path counts only buffers actually released and warns with errno on failure; DONTFORK reuses the same helper.

This folds in the page-alignment concern and the unused `use_errno=True` (now actually read).

Remaining items, not blocking — your call:

- **GRMSUpdater early-exit reboot path is not failure-symmetric.** In the no-update branch, `stop_stations; do_reboot; exit 0` runs unconditionally once `reboot_available` passes. If `shutdown -r now` fails between the probe and the call, stations are left stopped and never restarted. The post-update path correctly falls through to restart on failure — worth mirroring here (check `do_reboot`, restart on failure instead of `exit 0`).
- **`reboot_available()` resolves `shutdown` via PATH**, but the sudoers doc example pins `/usr/sbin/shutdown`. On distros where `shutdown` is `/sbin/shutdown`, the NOPASSWD rule may not match the resolved binary and the reboot is silently skipped. Worth a note in the header, or probe the absolute path.
- **`should_reboot` kernel fallback** `latest="$(ls /lib/modules/ | sort -V | tail -1)"`: the `local`/assignment masks failure under `set -e`, and it lists *all* module dirs (leftover headers can outrank the running kernel). Acceptable given the loop-guard, just noting the edge.
- Minor: the PR bundles three independent concerns (config / madvise+leak fixes / updater reboot) — fine to merge as-is, just harder to bisect or revert in isolation.

Cybis320:
Thxs. Addressed all points in last commit.

PR #862, opened 2026-05-06
RawFrameSave: Fix saving YUYV/UYVY frames

novasharper:
When frame has only two channels try to get lumina channel to convert it into a one channel frame. `imwrite` in OpenCV only supports 1, 3, and 4 channel images.


novasharper:
Created against wrong branch

PR #863, opened 2026-05-06
RawFrameSave: Fix saving YUYV/UYVY frames

novasharper:
When frame has only two channels try to get lumina channel to convert it into a one channel frame. `imwrite` in OpenCV only supports 1, 3, and 4 channel images.


novasharper:
It looks like the copilot review did not complete for some reason?

novasharper:
@dvida Could you try re-scheduling the copilot review?

Also, is there a different ref that I should try merging to? In the PR status block, I see `Merging is blocked. Cannot update this protected ref`

Issue #873, opened 2026-06-02
RMS crashes during external-script invocation if attempting to log stdout to the RMS log


markmac99:
## Problem Statement
RMS logger crashes fatally  with a stack overflow if an external script attempts to print a message on stderr.   

``` bash
2026/06/02 13:25:24-INFO-RunExternalScript-line:84 - Starting function "<function rmsExternal at 0xa18f7ed0>" from external script "<module 'ukmonPostProc' from '/home/pi/source/ukmon-pitools/ukmonPostProc.py'>" with logging inhibited
2026/06/02 13:25:30-INFO-ukmonPostProc-line:133 - about to import extl module

Fatal Python error: Cannot recover from stack overflow.

Current thread 0xb6fceb40 (most recent call first):
  File "/usr/lib/python3.7/logging/__init__.py", line 619 in format
  File "/usr[29.7%] Uploading: UK001L_20260601_210025_956563_imgdata.tar.bz2 (1.19 MB/4.01 MB) @ 809.72 KB/s - 4 sec remaining
/lib/python3.7/logging/__init__.py", line 880 in format
  File "/usr/lib/python3.7/logging/__init__.py", line 1034 in emit
  File "/usr/lib/python3.7/logging/__init__.py", line 905 in handle
  File "/usr/lib/python3.7/logging/__init__.py", line 1599 in callHandlers
  File "/usr/lib/python3.7/logging/__init__.py", line 1529 in handle
  File "/usr/lib/python3.7/logging/__init__.py", line 1519 in _log
  File "/usr/lib/python3.7/logging/__init__.py", line 1449 in log
  File "/home/pi/source/RMS/RMS/Logger.py", line 97 in write
  File "/usr/lib/python3.7/logging/__init__.py", line 1037 in emit
  File "/usr/lib/python3.7/logging/__init__.py", line 905 in handle
  File "/usr/lib/python3.7/logging/__init__.py", line 1599 in callHandlers
  File "/usr/lib/python3.7/logging/__init__.py", line 1529 in handle
  File "/usr/lib/python3.7/logging/__init__.py", line 1519 in _log
  File "/usr/lib/python3.7/logging/__init__.py", line 1449 in log
  File "/home/pi/source/RMS/RMS/Logger.py", line 97 in write

(above lines repeated many times)
```

[...]

PR #874, opened 2026-06-02
Bugfix for #873, logger crashes in external script

markmac99:
This is a workaround for #873 

During external script invocation, a stack overflow crash can arise in RMS.Logger, if the RMS log handlers have been closed before the external script is fully complete.


dvida:
Awesome - would you be ok if we merge this into prerelease and then expedite testing and merging to master? We have a few weeks to test before any major meteor showers show up.

markmac99:
sure, whatever works best. I'm obviously keen to get it live as i have had to manually run the uploads on UK001L for the last three weeks while i debug this!

PR #876, opened 2026-06-04
Delete PyGObject from requirements as it is NOT installed via pip.

Cybis320:
PyGObject is NOT installed via pip. On Linux it comes from the system package (python3-gi).
Pip-building it fails on Windows/macOS and GStreamer capture isn't used there anyway (BufferedCapture falls back to OpenCV when gi is absent).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,8 +27,6 @@ pyclean>=2.7.6; python_version >= "3.5"
 wheel>=0.30.0
 setuptools>=40.0.0
 ifaddr; sys_platform =="win32"
-# PyGObject installed via system package (python3-gi) on Linux
-PyGObject; python_version >= '3.6' and sys_platform != 'linux'
 rawpy; sys_platform == "win32"
 rawpy; sys_platform == "darwin"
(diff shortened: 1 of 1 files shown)


dvida:
If I understand what you're saying correctly, on linux when opencv is installed in a conda env, it will pull the right package inside the env, without having to install the system package (which it can't without sudo)?

Cybis320:
Right - since the last major update to master, GStreamer on Linux comes from system packages (apt's python3-gi plus the gstreamer1.0-* plugins). That works because the standard installer creates the venv with --system-site-packages, so the venv can see those system-installed packages.

Conda environments are isolated and don't inherit system site-packages, so that mechanism doesn't apply. For conda you have to pull  the whole stack from conda-forge instead. The README covers this, I think:
```
conda install -y -c conda-forge pygobject
conda install -y -c conda-forge gstreamer==1.22.3 gobject-introspection gst-libav gst-plugins-bad gst-plugins-base gst-plugins-good
gst-plugins-ugly
```
Caveat: I don't use conda myself and haven't tested this.

dvida:
Alright, let's go with this and see what happens!

Issue #882, opened 2026-06-13
Observation summary fix race condition


g7gpr:
On higher performance computers a race condition can cause the value from the start of an observation to fail to get recorded in the observation summary.

PR #883, opened 2026-06-13
Combine development work on ObservationSummary

g7gpr:
This PR is in response to Issue #882. 

On high performance computers a race condition caused the values recorded at the start of an observation session not to be recorded. This PR fixes this by recording all the values in a dict, then stored as a json in the captured files directory. This information is then written into a sqlite database with dynamic schema adjustment. 

Work by others to add additional keys has also been incorporated. These keys are 

```
kht_wrapper_count                         : 0
traceback_count                           : 0
```
kht_wrapper_count counts the occurences of kht wrapper errors, which might indicate a badly linked kht so file.
traceback_count counts the number of tracebacks in the log, useful for detecting execution anomalies.

```
camera_information                        : 50H20L
camera_firmware_build_date                : 2018-11-26 16:11:24
camera_firmware_version                   : V4.03.R12.00031520.11012.042800.00200
```

And also reports on the camera firmware. 

The ObservationSummary is rendered is a png file for display in the weblog, for future implementation; in this case 

```
AU000A_20260613_100215_850897_observation_summary.png
```
Which looks like

AU000A_20260613_100215_850897_observation_summary

### DIFF
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -1445,6 +1445,13 @@ class BufferedCapture(Process):
                     self.device = cv2.VideoCapture(self.config.deviceID, cv2.CAP_V4L2)
                     self.device.set(cv2.CAP_PROP_CONVERT_RGB, 0)
 
+                    try:
+                        log.info("Getting observation summary dict")
+                        addObsParam(getObservationSummaryDict(None, config=self.config), "media_backend", self.video_device_type)
+                        log.info("Got observation summary dict")
+                    finally:
+                        pass
+
                     return True
                 
                 except Exception as e:
@@ -1457,6 +1464,14 @@ class BufferedCapture(Process):
             elif (self.config.media_backend == 'cv2') or self.media_backend_override:
                 log.info("Initialize OpenCV Device.")
                 self.device = cv2.VideoCapture(self.config.deviceID)
+
+                try:
+                    log.info("Getting observation summary dict")
+                    addObsParam(getObservationSummaryDict(None, config=self.config), "media_backend",
+                                self.video_device_type)
+                    log.info("Got observation summary dict")
+                finally:
+                    pass
 
                 return True
 
--- a/RMS/BufferedCapture.py
+++ b/RMS/BufferedCapture.py
@@ -42,7 +42,7 @@ import json
 
 from RMS.Misc import obfuscatePassword
 from RMS.Routines.GstreamerCapture import GstVideoFile, getStructureValue
-from RMS.Formats.ObservationSummary import getObsDBConn, addObsParam
+from RMS.Formats.ObservationSummary import addObsParam, getObservationSummaryDict
 from RMS.RawFrameSave import RawFrameSaver
 from RMS.Misc import RmsDateTime, mkdirP, UTCFromTimestamp
 from RMS.Formats import FTfile, FTStruct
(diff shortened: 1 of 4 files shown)


g7gpr:
Tested for short duration capture on au0004. Testing overnight on au000a->f

g7gpr:
AU000A_20260613_102214_111917_observation_summary
AU000C_20260613_102532_784405_observation_summary
AU000D_20260613_103222_582331_observation_summary
AU000E_20260613_102525_852829_observation_summary
AU000F_20260613_103231_796250_observation_summary
AU000K_20260613_103500_839745_observation_summary

dvida:
Reviewed and pushed fixes in 0dce1562 (2 files, +174/−112). Summary of changes:

### Blocker / crash + hang
- **Infinite loop (polar night).** `getObservationDurationNightTime` looped `while isinstance(ephemeris_start_time, bool)`, but `captureDuration` returns a bool regardless of `start_time` during polar night (`NeverUpError`), so it spun forever at high latitudes. Now a bounded 24 h backoff with a fallback to the passed start time.
- **Crash + wrong-night write.** `BufferedCapture.initVideoDevice` used `getObservationSummaryDict(None, config=self.config)` to guess the latest captured dir, which returns `{}` when none matches → `saveObservationSummaryDict(night_dir=None)` → `getRMSStyleFileName(None, …)` `AttributeError` (and could target the previous night). Now uses `self.night_data_dir` (already passed to the constructor), guarded against `None`.

### Race / data loss (issue #882 class)
- **`saveObservationSummaryDict` is now concurrency-safe.** The capture child process (media_backend), the main process (start-of-session values) and Reprocess all read-modify-write the same working JSON. Added: `None` guard, an exclusive `fcntl` lock (degrades gracefully where unavailable), **merge with the on-disk JSON** (so concurrent writers adding distinct keys don't clobber each other), and an atomic temp-file + `os.replace` write.
- **No more silent deletion.** `getObservationSummaryDict` now backs an unparseable JSON up to `*.corrupt` instead of `os.remove`.

### Robustness / cleanup
- `writeToPNG`: falls back to the default PIL font when DejaVu is absent, honours the `font_size` arg (was hardcoded 16), and the whole render is wrapped so a failure can't break finalization.
- `media_backend` recorded as an explicit `gst`/`v4l2`/`cv2` label (v4l2 was mislabelled "cv2"); `video_device_type` deliberately left unchanged since downstream open/frame-skip logic keys off `== "cv2"` for all OpenCV-backed devices including v4l2.
- Turned off the `debug=True` SQL/column dumps to stdout in `finalizeObservationSummary`.
- Deduped the `serialize` `ordering` list (removed duplicate `media_backend` / `star_catalog_file` lines), used a solar day (24 h) for `days_since_last_detection`, parameterized + guarded SQL identifiers, added a `force_delete` existence check, removed the dead `getNextStartTime` (only referenced the old `records` table), and dropped the Python 2 import fallback.

Verified: all three modules compile; functional tests confirm the merge keeps concurrent keys, atomic write leaves no `.tmp`, the `None` guard doesn't raise, and the polar-night call returns in ~0.04 s instead of hanging.

dvida:
Dave, could you please test with the new changes that I've pushed?

g7gpr:
Will check by Canada Friday noon.

g7gpr:
The fix for the polar night condition was not needed, but it does no harm. Looks good. Running on AU0006 overnight. I'll confirm in the morning if it worked well.

g7gpr:
AU000A has been running this version for a couple of days, so I'm satisfied it is all OK.

PR #884, opened 2026-06-14
Add video_scale and video_crop parameters

walesto:
New parameters to limit original video size (by scale or/and by crop) for the purpose of capture and further processing while original raw video will still be saved in full resolution (if raw_video_save is true).


dvida:
Thank you - I made some small changes, could you test the latest version please?

walesto (commented):
Latest changes verified in below scenarios:
(source video size/video_scale parameters/video_crop parameters/config width,heigh/result
- 1920x1080/1280x720//1280x720/ok
- 1920x1080///1920x1080/ok
- 1920x1080///1280x720/warning for mismatch (after today commit) - ok
- 1920x1080/1280x720//1920x1080/warning for mismatch - ok
- 2596x1944/1296x977/bottom=1/1296x976/ok

PR #886, opened 2026-06-17
Add planetary overlay support to SkyFit2

AstroEloy:
This PR adds an optional planetary overlay to SkyFit2.

Features:
- Toggle planet display with Ctrl+Shift+P.
- Compute planet positions using PyEphem.
- Dynamically update positions with image time.
- Filter planets by limiting magnitude.
- Reuse the existing catalog star projection, filtering and rendering pipeline.

Planets are treated similarly to catalog stars and are rendered using their visual apparent magnitudes. The overlay is disabled by default and does not affect the existing SkyFit2 functionality.

PyEphem is already an RMS dependency, so no additional packages are required.

### DIFF
--- a/Utils/SkyFit2.py
+++ b/Utils/SkyFit2.py
@@ -16,6 +16,7 @@ import random
 import copy
 
 import numpy as np
+import ephem
 import matplotlib.pyplot as plt
 import matplotlib.gridspec as gridspec
 try:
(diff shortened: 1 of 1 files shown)


dvida:
Already completed in the prerelease branch.

AstroEloy:
Thanks! I noticed that the prerelease implementation uses Astropy, which introduces a new dependency. My implementation uses PyEphem, which is already an RMS dependency, so no additional packages are required.

Feel free to close this PR if the functionality is already covered in the prerelease branch.

Issue #889, opened 2026-06-18
LogArchiver: getTimeOfLastLogEntry crashes on multi-line log entries (e.g. pprint'd camera config), spams ERROR traceback


Cybis320:
getTimeOfLastLogEntry() reads only the last physical line of a log file and assumes that it is an RMS log and that it begins with a YYYY/MM/DD HH:MM:SS timestamp. When the newest log of any type ends in a multi-line entry whose final line has no timestamp prefix, strptime raises ValueError, a full traceback is logged at ERROR level on every LogArchiver run, and the function silently falls back to year 2000.

For instance, this I can be triggered by camControl logs: Utils/CameraControl.py writes camera config blocks with log.info(pprint.pformat(...)), which emits multiple lines. The last line of such a log looks like:
```  
   'WorkDay': 62}
```
**Observed log output**

```
2026/06/17 09:01:04-ERROR-LogArchiver-line:58 - Traceback (most recent call last):
  File "/home/ops/source/RMS/Utils/LogArchiver.py", line 56, in getTimeOfLastLogEntry
    last_log_entry_time_object = datetime.datetime.strptime(last_log_time_string, "%Y/%m/%d %H:%M:%S")
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_strptime.py", line 554, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_strptime.py", line 333, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data " 'WorkDay': 62}\n" does not match format '%Y/%m/%d %H:%M:%S'
```

  
**Root cause**
https://github.com/CroatianMeteorNetwork/RMS/blob/1942cebd158fe76c60f84d30b831127aaddc9c0d/Utils/LogArchiver.py#L36-L63

Taking the final line is unsafe — any log entry that spans multiple lines (pprint'd dict, traceback, etc.) breaks the timestamp assumption.

  
**Suggested fixes**
  
Remove the diagnostic — since its result is discarded, drop the getTimeOfLastLogEntry call + its log line (and the now-unused function). Lowest-risk; deletes fragile dead code.

[...]

PR #890, opened 2026-06-18
Remove redundant code from LogArchiver which caused errors

g7gpr:
This PR is in response to issue #889, and removes code which was intended to detect the latest entry in the log file, but could not handle multiple line entries. 

This code was only used for a log entry, which was of limited use outside development work, so no functionality is lost.


g7gpr:
Currently on test on AU0004

g7gpr:
Tested on AU0004, now running overnight.

PR #894, opened 2026-06-19
Allow specifying kernel UDP buffer size

Cybis320:
Many stations have been observed to silently drop UDP payload during bursts due to kernel UDP buffers being too small. This PR increases the default size from 1 MB to 16 MB and makes it configurable.

### DIFF
--- a/.config
+++ b/.config
@@ -104,6 +104,12 @@ device: rtsp://192.168.42.10:554/user=admin&password=&channel=1&stream=0.sdp
 ; provides more stable connectivity, but TCP has a longer track record.
 protocol: tcp
 
+; Per-socket UDP receive buffer size in bytes for the gst backend (rtspsrc).
+; Larger values give RTP bursts more room and prevent UDP RcvbufErrors ->
+; dropped frames. Requires net.core.rmem_max >= this value (run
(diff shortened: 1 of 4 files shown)


dvida:
Pushed 649f41da with two small follow-ups from review:

- **`Scripts/UpdateBuffers.sh`** — the warning/note strings and header docs still hardcoded the old `512KB`/`1MB` numbers after the constants moved to `MIN_RECOMMENDED=1MB` / `RECOMMENDED_SIZE=16MB`, so the script printed thresholds that didn't match what it actually checks. Now the messages use `human_readable $MIN_RECOMMENDED` / `human_readable $RECOMMENDED_SIZE` and the doc comments read 1MB min / 16MB recommended.
- **`RMS/BufferedCapture.py`** — fixed two comment typos: `can overflows` → `can overflow`, and dropped the stray `net` in `net UDP RcvbufErrors`.

Note for operators (not in this diff): existing installs that ran the old `UpdateBuffers.sh` have `rmem_max=1MB`, so the new 16MB default will silently clamp back to 1MB until they re-run the script. Worth flagging in install/upgrade notes.

Cybis320:
Looks good to me.

Issue #739, opened 2025-11-02
hardcoded NTP server in ObservationSummary


markmac99:
timestampFromNTP in ObservationSummary contains a hard-coded NTP server name.

This could be an issue on managed networks such as schools, univerities and other academic institutions which do not permit unfettered outbound connections without prior agreement. In such cases, we have found we generally have to provide a complete list of IP addresses, ports, protocols and services to the org's IT security team in advance, so they can configure a suitable set of f/w rules or in some cases create a virtual subnet to isolate the RMS station as per their organizational rules.  

I only noticed this because I was looking at the console while running a reprocess and an NTP timeout error popped up (and was visible because of bug #738 which prevented the usual overenthusiastic logging :) ) My own network blocks outbound NTP except from our in-house ntp server hence the attempt failed.

I suggest the NTP server should be an optional parameter in the config file so that orgs can point it at their own server if needed.  Happy to work on that and create a PR for the next release. In 99% of cases it won't be required of course. 

We should also document the NTP requirement in the Wiki, so that we can avoid unexpected challenges from IT security teams. I don't think there's currently anywhere on the GMN wiki that lists the IP address, port and protocol requirements for RMS so i'd be happy to add something. We have something for UKMON [here](https://github.com/markmac99/ukmon-pitools/wiki/networking-for-schools) so i can quickly plagiarise that!

markmac99:
Fix proposed via PR #910

markmac99:
closed as included in David's current work

PR #898, opened 2026-06-21
Handle ObservationSummary dict is None

g7gpr:
Avoid iterating over ObservationSummary dict when dict is None

### DIFF
--- a/.config
+++ b/.config
@@ -97,7 +97,7 @@ reboot_lock_file: .reboot_lock
 ; --------------
 
 ; If neither chrony nor NTP are available timeserver configured below will be
-; used to as a reference to calculate the time error.
+; used as a reference to calculate the time error.
 
 time_server: time.cloudflare.com
 
--- a/.config
+++ b/.config
@@ -93,6 +93,13 @@ reboot_after_processing: true
 ; (i.e. ~/RMS_data).
 reboot_lock_file: .reboot_lock
 
+; Time server
+; --------------
+
+; If neither chrony nor NTP are available following timeserver will be used to
+; calculate the time error for the Observation Summary report.
+
+time_server: time.cloudflare.com
 
 
 [Capture]
--- a/.config
+++ b/.config
@@ -58,7 +58,7 @@ network_name: none
 ; Camera group (e.g. a camera cluster or a location with multiple cameras)
 camera_group_name: none
 
-# timeserver to use for ObservationSummary - leave blank or omit to use default 'time.cloudflare.com'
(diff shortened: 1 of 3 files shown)


dvida:
@g7gpr Could you test this latest version with my modifications?

g7gpr:
Incoporated pr #910, from issue #739, with thanks to @markmac99

Retesting overnight on au0004,6,7

g7gpr:
This needed a bit more work to catch the edge case where a key that was not a valid column name was not used to create a column, but then tried to write into a column with that name. Also incorporated features from issue #739. However, the changes by @dvida are good.

g7gpr:
Tested and working on three stations. Ready for further review / merge.

dvida:
Reviewed this — one correctness regression worth flagging, which I've pushed a fix for (98c1eb8).

**`storeDictInDB` silently drops `stationID` (and any mixed-case key).**

```python
dict_filtered_by_columns = {k: v for k, v in d.items() if k in existing_columns}
```

Columns are always created lower-cased (`addRequiredColumns` → `ADD COLUMN {key.lower()}`), so `existing_columns` only ever contains lower-case names. The new filter compares the **original-case** key, so `stationID` (added at `addObsParam(d, "stationID", ...)`) is filtered out and never inserted — the `stationid` column ends up always NULL. Pre-PR code inserted `d.items()` directly and relied on SQLite's case-insensitive column matching, so it stored fine. This is a data-loss regression for the station identifier.

Fix matches case-insensitively and keys the filtered dict by the lower-cased name:
```python
dict_filtered_by_columns = {k.lower(): v for k, v in d.items() if k.lower() in existing_columns}
```
Verified in isolation: old filter drops `stationID`, new one keeps it.

Also ungated the dropped-keys `log.warning` (was behind `debug`), so genuine data loss is visible in normal logs rather than silent.

Everything else checked out: the `timestampFromNTP` 3-tuple change is consistent with its only caller, the `None`-dict guards are sound, and the non-dict-JSON rejection is a good hardening.

g7gpr:
Agree with these changes, penultimate commit introduced an error, which latest commit has corrected.

```
(vRMS) au0004@birch:~/source/RMS$ sqlite3 /srv/rms/RMS_data/AU0004/observation.db -header -column "select commit_date, stationid from observations order by commit_date desc limit 4;"
commit_date      stationid
---------------  ---------
20260629_151259  AU0004   
20260628_020646           
20260628_020646           
20260627_031825  AU0004   
```

I'm testing on au0004,6,7; there should be enough night left to capture a small amount of data.

g7gpr:
All looks good now.

PR #910, opened 2026-06-27
read timeserver from config, if present - see issue #739

markmac99:
Also added some exception handlling to trap cases where unix-style commands like timedatectl and systemctl are not available. or where calls to system functions fail for other reasons. This allows the observationsummary to run cleanly on Windows.


markmac99:
Closing as DR has included it in his work on the Observation Summary

Issue #911, opened 2026-06-28
RMS_Update.sh --force doesn't seem to rebuild kht_wrapper


markmac99:
In principle RMS_Update.sh --force should rebuild all required modules. 
However on UK004X i found that the kht_wrapper was not rebuilt. 
To ensure it was rebuilt i had to run `python setup.py install` but this has other undesirable side effects.
`kht_wrapper` is interesting because its a native DLL that is invoked by `LoadLibrary` in `ctypes`, rather than being Cython. 

Ideally, the --force parameter would make sure kht_wrapper is also rebuilt. 
I therefore propose to create a separate kht_wrapper specific script that can be called from RMS_Update, or provide some other method to compile the C++ .  
One alternative would be to make the package fully standalone and submit it to PyPi with its own build script.

Cybis320:
As a matter of fact, RMS_Update rebuilds all necessary modules either when an RMS update is detected or when the --force flag is used. However, if RMS itself is up to date and the --force flag is not used, a corrupted module will not be rebuilt.  

If you insist on a manual rebuild, please avoid using `python setup.py install` and instead use `pip install -e . --no-build-isolation`.

Note that a manual rebuilt is just as susceptible to corruption as the one performed by RMS_Update.

A corrupted module typically indicates a storage issue or a power interruption.

Why is the kht module more susceptible to corruption? It's the only module loaded by ctypes from a filesystem path, not by Python's import system. The other modules fail loudly and get rebuilt - not kht_wrapper. Any machine that's been crashing accumulates a bad kht artifact that a normal (non-force) update won't replace.

The long-term solution is to replace or repackage kht_module as you suggested. Or more simply, wrap it in Cython like the other modules: PR #912

markmac99:
As noted, on the machine in question RMS_Update --force did not recreate the KHT wrapper, irrespective of what it ought to have done!  I'm probably missing something but i cant quite see how pip can build this module. 

In any events if you have a cythonized version in the works thats fine. It turns out to be trivial to create a standalone script to build it, which I'll submit as a PR in case it takes a while to cythonize.  

[...]

PR #912, opened 2026-06-28
Build KHT as a Cython extension instead of a ctypes-loaded library

Cybis320:
Addresses issue #911. KHT was the only native module loaded by ctypes from a filesystem-searched path, with no ABI check — so a stale or partial build loaded silently and failed later with undefined symbol: kht_wrapper.

This compiles the KHT C++ sources into a regular Cython module (RMS.Routines.Kht), imported like every other native module. A bad build now fails fast with a clear ImportError and gets rebuilt.

No behavior change — output and performance is identical to the previous ctypes path.

### DIFF
--- a/.config
+++ b/.config
@@ -595,15 +595,6 @@ line_min_dist: 50
 ; Stripe width around the line for centroiding and photometry
 stripe_width: 28
 
-; Directory where binaries are built
-kht_build_dir: build
-
-; Name of the KHT binary
-kht_binary_name: kht_module
-
-; Extension of the KHT binary
-kht_binary_extension: so
-
 ; 3D matched filter parameters
 ; ----------------------------
 
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,7 @@ RMS/Routines/BinImageCy.c
 RMS/Routines/DynamicFTPCompressionCy.c
 RMS/Routines/Grouping3Dcy.c
 RMS/Routines/MorphCy.c
+RMS/Routines/Kht.cpp
 Utils/SaturationTools.c
 
 # Large GMN catalog that's too big for GitHub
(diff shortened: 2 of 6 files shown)


markmac99:
Let me know when there's something i can help test. I separately went down the route of creating a simple build script for kht_wrapper which does the trick and is cross-platform, but cythonising would be better.

Cybis320:
It works perfectly:
2026/06/29 11:35:14-INFO-DetectStarsAndMeteors-line:291 - TOTAL: 202 detected meteors.

dvida:
Can you do a 1:1 comparison between the old and new method across many cameras? Thanks!

Cybis320:
> Can you do a 1:1 comparison between the old and new method across many cameras? Thanks!

Yes, I can. Results will be identical- there is no change in the underlying code.

The KHT algorithm itself is byte-for-byte unchanged. Same .cpp files, same kht_wrapper, same compile-flag. Compiled on the same machine, the standalone extension and the Cython-bundled one produce identical machine code. The Cython wrapper does no math — it just hands the same pointers to the same function. So same input bytes -> bit-identical output. 

The image fed to KHT went from np.byte (int8) to np.uint8 , but the C side reads unsigned char either way, so the bytes it sees are identical.

There's no plausible "runs but gives subtly different detections" path.

I'll report back.

Cybis320:
Hm, there is a complication to perform clean A/B, code upstream of kht is non-deterministic...

Cybis320:
there is an unseeded np.random.choice in Detection. Seems like it should be seeded as all other randoms in the codebase are seeded. I'll make a PR for that and rerun A/B with seeded.

Cybis320:
ok, results are now byte-identical, performance difference are within noise threshold.

PR #914, opened 2026-06-30
SkyFit2: roll-aware WASD panning + pointing indicator

dvida:
## What

Two related SkyFit2 usability changes for adjusting the pointing.

### Roll-aware WASD panning
Previously WASD changed the reference azimuth/altitude directly, so when the camera frame was rolled the view moved in confusing, rotated directions, and panning toward the zenith got stuck at the pole.

- WASD now move the view in the apparent **on-screen** directions (W=up, S=down, A=left, D=right) regardless of camera roll or FOV size.
- The step advances the pointing along a **great circle** (rotating the pointing vector on the sphere), so every step is the same angular size anywhere on the sky and the pointing **crosses the zenith cleanly** instead of sticking — the horizon rotation flips ~180° as it goes over.
- Step size is **FOV-relative** (a fraction of the FOV diagonal), so a press feels equally responsive on a narrow lens and an all-sky camera; still tunable with `+`/`-`.
- Away from the poles the rotation w.r.t. horizon is held fixed (azimuth panning stays level); within ~one step of a pole the celestial roll is held instead, which is what lets it cross the pole.

New helper `screenNudgeToAzAltDelta` and methods `panStepDeg` / `nudgeReferenceScreen`.

### Pointing indicator HUD
A glyph pinned at the optical-axis plus (`PointingIndicator`), evaluated at the optical axis:
- **Zenith arrow** — points to the zenith on screen; its angle conveys the camera roll.
- **Elevation notch** — slides along the arrow (centre = horizon, tip = zenith).
- **Az/Alt readout** — apparent azimuth and altitude; decimal precision scales with the FOV (integer for wide/all-sky, finer for narrow).
- **Horizon bar** through the optical axis that triple-serves as: a horizon/roll indicator (lies along the E–W horizon), a pan **step-size gauge** (length = one WASD step, scaling with zoom), and an **azimuth compass** (middle = N, little W/E notches, a notch sliding to the current azimuth).

New helper `fovCentreZenithDirection`; wired into `updateLeftLabels` / `updateDistortionCenterMarker` / `changeMode` and the view range-change signal so it tracks pointing changes and zoom.

## Files
- `RMS/Astrometry/ApplyAstrometry.py` — `screenNudgeToAzAltDelta`, `fovCentreZenithDirection`
- `RMS/Routines/CustomPyqtgraphClasses.py` — `PointingIndicator`
- `Utils/SkyFit2.py` — `panStepDeg`, `nudgeReferenceScreen`, `updatePointingIndicator`, WASD handlers, hooks, imports

## Testing
Numerically verified (offscreen, sample platepars): equal 2.0° angular steps at altitudes up to 89.95°; clean zenith crossing under repeated W (climb → flip ~180° → descend) with no sticking; level azimuth panning away from the poles; pan direction stays screen-correct across rolls of 0/58/122/165/−123°; FOV-relative step scaling; indicator rendering for upright/rolled and N/E/W azimuth cases. Not yet exercised in a live interactive GUI session.

### DIFF
--- a/RMS/Astrometry/ApplyAstrometry.py
+++ b/RMS/Astrometry/ApplyAstrometry.py
@@ -574,6 +574,192 @@ def rotationWrtHorizon(platepar):
 
 
 
+def screenNudgeToAzAltDelta(platepar, screen_dx, screen_dy, key_increment, screen_y_sign=-1, h_px=10):
+    """ Map a requested on-screen nudge direction to a change in the apparent reference azimuth and
+        altitude, so that pressing a pan key always moves the view in the apparent screen direction
+        regardless of the camera roll (rotation w.r.t. horizon) or FOV size.
+
+        The screen direction is sampled at the FOV centre as a 3D tangent on the sky, and the pointing
+        is then advanced by rotating its unit vector along the corresponding great circle by exactly
+        key_increment degrees. This automatically absorbs the camera roll and lens distortion, keeps
+        every step an equal angular size regardless of pointing, and carries the pointing cleanly over
+        the zenith (no alt = 90 deg singularity). The raw-space rotationWrtHorizon angle is only used as
+        a fallback when the sampling is degenerate.
+
+    Arguments:
+        platepar: [Platepar object] Input platepar.
+        screen_dx: [float] Requested screen direction X (+1 = right, -1 = left, 0 = none).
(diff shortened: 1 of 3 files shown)


dvida:
Screenshot from 2026-06-29 23-20-21

Pointing indicator example

PR #917, opened 2026-07-02
Fix compile args

Cybis320:
Follow-up to #916. The .config [Build] compile flags were only respected by 4 of the 8 Cython modules, and only on the pyximport fallback path — setup.py ignored them entirely and hardcoded -O3 for Kht, which MSVC doesn't understand (warning D9002).

Changes:

- New RMS/CompileArgs.py: single source of truth for platform compile args, parsed from .config's [Build] section with safe defaults. Stdlib-only, so setup.py can import it at build time before any dependencies are installed (ConfigReader couldn't be used for this — it pulls in matplotlib and friends).
- setup.py: all 8 extensions now use the shared flags (fixes the MSVC D9002 warning; Linux/macOS get -O3 on every module instead of just Kht).
- All 8 Cython modules now have an identical .pyxbld using the shared flags — including new ones for BinImageCy, DynamicFTPCompressionCy, and SaturationTools, which previously had none. The fallback rebuild also no longer imports matplotlib just to read three flags.
- ConfigReader delegates to the shared module. This also fixes a real bug: choosePlatform matched 'arm' in os.uname()[4], so Apple Silicon Macs got the RPi's 32-bit NEON gcc flags, which clang rejects — every config-reading pyxbld rebuild was broken on M-series Macs.
- Dropped extra_link_args from the pyxblds (compile flags do nothing at link time with gcc and emit LNK4044 noise on MSVC).

Not tested: Windows/MSVC — @markmac99, a python -m RMS.Detection run would be appreciated. Note MSVC now gets -Wall (= /Wall) from win_pc_weave; if that's too noisy it's a one-line .config default change.


markmac99:
@Cybis320 tested on windows, rebuilds Kht and runs detection successfully.  

```
(RMS) PS C:\dev\testing\RMS\UK0006\CapturedFiles\mini> python -m RMS.Detection . -c ./.config -d -o .
...
running build_ext
cythoning C:\dev\meteorhunting\RMS\RMS\Routines\Kht.pyx to C:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\pyrex\RMS\Routines\Kht.cpp
building 'RMS.Routines.Kht' extension
"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\bin\HostX86\x64\cl.exe" /c /nologo /O2 /W3 /GL /DNDEBUG /MD -IC:\dev\meteorhunting\RMS\Native\Hough -IC:\users\mark\miniconda3\envs\RMS\lib\site-packages\numpy\core\include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\ATLMFC\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\VS\include" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\um" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\shared" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\winrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\cppwinrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um" /EHsc /TpC:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\pyrex\RMS\Routines\Kht.cpp /FoC:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\pyrex\RMS\Routines\Kht.obj -Wall
Kht.cpp
....
Time for the whole directory: 85.81026077270508
Detected meteors: 4
Shutting down logging...
Logging shutdown complete.

```

The C++ compiler emits a lot of other warnings which we could probably suppress such as warning C4711 "function xxx selected for inline expansion" "C4514: 'accumulator_t::bins': unreferenced inline function has been removed " but these are just advisories. This is an interesting one - warning C4127: conditional expression is constant. I'll add this to the list of things i'd like to silence in the cython compilation !

Cybis320:
This is expected if the .config build args in RMS root has `win_pc_weave: -Wall`
Can you you try with just `win_pc_weave:` (no args)?

> @Cybis320 tested on windows, rebuilds Kht and runs detection successfully.
> 
> ```
> (RMS) PS C:\dev\testing\RMS\UK0006\CapturedFiles\mini> python -m RMS.Detection . -c ./.config -d -o .
> ...
> running build_ext
> cythoning C:\dev\meteorhunting\RMS\RMS\Routines\Kht.pyx to C:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\pyrex\RMS\Routines\Kht.cpp
> building 'RMS.Routines.Kht' extension
> "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\bin\HostX86\x64\cl.exe" /c /nologo /O2 /W3 /GL /DNDEBUG /MD -IC:\dev\meteorhunting\RMS\Native\Hough -IC:\users\mark\miniconda3\envs\RMS\lib\site-packages\numpy\core\include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\ATLMFC\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\VS\include" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\um" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\shared" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\winrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\cppwinrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um" /EHsc /TpC:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\pyrex\RMS\Routines\Kht.cpp /FoC:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\pyrex\RMS\Routines\Kht.obj -Wall
> Kht.cpp
> ....
> Time for the whole directory: 85.81026077270508
> Detected meteors: 4
> Shutting down logging...
> Logging shutdown complete.
> ```
> 
> The C++ compiler emits a lot of other warnings which we could probably suppress such as warning C4711 "function xxx selected for inline expansion" "C4514: 'accumulator_t::bins': unreferenced inline function has been removed " but these are just advisories. This is an interesting one - warning C4127: conditional expression is constant. I'll add this to the list of things i'd like to silence in the cython compilation !

RMS/CompileArgs.py
markmac99, line 1: getCompilerArgs retrieves compile args from the *default* RMS config file ie in ~/source/RMS. This is generally okay, but in a multi-cam environment the user might expect them to be retrieved from the camera-specific file. Also when RMS moves to multi-cam layout for single-cam stations, it could be very confusing.  

I'm not sure there's an obvious fix since at compile-time RMS isn't initialised so this is more an FYI. It caught me out as i was testing with the -c flag and expecting the named config file to be used for compilation!

markmac99, line 1: I suggest changing win_pc_weave  in the .config file to 

/Wall /wd4711 /wd4514 /wd4710/wd5045 

which will silence the following compiler warnings:
function xxx selected for automatic inline
function zzz cannot be inlined
inline function yyy optimised out as unused
spectre mitigation inserted

markmac99 (approved):
All fine. I made some suggestions about changing the Windows compiler flags to silence irrelevant advisories.

RMS/CompileArgs.py
Cybis320: Right, that's always been the case, and I've always struggled with this. In MCL setup, it does make some sense as we can't have different build args per stations. I think the long-term solution is to have a common config in the Stations root folder. (for another PR).

markmac99:
Actually upon reflection i think we should remove all the flags for the Windows compile.  ie pc_weave should be blank and -Wall should be removed from the default args in CompilerArgs.py. 

Cythonised code generates A LOT of warnings with MSVC, mostly advisories or warning about cythonised code that isn't exactly following C++ 14 standards.  If i remove /Wall, then i get a reasonable set of real warnings about genuine issues -as shown below
``` bash
"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\bin\HostX86\x64\cl.exe" /c /nologo /O2 /W3 /GL /DNDEBUG /MD -IC:\dev\meteorhunting\RMS\Native\Hough -IC:\users\mark\miniconda3\envs\RMS\lib\site-packages\numpy\core\include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\ATLMFC\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\VS\include" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\um" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\shared" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\winrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\cppwinrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um" /EHsc /TpC:\dev\meteorhunting\RMS\Native\Hough\linking.cpp /FoC:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\dev\meteorhunting\RMS\Native\Hough\linking.obj
linking.cpp
C:\dev\meteorhunting\RMS\Native\Hough\linking.cpp(143): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
C:\dev\meteorhunting\RMS\Native\Hough\linking.cpp(145): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
```
whereas with /Wall i get literally 1000 lines of advisories - see attached build log. These don't provide any useful info to end-users and often cause alarm. 

[build.log](https://github.com/user-attachments/files/29616190/build.log)

RMS/CompileArgs.py
Cybis320: I looked into this and went the other way: win_pc_weave is now empty, so MSVC builds with its distutils defaults (/O2 /W3). Reasoning:

Everything in the noise flood is either level 4 or off-by-default: C4711, C4514, C4710, and C5045 only exist under /Wall, and C4127 (the one you specifically wanted gone) is level 4 — so /W3 silences all of them without any /wd list. Notably, /Wall /wd4711 /wd4514 /wd4710 /wd5045 would still let C4127 through, plus other /Wall-only noisemakers that Cython-generated code triggers en masse (C4820 struct padding, C4365 conversions, C4668 from the Windows headers). Making /Wall tolerable on generated code ends up being a whack-a-mole suppression list that grows as MSVC adds new advisories.

What survives at /W3 is a small set of arguably-genuine warnings (C4244/C4267 integer narrowing, C4996 deprecations) rather than hundreds of advisories.

If you want full diagnostics for development, an explicit /Wall (plus your own /wd list) in your local config still works.

Cybis320:
The issue now is that this PR has no win_pc_weave args (good), but all existing .config have `win_pc_weave: -Wall`. Users will have to manually update their config I suppose.

> Actually upon reflection i think we should remove all the flags for the Windows compile. ie pc_weave should be blank and -Wall should be removed from the default args in CompilerArgs.py.
> 
> Cythonised code generates A LOT of warnings with MSVC, mostly advisories or warning about cythonised code that isn't exactly following C++ 14 standards. If i remove /Wall, then i get a reasonable set of real warnings about genuine issues -as shown below
> 
> ```shell
> "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\bin\HostX86\x64\cl.exe" /c /nologo /O2 /W3 /GL /DNDEBUG /MD -IC:\dev\meteorhunting\RMS\Native\Hough -IC:\users\mark\miniconda3\envs\RMS\lib\site-packages\numpy\core\include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include -IC:\users\mark\miniconda3\envs\RMS\include -IC:\users\mark\miniconda3\envs\RMS\Include "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\ATLMFC\include" "-IC:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\VS\include" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\um" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\shared" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\winrt" "-IC:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\cppwinrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um" /EHsc /TpC:\dev\meteorhunting\RMS\Native\Hough\linking.cpp /FoC:\Users\Mark\.pyxbld\temp.win-amd64-cpython-38\Release\dev\meteorhunting\RMS\Native\Hough\linking.obj
> linking.cpp
> C:\dev\meteorhunting\RMS\Native\Hough\linking.cpp(143): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> C:\dev\meteorhunting\RMS\Native\Hough\linking.cpp(145): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> ```
> 
> whereas with /Wall i get literally 1000 lines of advisories - see attached build log. These don't provide any useful info to end-users and often cause alarm.
> 
> [build.log](https://github.com/user-attachments/files/29616190/build.log)

markmac99:
> The issue now is that this PR has no win_pc_weave args (good), but all existing .config have `win_pc_weave: -Wall`. Users will have to manually update their config I suppose.

Yea, no workaround for that unfortunately, but it will only apply to operators using it on Windows and so probably a smaller subset who can be guided through the process.

Cybis320:
> Actually upon reflection i think we should remove all the flags for the Windows compile. ie pc_weave should be blank and -Wall should be removed from the default args in CompilerArgs.py.

Already done in 74b55cdf — win_pc_weave is blank in the template and the default in CompileArgs.py is now [], so MSVC builds with plain distutils defaults (/O2 /W3). Since existing station configs never auto-update and virtually all of them carry -Wall from the old template, getCompileArgs() also strips the literal -Wall on Windows — that's what delivers the fix to the installed base via git pull alone. An explicit /Wall in a config is left untouched in case anyone wants full diagnostics locally.

If a dev really wanted Wall, setting config to /Wall will give you that, but -Wall will be ignored (For Windows only). Is this an acceptable workaround?

markmac99:
> If a dev really wanted Wall, setting config to /Wall will give you that, but -Wall will be ignored (For Windows only). Is this an acceptable workaround?

Yea that should do it !  If its in the branch i can test tomorrow.

markmac99:
Pulled the latest code and can confirm it works as expected. Interestingly `Detection.py` found a few more meteors (84 this time, vs 74 on the last run and 58 i think in "production").  I suspect this is within tolerances.

PR #919, opened 2026-07-03
Fix failed reboot silently starting a second parallel capture in continuous mode

Cybis320:
reboot_after_processing runs os.system('sudo shutdown -r now') without checking the exit code. When the command fails (e.g. no sudo rights under a systemd service user), the failure is invisible and capture resumes — but runCapture had returned with the old BufferedCapture still running, so a second capture starts in parallel. Every frame is then saved twice, and the old capture (holding a stale daytime_mode) labels its copies with the wrong _d/_n suffix. The interleaved duplicates shatter timelapse block grouping into 1–2 frame blocks that fail the 10-image minimum, are never cleaned up, and get rescanned every night (observed on a live station: ~96k warnings and a 40 MB log per night).

Two changes:
- Check the reboot command's exit status; on failure log an error and give up instead of silently retrying 240 times.
- In continuous mode, stop the capture before runCapture returns for the reboot, so a failed reboot can never leave two captures running.

### DIFF
--- a/RMS/StartCapture.py
+++ b/RMS/StartCapture.py
@@ -1264,7 +1264,13 @@ if __name__ == "__main__":
 
                     # Reboot the computer (script needs sudo privileges, works only on Linux)
                     try:
-                        os.system('sudo shutdown -r now')
+                        exit_status = os.system('sudo shutdown -r now')
+
+                        # os.system returns a wait status, not the exit code directly
+                        if exit_status != 0:
+                            log.error('Reboot command failed with exit code {}, '
+                                'giving up on rebooting'.format(exit_status >> 8))
+                            break
 
                     except Exception as e:
                         log.debug('Rebooting failed with message:\n' + repr(e))
@@ -1292,7 +1298,7 @@ if __name__ == "__main__":
 
             # If reboot didn't happen in continuous capture mode, reset so capture resumes normally
             if config.continuous_capture:
-                log.warning("Reboot failed after 4 hours of attempts, resuming capture")
+                log.warning("Reboot did not happen, resuming capture")
                 ran_once = False
 
 
--- a/RMS/StartCapture.py
+++ b/RMS/StartCapture.py
@@ -919,6 +919,11 @@ def runCapture(config, duration=None, video_file=None, nodetect=False, detect_en
                 dropped_frames = bc.stopCapture()
                 log.info('Total number of late or dropped frames: ' + str(dropped_frames))
 
+                # Record the dropped frame count in the observation summary, same as the
+                # normal stop path above
+                obs_dict = getObservationSummaryDict(night_data_dir)
+                addObsParam(obs_dict, "dropped_frames", dropped_frames)
+
             break
 
         ran_once = True
@@ -1276,11 +1281,28 @@ if __name__ == "__main__":
                     try:
                         exit_status = os.system('sudo shutdown -r now')
 
-                        # os.system returns a wait status, not the exit code directly
-                        if exit_status != 0:
+                        if exit_status == 0:
+
+                            # The reboot command was accepted - wait in place for the system to go
+                            # down. Don't re-run the command, as a repeat invocation during the
+                            # shutdown can return non-zero and falsely report a failure
+                            log.info('Reboot command accepted, waiting for the system to go down...')
+                            time.sleep(300)
+                            log.error('System still running 5 minutes after the reboot command '
+                                'succeeded, giving up on rebooting')
+
+                        else:
+
+                            # os.system returns a wait status, not the exit code directly
+                            if hasattr(os, 'waitstatus_to_exitcode'):
+                                exit_code = os.waitstatus_to_exitcode(exit_status)
+                            else:
+                                exit_code = exit_status >> 8
+
                             log.error('Reboot command failed with exit code {}, '
-                                'giving up on rebooting'.format(exit_status >> 8))
-                            break
+                                'giving up on rebooting'.format(exit_code))
+
+                        break
 
                     except Exception as e:
                         log.debug('Rebooting failed with message:\n' + repr(e))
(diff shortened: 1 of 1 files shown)


dvida (commented):
Reviewed both changes against the surrounding code. The diagnosis checks out and both fixes are correct:

- In continuous mode, `bc.stopCapture()` indeed only ran when `STOP_CAPTURE or (not config.continuous_capture)` (L712), so the reboot break at L911 returned with the capture still running, and the failed-reboot path (`ran_once = False`, L1294-1296) started a second parallel capture. The new stop is correctly guarded — standard mode already stops at L718, and the elif is only reachable when `STOP_CAPTURE` is false, so no double-stop.
- `BufferedCapture.stopCapture()` is bounded (~66 s worst case) and returns the dropped-frame count, so it's safe at the new call site.
- The old `for reboot_try in range(4*60)` loop really does spin through all 240 iterations in seconds on a failing `os.system` (no sleep on the `reboot_go` path), matching the described log spam.

A few findings:

**🟡 risk — capture gap of up to 4 hours (new stop before the break, ~L915).** Capture is now stopped *before* the main-loop reboot section, which sleeps 30 s and can then delay the reboot for up to 4 hours in one-minute waits (upload in progress or reboot lock file, L1246-1257). Previously the leaked capture kept recording through those delays; now a continuous station records nothing for the whole delay window. Probably an acceptable trade given the leaked capture was producing corrupt duplicates, but worth acknowledging explicitly — the alternative (stopping capture only once `reboot_go` is actually true) is a bigger restructure.

**🟡 risk — a successful reboot can now falsely report failure and resume capture mid-shutdown (L1274-1283).** When `shutdown -r now` succeeds (exit 0), the for-loop immediately re-invokes it with no sleep until systemd kills the process. If a repeat invocation returns non-zero while the shutdown is already in progress, the new code logs "Reboot command failed", breaks, resets `ran_once`, and starts a fresh capture that the shutdown then kills — leaving a junk partial capture dir that gets flagged as an incomplete capture on the next boot. Suggest sleeping (e.g. `time.sleep(60)`) or breaking out and idling after a zero exit status instead of re-running the command.

**🔵 nit — `exit_status >> 8` logs "exit code 0" if the shell was killed by a signal.** `os.waitstatus_to_exitcode(exit_status)` (Python ≥ 3.9) is exact; alternatively log the raw wait status alongside.

**🔵 nit — the reboot-path stop skips the observation-summary bookkeeping** that the normal stop path does (`getObservationSummaryDict` + `addObsParam(..., "dropped_frames", ...)`, L722-724), so the final night before a reboot logs its dropped-frame count but doesn't record it in the summary. Optional consistency fix.

The first two are worth addressing (or explicitly accepting) before merge; the nits are optional.

Cybis320:
Thanks for the careful review. Pushed a commit addressing three of the four:

- The success path now invokes `shutdown` once and waits in place for the system to go down instead of re-running the command, so a mid-shutdown non-zero return can no longer be misread as a failure. If the process is still alive 5 minutes after a successful command, it gives up and resumes capture (correct if the shutdown was canceled).
- Exit codes are decoded with `os.waitstatus_to_exitcode` where available, with the `>> 8` fallback for older Pythons.
- The reboot-path stop now records `dropped_frames` in the observation summary like the normal stop path.

On the capture gap: accepting the trade-off. During reboot delays `runCapture` has already returned, so nothing updates `daytime_mode` — the old "coverage" was stale-suffixed frames appended to an already-processed directory. A gap seems strictly better, and moving the stop inside the `reboot_go` branch would mean passing the capture handle out of `runCapture`, which feels disproportionate.

dvida (approved):
Re-reviewed `df96996` against the merge result of this branch with the current `prerelease` (`git merge-tree` — clean, and the merged file compiles). Both blocking findings are resolved and I'm happy with the accepted trade-off:

- The success path now runs `shutdown` once and waits in place, so a repeat invocation can no longer be misread as a failure. `os.system` returns 0 only on a zero exit, so the `exit_status == 0` branch is the right discriminator. Residual behaviour (a shutdown that is cancelled or takes over 5 minutes resumes capture) is bounded and correct.
- Exit-code decoding via `os.waitstatus_to_exitcode` with the `>> 8` fallback is signal-aware and keeps pre-3.9 working.
- Capture gap: agreed, the gap is strictly better. During the delay window the leaked capture was appending stale-suffixed frames to an already-archived night directory, so that "coverage" was never uploaded anyway.
- No double stop: the normal stop at the `STOP_CAPTURE or (not config.continuous_capture)` branch cannot coincide with the new one (the `elif` is unreachable when `STOP_CAPTURE` or `disk_full` is set, and the new stop is guarded by `config.continuous_capture`). `bc.stopCapture()` is bounded and safe on an already-dead process, so the watchdog path is fine too.
- The branch is a long way behind `prerelease` but merges cleanly. Worth noting `prerelease` has since changed `upload_manager.upload_in_progress.value` to `.is_set()`; this PR touches no adjacent line, so the merged result keeps `.is_set()`. `#945`'s `nightProcessingState` keys off `archiving == 'completed'`, so nothing here can trigger a reprocess loop.

Two optional nits, neither blocking:

**🔵 nit — the observation-summary write in the reboot path is a no-op, and my earlier suggestion can't be satisfied there.** Unlike the normal stop path, which runs *before* the post-processing block, this stop runs *after* `processNight()`. `finalizeObservationSummary()` has by then written the final JSON and deleted the working one, so `getObservationSummaryDict(night_data_dir)` creates a fresh working JSON containing only `night_data_dir` and `dropped_frames`, plus a `.lock` file, in an already-archived directory. The value never reaches the final summary or the uploaded archive. Harmless (post-#945 reprocessing is decided by the processing-status record), but it is dead weight plus a stray file — suggest either dropping the two lines and keeping the `log.info`, or recording `dropped_frames` before the processing block.

**🔵 nit — the `except Exception` path still skips the new `break`**, so an exception there resumes the old no-sleep 240-iteration spin. Pre-existing shape and near-unreachable now, so only worth a `break` in the handler if you're touching the file anyway.

Good to merge as-is.

PR #926, opened 2026-07-08
Fix OOM caused by orphaned RawFrameSaver processes after watchdog force-kills

Cybis320:
### Problem

When the watchdog force-kills a wedged `BufferedCapture` process (SIGKILL), its `RawFrameSaver` child is orphaned rather than terminated. The orphan spins forever on a shared exit `Event` that no one will ever set, holding its inherited ~450 MB frame buffer. Every watchdog restart adds another orphan, so over days a station accumulates hundreds of zombie savers and the box OOMs.

The root trigger is a GStreamer teardown hang: when a camera is connected but unresponsive, `pipeline.set_state(NULL)` can block indefinitely (RTSP TEARDOWN never completes). That freezes the capture loop's heartbeat, the watchdog fires, and the force-kill → orphan → leak cycle begins.

### Fix

Three layers, from root cause outward:

1. **Bound GStreamer teardown** (`RMS/BufferedCapture.py`) — Both the graceful and the forced `set_state(NULL)` calls now run through `runWithTimeout()` with a 10 s cap (`GST_TEARDOWN_TIMEOUT`), well under the 180 s watchdog limit. A hung teardown is abandoned (left to unwind in a daemon thread) so capture keeps its heartbeat and is never force-killed in the first place.

2. **Kernel-level orphan reaping** (`RMS/Misc.py`) — New `setParentDeathSignal()` uses Linux `PR_SET_PDEATHSIG` so the child receives SIGKILL the moment its parent dies, including the race where the parent died before the signal was armed (checked via `getppid() == 1`). No-op on non-Linux.

3. **Start-method-agnostic orphan guard** (`RMS/RawFrameSave.py`) — `PR_SET_PDEATHSIG` tracks the OS parent, which under the `forkserver` start method (Python 3.14's new Linux default) is the fork-server, not `BufferedCapture`. So the saver also records its logical parent's PID in `__init__` (which runs in the parent, correct under fork/spawn/forkserver alike) and polls it with `os.kill(pid, 0)` in its wait loop, self-terminating if the parent is gone. Covers Python 3.6–3.14 on all platforms.

### Changes

- `RMS/BufferedCapture.py` — time-bound both teardown paths; abandon a wedged `set_state(NULL)` instead of blocking the heartbeat
- `RMS/Misc.py` — add `setParentDeathSignal()` (prctl `PR_SET_PDEATHSIG` + already-orphaned fast path)
- `RMS/RawFrameSave.py` — arm the death signal in `run()`; add parent-PID liveness check as the forkserver-safe fallback

### Impact

- A wedged camera no longer trips the watchdog: capture abandons the stuck teardown and keeps running.
- If a force-kill does happen for any other reason, the `RawFrameSaver` now exits immediately (Linux) or within one poll cycle (fallback path) instead of leaking its buffer forever.
- No behavior change on the normal shutdown path.


Cybis320:
@dvida, this PR should be prioritized as the MQTT monitor see OOM crashes caused by this issue quite regularly. This PR has been running on a pathological station for over a week without restarting RMS instances and memory is now stable.

RMS/RawFrameSave.py
```diff
@@ -233,6 +246,18 @@ def run(self):
 
                         return None
 
+                    # Forkserver-safe orphan guard: if the logical parent (BufferedCapture)
+                    # is gone, our exit Event will never be set, so self-terminate instead
+                    # of spinning here forever holding the frame buffer. Complements
+                    # PR_SET_PDEATHSIG (which doesn't fire correctly under forkserver).
+                    try:
+                        os.kill(self.parent_pid, 0)
```
dvida: **[HIGH] `os.kill(pid, 0)` is not a liveness probe on Windows — it breaks frame saving there.**

On Windows, signal `0` == `signal.CTRL_C_EVENT`, so CPython routes `os.kill(pid, 0)` to `GenerateConsoleCtrlEvent(0, pid)` instead of a liveness check. Since `parent_pid` is not a process-group ID, that call fails → `OSError` → this guard concludes "parent gone" and the saver exits ~0.1 s after start, silently killing raw-frame saving while capture keeps running. (If it *did* succeed it would deliver Ctrl+C to the console group — also wrong.)

The PR description claims Python 3.6–3.14 "on all platforms"; impact is latent today (capture is effectively Linux-only per the README, and the watchdog itself uses `signal.SIGKILL`), but this is a deterministic trap for any future Windows capture support.

One-line fix — gate the probe:

```python
if os.name == 'posix':
    try:
        os.kill(self.parent_pid, 0)
    except OSError:
        ...
```

(or use `multiprocessing.parent_process().is_alive()` on py3.8+, which is correct on all platforms and start methods).

RMS/RawFrameSave.py
```diff
@@ -233,6 +246,18 @@ def run(self):
...
+                    except OSError:
```
dvida: **[MEDIUM] `except OSError` conflates EPERM with ESRCH.**

POSIX `kill(pid, 0)` returns `EPERM` when the process *exists* but isn't signalable — that also lands here and reads as "parent dead", terminating a healthy saver. Not reachable in RMS today (parent and child run as the same user), but the canonical guard is `except ProcessLookupError:` (the capture stack is py3-only — `BufferedCapture.py` already uses f-strings), or checking `e.errno == errno.ESRCH`.

RMS/RawFrameSave.py
```diff
@@ -233,6 +246,18 @@ def run(self):
...
+                        log.warning('RawFrameSaver: parent process %d gone, exiting orphan',
+                                    self.parent_pid)
+                        self.run_exited.set()
+                        return None
```
dvida: **[LOW, optional] Orphan-exit drops tail frames.**

`stop()` flushes leftover buffered frames before shutdown; this exit path returns without flushing. Since the parent is dead the buffers are final — flushing the last block here (same leftover-scan `stop()` does) would save up to one block of frames that would otherwise be lost. Optional: the PDEATHSIG SIGKILL path can't flush either, so this only improves the fallback path.

RMS/Misc.py
```diff
@@ -45,6 +45,39 @@
 log = getLogger("rmslogger")
 
 
+def setParentDeathSignal(sig=9):
```
dvida, line 48: **[MEDIUM] Docstring omits two hazards that matter for a generic `Misc.py` utility.**

Future callers will trust this docstring, so two prctl semantics are worth documenting:

1. **`PR_SET_PDEATHSIG` is armed against the parent *thread*, not the process.** Per prctl(2), the signal fires when the thread that created the child terminates — not when the whole parent process exits. Safe for the current caller (`RawFrameSaver` is spawned from `BufferedCapture.run()`, the process's main thread), but a spurious-SIGKILL trap for anyone who later spawns a worker from a short-lived helper thread.

2. **`getppid() == 1` misses subreapers.** Under systemd user sessions or containers, orphans are reparented to a subreaper, not init, so the already-orphaned race check can silently miss. `RawFrameSaver`'s PID poll covers that case; the docstring should note this check is best-effort.

RMS/Misc.py
```diff
@@ -45,6 +45,39 @@
...
+
+    Linux-only (prctl PR_SET_PDEATHSIG); no-op elsewhere. Call as early as possible in a
+    child's run(). Also handles the race where the parent already died before the call.
+
+    Arguments:
+        sig: [int] Signal to receive on parent death. Default SIGKILL.
+    """
+    try:
+        import ctypes
+        libc = ctypes.CDLL("libc.so.6", use_errno=True)
```
dvida: **[LOW] Hardcoded `libc.so.6` + fully silent failure.**

On musl-based systems (Alpine) `CDLL("libc.so.6")` raises, and the bare `except Exception: pass` means the whole protection is lost with zero trace. Suggest:

```python
import ctypes.util
libc_name = ctypes.util.find_library("c") or "libc.so.6"
libc = ctypes.CDLL(libc_name, use_errno=True)
if libc.prctl(1, int(sig), 0, 0, 0) != 0:
    log.debug("setParentDeathSignal: prctl failed (errno %d)", ctypes.get_errno())
```

and a `log.debug` in the `except` handler so a station losing this protection is at least visible in debug logs. Also `sig=9` → `signal.SIGKILL` for readability.

RMS/BufferedCapture.py
```diff
@@ -1593,27 +1600,49 @@ def _timedCall(fn, timeout_s=2):
...
+                ok, ret, exc = runWithTimeout(self.pipeline.set_state,
+                                              args=(Gst.State.NULL,),
+                                              timeout=GST_TEARDOWN_TIMEOUT)
+                if not ok:
+                    # Teardown hung. Abandon it (the call keeps running in a daemon thread,
+                    # holding the old pipeline until it eventually unwinds) and drop our
+                    # reference below. Do NOT retry/verify - keeping capture alive matters
+                    # more than a clean teardown of an already-dead camera.
+                    log.warning("releaseResources: set_state(NULL) did not return within %ds - "
+                                "abandoning teardown to keep capture alive (pipeline left to GC)",
```
dvida, line 1621: **[LOW, informational] Abandoned teardowns accumulate a thread + pipeline + socket fd each.**

Each abandoned `set_state(NULL)` parks a daemon thread that holds the pipeline (and its RTSP socket) until the call unwinds — possibly never. Repeated wedge/reconnect cycles therefore leak one thread+pipeline+fd per wedge. That's vastly better than leaking 450 MB processes and it's bounded by reconnect frequency, but `runWithTimeout` already has an unused `on_late_completion` hook — passing e.g. `on_late_completion=lambda: log.info("abandoned set_state(NULL) finally unwound")` would make late unwinds (vs. permanent hangs) visible in station logs for free.

dvida (changes requested):
Thorough review of all three layers, checked against the `prerelease` sources (spawn site, watchdog behavior, heartbeat budget, teardown paths).

## What checks out

- The layered architecture is right: bound the root-cause hang, add a kernel guarantee (`PR_SET_PDEATHSIG`), keep a portable fallback (PID poll). Each layer has a stated reason.
- `RawFrameSaver` is spawned from `BufferedCapture.run()` (`RMS/BufferedCapture.py:2102`), i.e. the capture process's **main thread**, so the prctl fires-on-parent-*thread*-death gotcha (see inline comment) does not bite today.
- `self.parent_pid = os.getpid()` in `__init__` is correct — `__init__` runs in the parent and the value survives fork/spawn/forkserver pickling.
- The watchdog (`RMS/StartCapture.py:499-510`) joins after SIGKILL, so the parent zombie is reaped and the `os.kill(pid, 0)` probe can't be fooled by a lingering zombie.
- The timeout logic is careful: a timed-out graceful `set_state` does **not** fall through to the force path, so two `set_state(NULL)` calls never run concurrently on the same pipeline.
- Heartbeat budget: worst case ~30 s of teardown blocking (10 s graceful + 10 s force + 6 s bus join + short waits) ≪ 180 s watchdog limit.
- Normal shutdown path is unchanged; flush-on-stop behavior intact.

## Blocking findings

The two blocking items are the Windows behavior of `os.kill(pid, 0)` (inline comment on `RawFrameSave.py`) and a remaining unbounded teardown path:

**Coverage gap — unbounded `set_state(NULL)` remains in the pipeline-creation retry path.** `createGstreamDevice`'s per-attempt `except` block (around `RMS/BufferedCapture.py:1250`) still calls `self.pipeline.set_state(Gst.State.NULL)` raw:

```python
except Exception as e:
    log.error("Attempt {} failed: {}".format(attempt + 1, str(e)))
    # Clean up any partial pipeline that was created
    if hasattr(self, 'pipeline') and self.pipeline:
        try:
            self.pipeline.set_state(Gst.State.NULL)   # <-- unbounded
```

In the reconnect loop the savers are already running, so a wedge here re-creates the exact hang → watchdog kill → orphan scenario this PR fixes (layers 2/3 now contain the damage, but the heartbeat still dies). For symmetry this call should also go through `runWithTimeout(..., timeout=GST_TEARDOWN_TIMEOUT)`.

## Non-blocking

**Duplicate timeout helpers now coexist in the same function.** `releaseResources` keeps its local `_timedCall()` (line ~1557, used for `cap.release()`) right next to the new `runWithTimeout` calls — same job, two idioms. Suggest replacing `_timedCall(self.device.release)` with `runWithTimeout(self.device.release, timeout=2)` and deleting the local helper.

Remaining smaller points are inline. Overall: sound fix for a real, well-diagnosed leak, and the PR description is exemplary — the blocking items are both small, local changes.

Cybis320:
Thanks for the thorough review — all points addressed in 6c00949 except the optional orphan-flush.

**Blocking:**

- **Retry-path `set_state(NULL)`** — now goes through `runWithTimeout(..., timeout=GST_TEARDOWN_TIMEOUT)` like the other teardown paths, with a warning when abandoned. Agreed this was a coverage gap: in the reconnect case the savers are already running, so a wedge there would have recreated the exact scenario.
- **`os.kill(pid, 0)` on Windows** — probe is now gated to `os.name == 'posix'`. Went with the gate rather than `multiprocessing.parent_process()` since that's py3.8+ and we still claim 3.6.

**Non-blocking:**

- **EPERM/ESRCH** — now catches `ProcessLookupError` only.
- **Docstring** — added both hazards: `PR_SET_PDEATHSIG` arms against the parent *thread* (safe here, `RawFrameSaver` is spawned from `BufferedCapture.run()`'s main thread), and `getppid() == 1` misses subreapers, so it's documented as best-effort with the PID poll as the hard guarantee.
- **libc** — resolved via `ctypes.util.find_library("c")` with `libc.so.6` fallback, prctl return code checked, and failures now logged at debug instead of silently swallowed. One deviation from the suggestion: kept the default `sig=9` numeric instead of `signal.SIGKILL`, because `signal.SIGKILL` doesn't exist on Windows and a default arg evaluates at import time — noted in the docstring.
- **`_timedCall` duplication** — deleted; `cap.release()` now uses `runWithTimeout(timeout=2)`.
- **Late-unwind visibility** — all three abandoned-teardown sites now pass `on_late_completion` logging at info, so station logs distinguish late unwinds from permanent hangs.

**Skipped:** the orphan-exit tail-frame flush. The PDEATHSIG path can't flush anyway, and adding flush logic to an emergency-exit path adds failure modes to code that only runs when things are already broken — happy to revisit if you feel strongly.

PR #927, opened 2026-07-09
Filter catalog stars with spherical FOV polygon in ClearSkyDetector

Cybis320:
## Problem

`detectFrame` in `RMS/ClearSkyDetector.py` projected the entire star catalog with `getCatalogStarsImagePositions` and kept anything passing a rectangular `0 <= x < w, 0 <= y < h` bounds check. Stars far outside the FOV (including behind the camera) get folded back into pixel bounds by the distortion model, so they were counted as expected stars. This inflates `expected_count` and pollutes `star_positions` with phantom stars that can never match a detection, biasing the detector toward cloudy verdicts — worst for wide-FOV lenses.

## Fix

Filter the catalog the same way `predictStarNumberInFOV` (Utils/Flux.py) does before projecting: build the image-outline polygon (5 points per side), convert it to RA/Dec at the frame's JD with `xyToRaDecPP`, and keep only catalog stars inside the spherical polygon (`pointInsideConvexPolygonSphere`) before calling `raDecToXYPP`. The existing bounds and mask checks are unchanged.

## Validation

Synthetic whole-sky catalog (2° grid, 16,110 stars) against a default 1280x720 platepar: the old bounds-only check kept 16,020 "stars" (nearly the whole sphere folds back into bounds), of which 12,603 were outside the FOV cone entirely. The fixed code keeps 89, matching the actual ~350 sq deg FOV.

### DIFF
--- a/RMS/ClearSkyDetector.py
+++ b/RMS/ClearSkyDetector.py
@@ -17,10 +17,12 @@ import numpy as np
 from scipy.ndimage import maximum_filter, uniform_filter
 from scipy.spatial import Voronoi
 
-from RMS.Astrometry.Conversions import date2JD
+from RMS.Astrometry.ApplyAstrometry import raDecToXYPP, xyToRaDecPP
+from RMS.Astrometry.Conversions import date2JD, jd2Date
 from RMS.ConfigReader import findConfigInDir
 from RMS.Formats import StarCatalog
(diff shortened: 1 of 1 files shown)


Cybis320:
Superseded — the file was included in prerelease by accident and is being removed entirely (see the removal PR). The fixed version of the file is preserved on the `fix-clearsky-fov-filter` branch for out-of-tree development.

PR #932, opened 2026-07-12
Add star coverage gate to per-FF recalibration

Cybis320:
## Summary

Reject a per-FF recalibration fit that matched fewer than a minimum fraction of the detected stars (default 0.6, configurable via `recalibration_min_match_fraction`).

`checkFitGoodness` only inspects the residuals of the stars that did match, so a sparse spurious subset at a wrong pointing could pass, get stamped `auto_recalibrated`, and chain forward as the seed for the next FF — walking the pointing away over a cloudy night. The gate applies even with `force_platepar_save`, so the last-resort coarse-radius fallback can no longer save such fits either.

The gate and the photometry fit now use the matched stars from the last successful fit iteration instead of whatever the last `matchStarsResiduals` call produced — a failed tighter-radius iteration used to overwrite `matched_stars` with matches of a platepar that was never accepted (and could leave it unbound under `force_platepar_save`).

SkyFit2's manual pointing-only recalibration disables the gate since the result is user-inspected.

## Changes

- `RMS/Astrometry/ApplyRecalibrate.py` — coverage gate + track matched stars from the last accepted iteration
- `RMS/ConfigReader.py`, `.config` — new `recalibration_min_match_fraction` option (default 0.6)
- `Utils/SkyFit2.py` — disable the gate for manual pointing-only recalibration

### DIFF
--- a/.config
+++ b/.config
@@ -797,6 +797,12 @@ min_matched_stars: 20
 ; Maximum number of stars to use for the fit on individual images
 recalibration_max_stars: 200
 
+; Minimum fraction of detected stars that the per-image recalibration fit
+; must match to be accepted. Rejects sparse spurious fits (e.g. on cloudy
+; nights) which would otherwise make the pointing drift away when chained
+; from image to image
+recalibration_min_match_fraction: 0.6
+
 ; If the average distance (pixels) between catalog and image stars is below
 ; this threshold, astrometry recalibration will not run but the existing
 ; calibration will be used
--- a/RMS/Astrometry/ApplyRecalibrate.py
+++ b/RMS/Astrometry/ApplyRecalibrate.py
@@ -108,6 +108,7 @@ def recalibrateFF(
     lim_mag=None,
     ignore_distance_threshold=False,
     ignore_max_stars=False,
+    min_match_fraction=None,
 ):
     """Given the platepar and a list of stars on one image, try to recalibrate the platepar to achieve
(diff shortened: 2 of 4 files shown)


dvida:
## Testing report

Drove `recalibrateFF` over 7 datasets spanning **FOV 20°→146°** (narrow/wide/ultrawide, dense CALSTARS), ~120 stride-sampled FFs each, run at gate 0.0 and 0.6 — **840 calls**.

- **0 crashes, 0 gate-decision inconsistencies** across all FOVs. Gate accepts iff fit succeeded AND matched fraction ≥ 0.6.
- On clear-sky data the gate rejects ~nothing (median matched fraction 0.91–1.00) — no over-rejection of legitimate fits.
- Gate fired once naturally: an ultrawide (146°) frame at fraction 0.538 → rejected at 0.6, accepted at 0.0. Intended low-coverage rejection.
- `force_platepar_save=True` + impossible gate → `None`, confirming the gate applies even under force-save.

**Latent-crash fix confirmed before/after** (wrong-pointing + `force_platepar_save`, same input):
- base `prerelease`: `UnboundLocalError: 'matched_stars'` at `ApplyRecalibrate.py:364`
- this PR: returns `None` gracefully

So the gate does what it claims without regressing good fits, and the PR also fixes a real, reachable crash in the coarse force-save fallback. LGTM.

PR #933, opened 2026-07-13
Make UDP buffer sizes persist across reboots

Cybis320:
## Problem
`UpdateBuffers.sh` writes UDP buffer sizes to `/etc/sysctl.d/99-rms-udp-buffers.conf`, but the values silently revert to the old default after a reboot.

Cause: `systemd-sysctl` applies config files in lexical filename order (last wins). `/etc/sysctl.conf` is loaded via the symlink `/etc/sysctl.d/99-sysctl.conf`, which sorts *after* `99-rms-udp-buffers.conf` (`s` > `r`). So any `rmem_max`/`wmem_max` line left in `/etc/sysctl.conf` (e.g. from an older version of this script that edited it directly) overrides the drop-in at every boot. The change only appears to work because the script also applies it live via `sysctl -p`.

## Fix
When `/etc/sysctl.conf` contains active `net.core.rmem_max`/`wmem_max` lines, the script now backs it up (`.rms.bak`) and comments them out, so the drop-in is the sole authority and the setting survives reboots. Idempotent and a no-op when there's nothing to heal.

## Compatibility
Raspberry Pi OS and Ubuntu — both are systemd-based Debian-family systems with the same sysctl load ordering and GNU userland (`grep -E`, `sed -E`, `cp -n`).

### DIFF
--- a/Scripts/UpdateBuffers.sh
+++ b/Scripts/UpdateBuffers.sh
@@ -107,6 +107,16 @@ echo "Created $SYSCTL_DROP_IN with:"
 echo "  net.core.rmem_max=$RECOMMENDED_SIZE"
 echo "  net.core.wmem_max=$RECOMMENDED_SIZE"
 
+# Comment out any rmem_max/wmem_max lines left in /etc/sysctl.conf.
+# It loads after our drop-in at boot and would otherwise override it.
+SYSCTL_CONF="/etc/sysctl.conf"
+if [ -f "$SYSCTL_CONF" ] && grep -Eq '^[[:space:]]*net\.core\.(rmem|wmem)_max[[:space:]]*=' "$SYSCTL_CONF"; then
+    echo -e "\nFound conflicting UDP buffer settings in $SYSCTL_CONF; disabling them so the drop-in wins at boot."
+    cp -n "$SYSCTL_CONF" "${SYSCTL_CONF}.rms.bak" && echo "Backed up to ${SYSCTL_CONF}.rms.bak"
+    sed -i -E 's@^([[:space:]]*net\.core\.(rmem|wmem)_max[[:space:]]*=.*)@# Disabled by UpdateBuffers.sh: \1@' "$SYSCTL_CONF"
+    echo "Commented out stale rmem_max/wmem_max lines in $SYSCTL_CONF"
+fi
+
 # Apply changes immediately
 echo "Applying changes..."
 sysctl -p "$SYSCTL_DROP_IN" >/dev/null 2>&1


dvida (approved):
Verified the root-cause analysis and the fix — both check out. Approving with two suggestions.

**Verification notes:**
- The stale lines are self-inflicted by RMS history: the original script (43c6c5d9) edited `/etc/sysctl.conf` directly with the old 1 MB value; d0c328d3 switched to the drop-in but never cleaned up, and 0823b0d6 later raised the recommendation to 16 MB. So any station that ran the pre-Dec-2025 script has `net.core.rmem_max=1048576` in `sysctl.conf`, and after a reboot the kernel clamps rtspsrc's 16 MB `udp-buffer-size` back down → RcvbufErrors/dropped frames on UDP stations. Real operational trap, well caught.
- Ordering claim confirmed against `man sysctl.d` (lexicographic, latest wins; `99-rms-udp-buffers.conf` < `99-sysctl.conf`), and the `/etc/sysctl.d/99-sysctl.conf → ../sysctl.conf` symlink is standard on Debian/Ubuntu/RPi OS.
- The heal is idempotent (commented lines no longer match the grep guard) and a safe no-op when `sysctl.conf` is absent.

**Suggestion 1 — heal block is unreachable when live values are already OK:**
`show_settings` exits early ("at or above recommended", line 73–74) before the new block runs. A station that already ran the drop-in version this boot (live 16 MB via `sysctl -p`, stale 1 MB lines still in `sysctl.conf`) gets "all fine, exiting" on a re-run — the latent persistence bug survives until the next reboot degrades the values, and only a second run after that reboot heals it. Consider moving the heal (or at least the detection) before the early exit so a single run fixes persistence regardless of current live values.

**Suggestion 2 — `cp -n` (nit):**
On coreutils ≥ 9.4 (Ubuntu 24.04+, Debian 13) `cp -n` prints `cp: warning: behavior of -n is non-portable and may change in future` to stderr on every invocation, and coreutils 9.2 briefly made the skip exit non-zero. Nothing breaks here (no `set -e`; the `&&` only gates the echo), but it's noisy on the very OS d0c328d3 targeted. A portable quiet alternative:
```sh
[ -e "${SYSCTL_CONF}.rms.bak" ] || cp "$SYSCTL_CONF" "${SYSCTL_CONF}.rms.bak"
```
(`--update=none` needs coreutils ≥ 9.3, so the test guard is the safer choice.)

Out-of-scope but worth knowing: conflicts in other late-sorting drop-ins (e.g. a user's `99-zz-*.conf`) or slash-form keys aren't handled — fine, since RMS's own legacy only ever wrote dot-form lines to `sysctl.conf`.

PR #934, opened 2026-07-14
Fix three crashes that log tracebacks on healthy stations

Cybis320:
Three exceptions that RMS catches and logs as full tracebacks on stations that are capturing perfectly well. They are noise that masks real failures — and one of them fires on **every camera, twice a day** on our test stations (see scoping note under bug 1).

All three were found by grepping capture logs across a 6-camera fleet, and each is reproduced below from a real log line.

---

### 1. `RawFrameSave.stop()` — `TypeError: 'NoneType' object is not iterable`

```
ERROR-BufferedCapture - Error in capture process: 'NoneType' object is not iterable
  File "RMS/RawFrameSave.py", line 197, in stop
    for frame, ts in zip(self.array1, self.timeStamps1):
TypeError: 'NoneType' object is not iterable
```

`stop()` zips `self.array1` / `self.timeStamps1`, but the shared arrays can already be gone by then:

* `stop()` frees them itself (the `del self.array1 …` at the end of the same function), so it is not re-entrant; and
* `BufferedCapture.releaseRawArrays()` calls `stop()` on **both** the shutdown path **and** the day/night mode switch (`"Cleaning up existing raw frame saver before mode change"`), where the saver is torn down before the arrays are re-created for the new frame shape.

**Impact scoping (edited after review):** the quoted traceback and the 2×/day frequency come from stations running prerelease with #938's lazy view building plus in-development continuous-capture changes that tear the saver down at every day/night switch. On plain prerelease this exact `TypeError` cannot fire; the same root cause instead surfaces as a latent re-entry `AttributeError` (concurrent teardown race). The guard fixes both shapes, and #938 depends on it — see the review discussion below.

Fix: skip the flush when the buffers are already released (there is nothing to flush), and only zero the timestamp arrays that still exist.

---

### 2. `getDaysSinceLastDetection()` — `ValueError: time data '0'`

```
ERROR-ObservationSummary - Failed to calculate time since last detection:
ValueError("time data '0' does not match format '%Y-%m-%d %H:%M:%S'")
  File "RMS/Formats/ObservationSummary.py", in getDaysSinceLastDetection
    time_last_fits_file_for_session = datetime.datetime.strptime(result, "%Y-%m-%d %H:%M:%S")
```

The `if result is None` check guards a missing **row**, but `time_last_fits_file` holds **`0`, not NULL**, until the session's first FITS file is recorded. So the row exists, `str(result[0])` is `'0'`, and `strptime` raises.

**Impact: fires nightly.**

---

### 3. `getDaysSinceLastDetection()` — `TypeError: 'NoneType' object is not subscriptable`

A latent crash in the *second* query block of the same function, found while fixing #2:

```python
cursor = conn.execute(last_detection_time_for_session_sql)
result = cursor.fetchone()[0]   # fetchone() is None when no row matches
```

A station with **no detections yet** matches no row, so `fetchone()` returns `None` and indexing it raises. This hits new stations hardest.

---

### The fix

* Add `parseObsTimestamp()` — parses a column that may be `None`, `0`, absent, or a timestamp with/without microseconds, returning `None` when there is no usable value. This replaces the ad-hoc `if '.' in result` format branch.
* Return `"Unknown"` for both #2 and #3: having nothing to measure *from* (no FITS file yet) or *to* (no detection yet) is not an error.
* Guard the flush in `RawFrameSave.stop()` against already-released buffers.
* **Connection leak:** `conn.close()` was only reached on the success path in both query blocks, so the early `return "Unknown"` and the `except` handler both leaked the DB connection. Now closed in `finally`.

### Behaviour change

A non-parseable last-detection time now returns `"Unknown"` rather than `"Error"`, which better reflects "no detection recorded yet". Flagging it explicitly in case any downstream consumer distinguishes the two.

### Testing

Both original failures reproduce exactly against the pre-fix code (matching the log lines above verbatim) and are cleared by the patch. `parseObsTimestamp` round-trips `0`, `"0"`, `None`, `""`, `garbage` → `None`, and both timestamp formats (with and without microseconds) → correct `datetime`. Both files compile clean.

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -513,6 +513,39 @@ def timeSyncStatus(config, d, force_client=None):
 
     return ahead_ms
 
+
+def parseObsTimestamp(value):
+    """Parse an observation-database timestamp, or return None if there isn't one.
+
+    The observation-DB time columns are not NULL before they are first written --
+    they hold 0 -- so checking the row for None is not enough. str(0) is '0', and
+    strptime('0', "%Y-%m-%d %H:%M:%S") raises ValueError.
+
+    Arguments:
+        value: the raw column value (may be None, 0, or a timestamp string).
+
+    Return:
+        [datetime] the parsed time, or None if the column holds no usable timestamp.
+    """
+
+    if value is None:
+        return None
+
+    text = str(value).strip()
+
+    if (not text) or (text == "0"):
+        return None
+
+    # With and without microseconds
+    for time_format in ("%Y-%m-%d %H:%M:%S.%f", "%Y-%m-%d %H:%M:%S"):
+        try:
+            return datetime.datetime.strptime(text, time_format)
+        except ValueError:
+            continue
+
+    return None
+
+
 def getDaysSinceLastDetection(config, data_dir, d=None, debug=False):
     """Get the number of days since the last meteor detection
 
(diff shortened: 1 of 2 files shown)


dvida:
Reviewed with empirical repros against pre/post-fix code: bugs 2 and 3 and the connection leak all reproduce exactly as described and are fixed by this patch, with the healthy path unchanged. LGTM overall.

One question on bug 1: upstream, `stop()` has a single call site (`releaseRawArrays()`, which nulls the ref in `finally`), and a re-entrant call raises `AttributeError` (attributes are `del`'d), not the observed `TypeError`. Also the quoted traceback's line 197 doesn't match master or prerelease (the `zip` is at ~183). Were those logs from a modified fork? The guard itself is good defensive code either way — just worth softening the "2×/day on every camera" claim if it's fork-specific.

Minor: a `log.debug` in the skip path would keep genuine tail-frame loss diagnosable.

Cybis320:
Checked applicability against prerelease directly — **yes, the guard applies as-is, though on prerelease it's defensive hardening rather than a live crash fix**. Details:

On prerelease, `stop()`'s flush zips `self.array1` directly, and the attributes exist from construction until the `del` at the end of `stop()` itself — so a first `stop()` can never raise the `TypeError` I quoted. You were right to question the provenance: those logs came from stations where we run prerelease with several of the open PRs merged in for testing — including #938, whose spawn-safety rework builds the saver's numpy views lazily from the `mp.Array` bases (so `self.array1` can legitimately be `None`), plus in-development continuous-capture changes that tear the saver down and rebuild it at every day/night switch (hence "2×/day", and the shifted line number). I've re-scoped the claim in the PR description accordingly.

What prerelease does have is the re-entry shape of the same root cause: a second `stop()` on one instance hits the zip with `del`'d attributes → `AttributeError`, and `releaseRawArrays()`'s `try/finally` has no `except`, so it propagates. Sequential double-calls are already prevented by the null-in-`finally`; the residual exposure is concurrent teardown — a race, not a nightly certainty.

One dependency worth making explicit: **#938 needs this guard.** Its reworked `stop()` calls `ensureViews()` and then zips `self.array1` without these checks, so the re-entry crash returns there (`ensureViews` reads a `del`'d attribute). Running the two together on our stations surfaced exactly that conflict, and the resolution that works is the composed form: the `getattr`/None guard first, with `ensureViews()` in a try — ensureViews rebuilds views while the shared bases still exist, the guard covers genuinely-released teardown. Suggest merging this PR first and letting #938 rebase onto that. The upcoming day/night capture PR (where saver teardown happens twice daily by design — the source of my logs) will depend on this guard even harder.

Added the `log.debug` in the skip path per your suggestion.

g7gpr:
Great to see that traceback feature getting used!

PR #935, opened 2026-07-16
Fix multiprocessing deadlock cascade (OOM)

Cybis320:
## Incident

An 8-day soak on a 5-station machine (NZ005A-E, 16 GB RAM) ended in machine-wide OOM. Post-mortem with py-spy on the live processes reconstructed the full chain:

1. **Dawn, day 7:** on two stations (D, E, within 65 s of each other), the logging pipeline wedged: a process died holding one of the logging queue's shared internal locks. From that instant both stations went silently log-dark — no log lines, no new log files (the listener process creates them) — while continuing to capture. Confirmed by py-spy: the listener was alive and idle in `queue.get()`, starved.
2. Every producer process then buffered log records in its `QueueFeederThread` without bound (~700 MB/process/day — the "memory leak" we had been chasing). The stations ran blind through a full day/night cycle.
3. **Dawn, day 8:** both mains wedged permanently inside `Compressor.stop()` at `self.exit.set()` — `multiprocessing.Event.set()` blocked forever on a semaphore orphaned by a killed process. Identical py-spy stacks on both stations.
4. An external monitor saw stations with no fresh log files and respawned `StartCapture` every ~10 s with no already-running guard: **~40 duplicate instances at ~700 MB each**. The kernel OOM killer logged 17 kills in 4 bursts; each burst stalled the box hard enough to knock healthy stations' cameras offline (every "camera disconnect" on station A that day matches an OOM burst to the second).

## Fixes

Each commit is independent:

- **Logger: bound the queue, drop-don't-block, and auto-recover.** The queue is now bounded (30k records) so a stalled pipeline caps producer memory instead of growing it forever. New `checkLoggingHealth()` (called from the capture watchdog every 60 s) detects a dead listener *or* a backlog that stops draining, and restarts the pipeline with a fresh queue + listener. A wedged logging pipeline is now a ~3-minute blip instead of a silent permanent outage.
- **QueuedPool: stop terminating workers that exited on their own.** `pool.terminate()` ran unconditionally on every shutdown path — even after workers swallowed their poison pills. Killing a worker mid-write on the shared logging queue is exactly the lock-orphaning hazard above. All shutdown paths now close and wait (bounded) for graceful exit; `terminate()` remains only as a last resort and for genuinely stuck workers.
- **Compression: lock-free exit flags.** `exit`/`run_exited` are only ever set and polled, never waited on, so they don't need `multiprocessing.Event`'s semaphore — a plain shared byte can't deadlock no matter what dies. This directly fixes the observed `stop()` wedge.
- **StartCapture: per-station `flock()` single-instance guard + watchdog hook.** A duplicate instance now refuses to start, before any heavy initialization (a refused clone exits in milliseconds instead of idling at 700 MB). `flock` is released by the kernel however the process dies, so a crashed instance never blocks a legitimate restart. Verified including the SIGKILL case.

## Testing

- End-to-end test of the logging watchdog: dead-listener detection replaces the listener and subsequent records land in a new log file; starvation strikes (3 consecutive checks with a non-draining backlog) trigger a restart; the drop-on-full handler neither blocks nor raises. All passed.
- Flock guard: second instance exits with code 1 and a clear message; after SIGKILL of the holder, a new instance acquires the lock immediately. Passed.
- `qsize()` is unavailable on macOS (`NotImplementedError` is handled — health check degrades to listener-aliveness only); stations run Linux where it works.

## Notes / follow-ups

- Child processes forked before a logging restart keep logging into the abandoned (bounded) queue; their records are lost until those processes are recycled. Fixing that fully needs a per-connection transport (e.g. `SocketHandler`) — proposed as follow-up.
- `oom_score_adj` protection was considered and deferred: making workers preferred OOM victims *increases* the odds of orphaning shared queue locks; with the recovery watchdog in place this trade-off should be revisited separately.


Cybis320:
Added a fifth commit sweeping the rest of the codebase for the same deadlock class (a killed process orphaning a shared multiprocessing lock):

- `AtomicFlag` moved to `RMS.Misc` with a poll-based `wait()` for drop-in `Event` compatibility.
- All set/poll exit events converted: **BufferedCapture** (SIGKILLed by the watchdog, `is_set()` polled in the capture hot loop — the highest-traffic instance), **RawFrameSaver** (terminated in the normal disconnect path), UploadManager, EventMonitor, LiveViewer, GstreamerCapture, VideoExtraction, and `QueuedPool.kill_workers`.
- Single-writer shared `Value`s (`multiprocessing.Value` acquires a shared lock on **every** `.value` access by default) switched to `lock=False`: capture heartbeat, `daytime_mode`, `camera_mode_switch_trigger`, frame-block start timestamps.
- `SafeValue` counters genuinely need locked read-modify-write, so they get bounded acquisition instead: after 5 s the operation warns and proceeds without the lock — stale counter beats permanent wedge. Test demonstrates the before/after: a process dying while holding the lock previously hung `value()` forever, now it returns after the timeout.

Cross-process behavior of `AtomicFlag` (set visibility, `wait()` unblock/timeout) verified under the `spawn` start method, so Windows semantics are covered too. All conversions are `.format()`-only / stdlib-API ≥3.2, keeping the ≥3.6 floor.

Utils/LiveViewer.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, dir_path, image=False, slideshow=False, slideshow_pause=2.0,
         self.banner_text = banner_text
         self.update_interval = update_interval
 
-        self.exit = multiprocessing.Event()
+        self.exit = AtomicFlag()
```
dvida, line 88: **Blocking:** `AtomicFlag` is used here but never imported (no `from RMS.Misc import AtomicFlag` in this file) — `NameError` at instantiation, confirmed with pyflakes. This runs in the main capture process via `StartCapture.py:726` (`live_maxpixel_enable`) and `:1511` (`slideshow_enable`), so any station with live view or the slideshow enabled crashes deterministically at capture start.

RMS/Routines/GstreamerCapture.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, device_url, gst_decoder, video_format='BGR', video_file_dir=N
 
         super(GstCaptureTest, self).__init__()
 
-        self.exit = multiprocessing.Event()
+        self.exit = AtomicFlag()
```
dvida, line 59: Same missing import as `Utils/LiveViewer.py` — `AtomicFlag` is never imported in this module, so `GstCaptureTest` raises `NameError`. Only reachable from this module's `__main__` diagnostic, but it's dead-on-arrival code.

RMS/StartCapture.py
```diff
@@ -95,6 +163,52 @@ def resetSIGINT():
...
+    if _early_station_lock is not None:
+        return _early_station_lock
+
+    import fcntl
+    import tempfile
+
+    lock_path = os.path.join(tempfile.gettempdir(),
+        'rms_startcapture_{:s}.lock'.format(str(config.stationID)))
+
+    lock_file = open(lock_path, 'w')
```
dvida: **Blocking:** the lock fd is inherited by every child forked after acquisition (logging listener, BufferedCapture, Compressor, pool workers, UploadManager, EventMonitor — Linux default start method is fork, and nothing closes the fd post-fork). So if the main process dies but any child survives as an orphan — exactly the incident's failure mode — the orphan keeps holding the flock and **every supervisor restart is refused forever** until someone manually kills it. Verified empirically: parent takes flock → forks child → parent SIGKILLed → new instance gets EAGAIN.

This inverts the commit's claim that the lock "never blocks a legitimate restart": it holds only if all descendants die too. Suggested fixes: close the lock fd in the child preambles (e.g. `os.close()` it in each `Process.run()` before the main loop, or set `FD_CLOEXEC` and use a re-exec'd holder), or write the holder PID and add a liveness check as a secondary guard. Same applies to the early lock at line 80.

RMS/Logger.py
```diff
@@ -430,9 +459,98 @@ def get_log_level(level_str, default=logging.INFO):
...
+            if (self.listener_process is not None) and self.listener_process.is_alive():
+                self.listener_process.terminate()
+                self.listener_process.join(timeout=2)
+        except Exception as e:
+            print('LOGGING WATCHDOG: could not stop the old listener: {}'.format(e))
+
+        # Abandon the old queue - it may be wedged beyond recovery. Records buffered in it are
+        # lost, and child processes forked before this point keep logging into the abandoned
+        # (bounded) queue, so their output is lost until they are restarted.
+        self._spawnListener()
```
dvida, line 699: With `log_stdout: true`, this restart path creates a recursive log-amplification loop. At initial init the listener is spawned *before* the stream redirect (lines 442 vs 457), but here the fork happens from a process whose `sys.stdout` is already a `LoggerWriter`. The new listener's console handler becomes `StreamHandler(LoggerWriter)` (line 649), so every record it handles → console emit → `LoggerWriter.write` → `root_logger.log` → the listener's own file+console handlers again, recursing until `RecursionError` is swallowed. Each record lands in the log file many times over with nested prefixes — the recovery becomes a disk/CPU storm on exactly the configs where you'd want it working.

Simplest fix: in `_listener_process`, reset `sys.stdout = sys.__stdout__` (and stderr) before `_listener_configurer` runs.

RMS/Logger.py
```diff
@@ -430,9 +459,98 @@ def get_log_level(level_str, default=logging.INFO):
...
+                self._stall_strikes = 0
+
+            if self._stall_strikes >= LOG_STALL_STRIKES:
+                self._restartLogging('backlog of {:d} records is not draining'.format(backlog))
+                return False
+
+            return True
+
+
+    def _restartLogging(self, reason):
```
dvida, line 663: `_restartLogging` has no backoff, restart cap, or cleanup of the abandoned queue. If the failure is persistent — e.g. the log dir becomes unwritable and no `safedir` was passed (`StartCapture.py` calls `initLogging(config)` without one), so `CustomHandler.__init__` raises and the listener dies instantly — the watchdog restarts every 60 s, each cycle leaking the abandoned queue's two pipe fds plus a feeder thread. At the default 1024-fd soft limit that's fd exhaustion in ~8.5 h, after which the capture process's own camera/file I/O starts failing — a resource-exhaustion outage introduced by the recovery path. Suggest: `close()` the abandoned queue (its buffered records are already forfeit), and add a restart cap or exponential backoff so a persistently failing listener degrades to console-only instead of looping.

RMS/Logger.py
```diff
@@ -430,9 +459,98 @@ def get_log_level(level_str, default=logging.INFO):
...
+            if not self.is_initialized:
+                return True
+
+            # Listener process died
+            if (self.listener_process is None) or (not self.listener_process.is_alive()):
+                self._restartLogging('listener process is dead')
+                return False
+
+            # Listener alive - check that the queue is actually draining
+            try:
```
dvida, line 615: The stall test is an absolute threshold (`backlog > 500` on 3 consecutive checks), not a drain-rate test — so a *healthy* listener draining a large burst gets killed. E.g. detection logs a 20k-record burst on a slow SD card draining ~80 rec/s: backlog goes 20,000 → 15,000 → 10,000 across three 60 s checks, always >500 though clearly draining; the watchdog then terminates the healthy listener mid-write and discards up to 30k queued records, and repeats on every burst. Comparing successive `qsize()` readings (restart only if the backlog isn't decreasing) fixes this with the state you already track in `_stall_strikes`.

RMS/Logger.py
```diff
@@ -430,9 +459,98 @@ def get_log_level(level_str, default=logging.INFO):
...
+                return False
+
+            return True
+
+
+    def _restartLogging(self, reason):
+        """ Replace the queue and listener process. Must be called with init_lock held. """
+
+        # This goes to stderr directly - the logging pipeline is the thing that is broken
+        print('LOGGING WATCHDOG: {:s} - restarting the logging pipeline...'.format(reason))
```
dvida: The comment says stderr, but `print()` writes to `sys.stdout` — and with `log_stdout: true`, stdout is a `LoggerWriter` whose queue handler at this moment still points at the *old wedged queue* (the swap happens below), so this diagnostic is silently dropped and the operator never learns why the pipeline restarted. Use `print(..., file=sys.__stderr__)` to bypass the redirect entirely.

RMS/QueuedPool.py
```diff
@@ -576,13 +644,8 @@ def updateCoreNumber(self, cores=None):
...
 
-        # Join the previous pool
-        self.printAndLog('Closing pool...')
-        self.pool.close()
-        self.printAndLog('Terminating pool...')
-        self.pool.terminate()
-        self.printAndLog('Joining pool...')
-        self.pool.join()
+        # Join the previous pool, terminating only if the workers did not exit on their own
+        self._shutdownPool()
```
dvida, line 735: Graceful exit is structurally unreachable on this path for idle workers: no poison pills are sent here, and `_workerFunc` only checks `kill_workers` *after* completing a job (line 422) — a worker blocked in `input_queue.get(True)` (line 354) never sees the flag. So the sequence is: 100 s wait above expires → `_shutdownPool()` burns its full 30 s grace → `terminate()` fires anyway — the same SIGTERM the PR set out to avoid, now with +130 s of stall, on a path `StartCapture` runs every night after capture ends. (Pills can't be naively added because the reused `input_queue` would poison the new pool's workers.) The clean fix is a `get(timeout=...)` poll loop in `_workerFunc` that re-checks `kill_workers` between timeouts.

RMS/StartCapture.py
```diff
@@ -410,12 +532,12 @@ def runCapture(config, duration=None, video_file=None, nodetect=False, detect_en
     sharedArrayBase = multiprocessing.Array(ctypes.c_uint8, 256*(config.width + array_pad)*(config.height + array_pad))
     sharedArray = np.ctypeslib.as_array(sharedArrayBase.get_obj())
     sharedArray = sharedArray.reshape(256, (config.height + array_pad), (config.width + array_pad))
-    startTime = multiprocessing.Value('d', 0.0)
+    startTime = multiprocessing.Value('d', 0.0, lock=False)
```
dvida, line 591: Heads-up on the `c_double` conversions (this one, `start_time2` below, and `start_raw_time1/2` in `BufferedCapture.py`): on 32-bit ARM (ARMv7 RPi builds) a double store/load is two 32-bit accesses, not single-copy-atomic — and these values are written on one side of a cross-process handoff while the other side polls at 10 Hz. A read that combines the new high word with the old low word during the 0→t transition yields a timestamp floored to ~1024 s granularity (the low 32 mantissa bits of an epoch-magnitude double span ~2^10 s) — accepted as a valid `startTime`, so the FF file / raw block gets mistimed by up to ~17 minutes, silently corrupting multi-station trajectory correlation. Previously the implicit lock made each access atomic. Nanosecond window × 10 Hz polls = astronomically rare per station, but nonzero across a 24/7 fleet. If 32-bit builds are still supported, these four probably deserve to keep a lock (they're not on any path that can wedge shutdown) or use a seqlock-style double-read validation. The `c_bool` conversions are all fine (single byte).

RMS/BufferedCapture.py
```diff
@@ -200,10 +200,12 @@ def __init__(self, array1, start_time1, array2, start_time2, config, video_file=
...
 
         # Flag for process control
-        self.exit = Event()
+        self.exit = AtomicFlag()
 
         # Heartbeat timestamp for watchdog - updated every frame block to detect hangs
-        self.heartbeat = Value('d', 0.0)
+        # lock=False: single writer, and a locked Value can deadlock the watchdog if this
+        # process is killed while holding the lock
+        self.heartbeat = Value('d', 0.0, lock=False)
```
dvida, line 214: Same ARM32 torn-read caveat as the `startTime` doubles: a tear across the upper-word rollover can make a single heartbeat read look up to ~1024 s stale (> `HEARTBEAT_TIMEOUT` of 180 s), and the watchdog acts on one read with no confirming re-read — spuriously SIGKILLing a healthy capture. Self-healing and very rare, but a one-line "re-read and confirm before declaring stale" in the watchdog eliminates it entirely.

RMS/UploadManager.py
```diff
@@ -547,7 +547,9 @@ def __init__(self, config):
         # Construct the path to the queue backup file
         self.upload_queue_file_path = os.path.join(self.config.data_dir, self.config.upload_queue_file)
 
-        self.exit = multiprocessing.Event()
+        # Lock-free flag: stop() escalates to terminate(), which must not be able to
+        # orphan an Event lock (see AtomicFlag)
+        self.exit = AtomicFlag()
```
dvida, line 552: The sweep stopped one line short: `upload_in_progress` (line 553, directly below) is still a *locked* `Value`, written by the child and polled by the main process at `StartCapture.py:1376` to gate the pre-reboot wait. `stop()` escalates to `terminate()` (line 600) and the process is OOM-killable — a kill mid-assignment orphans the semaphore and the reboot poll blocks forever. It's a set/poll bool squarely in this commit's stated scope; same `AtomicFlag` conversion applies.

Related, one step harder: `last_runtime`/`next_runtime` and their explicit `Lock()`s (lines 569-574) are acquired by the child roughly every second in the run loop and by the *parent* in `addFiles()`/`delayNextUpload()` (called from `StartCapture` several times per night). A terminate/OOM while the child is inside one of those `with` blocks wedges next night's `addFiles()` in the main process. These need either bounded acquisition (like `SafeValue`) or a lock-free handoff design.

RMS/Misc.py
```diff
@@ -45,6 +47,53 @@
...
+    def set(self):
+        self._flag.value = 1
+
+    def clear(self):
+        self._flag.value = 0
+
+    def is_set(self):
+        return bool(self._flag.value)
+
+    def wait(self, timeout=None, poll_interval=1.0):
```
dvida, line 134: Minor: `wait()` measures the timeout with `time.time()` — an NTP step (routine on Pi boot, and RMS stations force clock sync) stretches or truncates it. `time.monotonic()` is the right clock here and exists on all supported Pythons (3.3+). Also worth noting in the docstring that wake-up latency after `set()` is up to `poll_interval` (1 s default) vs `Event`'s immediate wake — fine for the current call sites, but a surprise for future ones.

dvida (changes requested):
Thanks — this is a genuinely impressive post-mortem and the architecture of the fix is right: bounding the queue, drop-don't-block, lock-free flags for set/poll, and a kernel-released instance lock are all the correct calls. I did a deep review (four independent passes over the subsystems, with the key claims re-verified against the code and a couple of live experiments). Much of it checks out cleanly: the `AtomicFlag` core is sound (single shared byte, `wait()` semantics match `Event.wait`), the drop-on-full handler genuinely can't block or raise (verified against CPython's `Queue._feed`), no `multiprocessing.Event` remains anywhere in `RMS/`/`Utils/`, `SafeValue` never releases a lock it didn't acquire, the flock guard is import-safe and a refused duplicate really does exit in ~50 ms, and the commits are independent as claimed.

That said, there are two blocking problems and a few places where the sweep missed instances of the exact hazard class it targets. Inline comments carry the details; the items below have no diff line to anchor to:

**Missed locked `Value`s (same class as the `Compressor.stop()` wedge):**
- `RMS/BufferedCapture.py:198` — `self.dropped_frames = Value('i', 0)` is a locked `Value`. The child does `.value +=` regularly (line 2218 and three more sites), and `stopCapture()` SIGKILLs the child (line 278) then reads `.value` (line 296). A kill landing while the child holds the implicit semaphore wedges the main process at shutdown — identical to the original `Compressor.stop()` wedge.
- `RMS/UploadManager.py:553` and `:569-574` — see inline comment.

**Pre-existing QueuedPool hangs that now gate the new graceful path** (fine as follow-ups, but worth flagging since the PR's goal is bounded shutdown):
- The pill-swallow loop (`closePool`, lines 592-601) has no wall-clock escape. A worker OOM-killed mid-job permanently skews `available_workers`/`active_workers` (decrement at 374 never balanced), the pool respawns a replacement that re-increments `active_workers`, and the loop's only exits can then never fire — `closePool` hangs forever *before* reaching the new `_shutdownPool`. This is precisely the killed-worker scenario the PR targets.
- The stuck-worker path (lines 524-538) still terminates all N workers immediately, with no pills or grace for the N−1 healthy ones — the original mid-write kill hazard, retained.
- `pool.terminate()` itself is unbounded: CPython's `_terminate_pool` joins each worker with no timeout, so a worker in uninterruptible I/O (hung camera/USB) hangs shutdown forever. There's no SIGKILL escalation anywhere in the file.

**Smaller points:**
- One orphaned `SafeValue` lock makes every operation in every process pay the full 5 s timeout + a `print()` forever (no memoization/lock recreation) — `closePool` worst case stretches to ~15 minutes with hundreds of console lines. Consider caching the failure or shortening subsequent timeouts.
- The early pre-import lock never engages for the default deployment: `Scripts/RMS_StartCapture.sh` launches without `-c <file>`, and the early peek requires an explicit config file path — so default stations pay the full heavy-import cost before the post-config guard refuses. Peeking `./.config` when no `-c` is given would close this.
- The refusal message goes to stdout, not stderr — a supervisor capturing stderr sees a bare exit 1.
- `open(lock_path, 'w')` sits outside the try block: a stale lock file owned by a different user (sticky /tmp) raises a raw `PermissionError` instead of the clean refusal, and `'w'` truncates the holder's PID. Also note `tempfile.gettempdir()` interacts badly with systemd `PrivateTmp=true` (a service instance and a manual debug instance can't see each other's locks → both run).
- `checkLoggingHealth()` runs only inside `wait()`'s watchdog block, so the multi-hour post-capture/upload phases are unmonitored — a stall there stays log-dark until the next capture session.
- With `total_jobs == 0` and `delay_start > 0`, pills are counted from `active_workers` before any worker has incremented it — zero pills are sent and graceful shutdown silently degrades to the 30 s-timeout terminate.
- Dropped log records are completely unaccounted (`_DroppingQueueHandler.enqueue` bare `pass`) — a simple counter with a periodic "N records dropped" line would make loss visible.

**Tests:** the PR description details an end-to-end test suite (watchdog restart, starvation strikes, flock SIGKILL case) but the diff adds no test files. Could you commit them? The missing-import crash in `Utils/LiveViewer.py` is exactly the kind of thing an import smoke test would have caught, and without the tests in-repo the verification isn't repeatable by reviewers or CI.

Happy to help with any of the mechanical fixes (the missing imports and the three `Value` conversions in particular).

Cybis320:
All review items are addressed in 9 commits (d4953602..4ce6b263), each mapped to a finding:

**Blocking**
- `AtomicFlag` imports added in `LiveViewer` and `GstreamerCapture` (d4953602).
- Station lock survives orphaned children: the fd is closed in every forked child via `register_after_fork`, so the parent is the sole holder and the kernel releases on its death (8f629ff9). Verified empirically with your protocol — parent takes lock, forks child, parent SIGKILLed, orphan alive, new instance acquires immediately — and committed as a test.

**Missed locked `Value`s**
- `BufferedCapture.dropped_frames` → `lock=False` (the capture child is the sole writer — including the day/night reset, which lives in the child's loop; int32 access is single-copy atomic incl. ARMv7) (4ce6b263).
- `UploadManager.upload_in_progress` → `AtomicFlag`; `last_runtime`/`next_runtime` keep their mutual exclusion via a new `BoundedLock` in `RMS.Misc` (bounded acquire, warn once, proceed) (530eb2f8).

**QueuedPool** (caf305cb)
- Workers use a timed `get()` and honor `kill_workers` when idle — graceful exit is now reachable on every path, including the nightly post-capture close.
- The stuck-worker branch no longer terminates all N: `_shutdownPool` sets the flag, healthy workers exit within their timeout, and only the genuinely stuck one meets terminate — which now escalates to SIGKILL before the previously unbounded `join()`.
- Pills are counted from the real spawned-worker list (fixes `total_jobs == 0` + `delay_start`); the pill-swallow loop gets an unconditional wall-clock escape (the skewed-counter hang you traced).
- `SafeValue` memoizes a broken lock: one full-timeout warning, then a 0.5 s fast path.

**Logger** (35c9f437)
- Drain-rate stall test — strike only on a high *and* non-shrinking backlog, so a listener draining a burst lives.
- Restart backoff (2 min doubling to a 1 h cap) plus closing our end of the abandoned queue — bounded fds under a persistent failure.
- The restarted listener builds its console handler on `sys.__stdout__` — kills the `log_stdout` recursion you found.
- Watchdog diagnostics via `sys.__stderr__`; dropped records are counted per process and reported through the recovered pipeline.
- The manager runs an always-on health thread, so the multi-hour post-capture phases are monitored too.

**ARM32 torn doubles** (568079a5, 6f6292c8)
- The heartbeat watchdog re-reads before declaring a hang; both start-time consumers use `stableDoubleRead` (two consecutive equal reads cannot both be torn by one in-flight store).
- `AtomicFlag.wait` measures on `time.monotonic`, with the poll-latency note in the docstring.

**Tests** (c6f0ab9e): committed as `Tests/test_DeadlockFixes.py` — 12 tests covering the drop-on-full handler, orphaned-lock SafeValue/BoundedLock, AtomicFlag cross-process semantics, the flock guard against the real AST-extracted StartCapture functions (duplicate refusal, SIGKILLed holder, and the blocking orphaned-child case in a fresh interpreter), and both pool shutdown paths. All green alongside the existing suite.

Not addressed here, per your "fine as follow-ups": the per-connection logging transport (`SocketHandler`) for children forked before a restart — noting the practical exposure is bounded, since children recycle at every day/night switch and pool cycle, so post-restart loss is hours at worst on an event the rest of this PR makes rare — and the `oom_score_adj` question.

Ready for re-review.

Cybis320:
Another machine, CAC0A3, has experienced an OOM on prerelease, preceded by logging failing and a cascade of StartCapture being launched, which would be fixed by this PR. This branch has been running for 9 days continuously on a very troublesome Multicam machine and shows no sign of memory leak. And another branch which incorporates these fixes has been running on 10 other machines including a RPI4 with no issues.

dvida (commented):
Re-reviewed the final state against every item from my last pass. Both blocking findings are properly fixed, and the conversion sweep holds up: no `multiprocessing.Event` left in `RMS/`/`Utils/`, every shared `Value` on a kill path is `lock=False`, and the `Array`s are only ever touched through `get_obj()` numpy views so their implicit locks are never taken. The flock inversion is genuinely closed — `register_after_fork` does fire in `Process._bootstrap`, and `subprocess` closes fds by default, so the parent really is the sole holder. The early lock peeking `./.config` makes it engage for default stations, and `'a+'` + owner-only truncate fixes the stale-file and PID-clobbering cases. Together with the 9-day multicam soak and the RPi4 fleet, I'm happy to take this.

A few things are still open. None of them block merging.

**1. The SIGKILL escalation in `_shutdownPool` is unreachable.** `self.pool.terminate()` runs at `QueuedPool.py:505` and the escalation loop sits *after* it at `:519` — but CPython's `Pool._terminate_pool` ends with an unbounded `p.join()` over every worker still alive, so a worker that ignores SIGTERM hangs *inside* `terminate()` and never reaches the loop. My original "`pool.terminate()` is unbounded, there is no SIGKILL escalation anywhere" point therefore still stands. Reordering it — bound the wait, SIGKILL the survivors, *then* `terminate()`/`join()`, which is what `BufferedCapture.stopCapture()` already does — closes the SIGTERM-ignoring case. (A worker in uninterruptible I/O is unkillable either way; that one needs the join bounded and the process abandoned.) `Compressor.stop()` has the same shape: `terminate()` followed by a bare `join()`, no escalation.

**2. The stall test can still restart a healthy listener.** The drain-rate comparison fixes the draining case, but not a *growing* backlog: producers can legitimately outrun a healthy listener (a detection burst on a slow SD card), and `backlog >= previous` is then true on every check. With two callers on 60 s cycles — `wait()` and the new always-on health thread — three strikes can land within about a minute, terminating a healthy listener mid-write and discarding up to 30k queued records.

I have a patch for this ready on my side: the listener publishes a lock-free count of the records it has taken off the queue, which separates the two cases exactly (it stops advancing only when nothing is being consumed), a strike requires a high backlog *and* no progress since the previous sample, and samples are spaced at least 45 s apart. Detection latency is unchanged at ~3 checks. Tests included for all three cases. Happy to push it here or leave it to you.

**3. Smaller points**

- `_restartLogging`: `Queue.close()` only appends a sentinel to the feeder thread's buffer. If that thread is blocked on the orphaned write lock — i.e. the actual wedge — it never exits, so its writer fd, the thread and its buffered records are retained. The comment's "releases this process's two pipe fds and its feeder thread" overstates it. The new backoff bounds the leak to something harmless, so I'd just soften the comment.
- Dropped-record counts are per-process, but only the process that ran `initLogging` ever calls `checkLoggingHealth`, so child drops are never reported — and the children were the heavy producers in the incident.
- `StartCapture.py:304` is the one watchdog diagnostic still on a plain `print()`, so with `log_stdout: true` it goes into the LoggerWriter, i.e. the wedged pipeline it is reporting on. Should be `file=sys.__stderr__` like the rest.
- The early lock runs before argparse, so `python -m RMS.StartCapture -h` now exits 1 while a station is capturing. Correct for the duplicate-capture case, collateral for `--help`.
- Latent: the always-on health thread means `init_lock` (a plain `threading.Lock`) can now be held at fork time, and a child inheriting it locked would deadlock on any `LoggingManager` call. Nothing forked calls one today — every `initLogging` is in a `__main__` block — so it is only worth a comment so nobody adds one later.
- The tests cover the primitives well, but nothing exercises `checkLoggingHealth`, `_restartLogging`, the backoff or the `log_stdout` recursion fix, which is the most intricate new code and the part that runs unattended. The patch in (2) adds coverage for the stall path.

Deferring the `SocketHandler` transport and `oom_score_adj` as agreed.

RMS/Logger.py
```diff
@@ -353,6 +353,49 @@ def doRollover(self):
...
+# Upper bound on records buffered between producers and the listener. A stalled listener pipeline
+# otherwise grows every producer's queue feeder buffer without bound (observed as ~700 MB/process/day
+# in production when a process died while holding the queue's shared write lock)
+LOG_QUEUE_MAXSIZE = 30000
+
+# Backlog size above which the listener pipeline is considered stalled
+LOG_QUEUE_STALL_THRESHOLD = 500
+
+# Number of consecutive health checks the backlog must persist before the pipeline is restarted
+LOG_STALL_STRIKES = 3
```
dvida, line 365: The new spacing constant.

```suggestion
LOG_STALL_STRIKES = 3

# Minimum spacing between two strike-eligible samples. Both the capture watchdog and the
# manager's own health thread call the check every 60 s, so without this the samples can land
# seconds apart and all LOG_STALL_STRIKES accumulate inside a single burst
LOG_STALL_SAMPLE_INTERVAL = 45
```

RMS/Logger.py
```diff
@@ -361,6 +404,21 @@ def __init__(self):
         self.is_initialized = False
         self.init_lock = threading.Lock()
 
+        # Arguments used to spawn the listener, kept so a wedged listener can be respawned
+        self._listener_args = None
+
+        # Consecutive health checks with an excessive, non-draining backlog
+        self._stall_strikes = 0
+        self._last_backlog = None
```
dvida: State swap: the previous backlog reading is replaced by the previous progress count and the time it was taken.

```suggestion
        # Consecutive health checks with an excessive backlog and no listener progress
        self._stall_strikes = 0
        self._last_processed = None
        self._last_sample_time = None

        # Records the current listener has taken off the queue (lock-free, single writer)
        self._records_processed = None
```

RMS/Logger.py
```diff
@@ -430,21 +483,200 @@ def get_log_level(level_str, default=logging.INFO):
...
             atexit.register(self.shutdownLogging)
 
+
+    def _spawnListener(self):
+        """ Create a fresh (bounded) logging queue and start a listener process draining it. """
+
+        self.logging_queue = multiprocessing.Queue(LOG_QUEUE_MAXSIZE)
+        self.listener_process = multiprocessing.Process(
+            target=_listener_process,
+            args=(self.logging_queue,) + self._listener_args
```
dvida: The counter is created fresh with every listener, so a restart starts from zero and a stale count can never be mistaken for a stall.

```suggestion
        self.logging_queue = multiprocessing.Queue(LOG_QUEUE_MAXSIZE)

        # Progress counter for the health check. Created fresh with every listener, so a
        # restart starts from zero. lock=False: the listener is the only writer and the health
        # check only ever compares it against its own previous reading, so a locked Value
        # would add a semaphore a dying listener could orphan for no benefit
        self._records_processed = multiprocessing.Value('l', 0, lock=False)

        self.listener_process = multiprocessing.Process(
            target=_listener_process,
            args=(self.logging_queue, self._records_processed) + self._listener_args
```

RMS/Logger.py
```diff
@@ -430,21 +483,200 @@ def get_log_level(level_str, default=logging.INFO):
...
+                          self._next_restart_allowed - now),
+                          file=sys.__stderr__)
+                    return False
+                self._restartLogging('listener process is dead')
+                return False
+
+            # A long stretch of health clears the restart pacing
+            if self._restart_count and (now > self._next_restart_allowed
+                    + LOG_RESTART_BACKOFF_MAX):
+                self._restart_count = 0
```
dvida, line 604: Sample spacing. Without it the watchdog call and the health thread can land seconds apart and three strikes accumulate inside a single burst.

```suggestion
                self._restart_count = 0

            # Space the samples out: this check has two callers (the capture
            # watchdog and the always-on health thread), each on its own 60 s
            # cycle, so back-to-back samples are possible and three strikes
            # could otherwise be collected within a single burst
            if (self._last_sample_time is not None) \
                    and ((now - self._last_sample_time) < LOG_STALL_SAMPLE_INTERVAL):
                return True
```

RMS/Logger.py
```diff
@@ -430,21 +483,200 @@ def get_log_level(level_str, default=logging.INFO):
...
+            # slowly draining a large burst must not be killed mid-write
+            # (review finding). Strike only when the backlog is high AND not
+            # shrinking against the previous reading.
+            if (backlog > LOG_QUEUE_STALL_THRESHOLD) \
+                    and (self._last_backlog is not None) \
+                    and (backlog >= self._last_backlog):
+                self._stall_strikes += 1
+            else:
+                self._stall_strikes = 0
+            self._last_backlog = backlog
```
dvida: The core of it: strike on a high backlog with no listener progress, instead of on a backlog that is not shrinking. A listener that consumed anything cannot strike, so both the slow-drain case and the growing-burst case are safe.

```suggestion
            processed = self._records_processed.value

            # A stall is the listener not CONSUMING - not a large backlog, and not even a
            # backlog that keeps growing: producers can legitimately outrun the listener for
            # a while (a detection burst on a slow SD card), and killing it then discards
            # everything queued. The listener's own progress counter separates the two cases
            # exactly: it stops advancing only when nothing is being taken off the queue.
            if (backlog > LOG_QUEUE_STALL_THRESHOLD) \
                    and (self._last_processed is not None) \
                    and (processed == self._last_processed):
                self._stall_strikes += 1
            else:
                self._stall_strikes = 0

            self._last_processed = processed
            self._last_sample_time = now
```

dvida (commented):
Here is the stall-detection patch from my last comment as accept-or-reject suggestions (item 2).

The idea: the queue backlog alone cannot tell a wedged pipeline from producers outrunning a healthy listener, so the listener publishes a lock-free count of the records it has taken off the queue. A strike now needs a high backlog **and** no progress since the previous sample, and samples are spaced at least 45 s apart because the check has two callers on 60 s cycles (`wait()` and the health thread) and back-to-back calls could otherwise collect all three strikes inside one burst. Detection latency is unchanged at three checks.

Seven suggestions on `RMS/Logger.py` plus one on the test file below. **Two more edits are needed on the listener side**, which GitHub will not let me post as suggestions because those lines are not part of this PR's diff — they have to be applied by hand:

```diff
-def _listener_process(queue, config, log_file_prefix, safedir, console_level=logging.INFO, file_level=logging.DEBUG):
+def _listener_process(queue, records_processed, config, log_file_prefix, safedir,
+                      console_level=logging.INFO, file_level=logging.DEBUG):
     """ Target function for the logging listener process.
     Ignores SIGINT and processes messages in strict FIFO order.
+
+    records_processed is a lock-free shared counter of the records taken off the queue. It is
+    the health check's only reliable "the listener is consuming" signal - the queue backlog
+    alone cannot distinguish a wedged pipeline from producers outrunning a healthy listener.
     """
```

```diff
     while True:
         try:
             record = queue.get()
+
+            # Count progress as soon as the record leaves the queue: the backlog the health
+            # check reads shrinks at this point, so the two readings stay consistent even if
+            # a handler write is slow
+            records_processed.value += 1
+
             if record is None:       # shutdown sentinel
                 break
```

Those two are not optional - `_spawnListener` passes the counter positionally, so taking the suggestions without them breaks the listener. Apply them first, then the suggestions.

Verified on my side: the three new tests pass alongside the existing twelve, and a live run of 200 records through a real `initLogging` gives `records_processed == 201` (the 200 plus the init line), backlog 0, health check clean. Reject freely if you would rather keep the backlog-only heuristic - but as it stands a detection burst that outpaces the listener for a minute will restart it and discard the queue.

RMS/Logger.py
```diff
@@ -430,21 +483,200 @@ def get_log_level(level_str, default=logging.INFO):
...
+
+            if self._stall_strikes >= LOG_STALL_STRIKES:
+                if now < self._next_restart_allowed:
+                    print('LOGGING WATCHDOG: backlog stalled; next restart '
+                          'attempt in {:.0f} s'.format(
+                          self._next_restart_allowed - now),
+                          file=sys.__stderr__)
+                    return False
+                self._restartLogging('backlog of {:d} records is not draining'.format(backlog))
```
dvida: Message matches what is now being detected.

```suggestion
                self._restartLogging('backlog of {:d} records with no listener progress in '
                    '{:d} checks'.format(backlog, LOG_STALL_STRIKES))
```

RMS/Logger.py
```diff
@@ -430,21 +483,200 @@ def get_log_level(level_str, default=logging.INFO):
...
+            pass
+
+        self._restart_count += 1
+        self._next_restart_allowed = time.monotonic() + min(
+            LOG_RESTART_BACKOFF_BASE*(2**(self._restart_count - 1)),
+            LOG_RESTART_BACKOFF_MAX)
+
+        self._spawnListener()
+        self._stall_strikes = 0
+        self._last_backlog = None
```
dvida: Reset the new state on restart.

```suggestion
        self._last_processed = None
        self._last_sample_time = None
```

Tests/test_DeadlockFixes.py
```diff
@@ -0,0 +1,343 @@
...
+
+    from RMS.QueuedPool import QueuedPool
+
+    qp = QueuedPool(_double, cores=2, log=None, delay_start=2, print_state=False)
+    qp.startPool()
+
+    t_beg = time.monotonic()
+    qp.closePool()
+    assert (time.monotonic() - t_beg) < 30
+    qp.shutdownManager()
```
dvida, line 343: Coverage for the three cases: a growing backlog with a consuming listener must never restart, a frozen listener must restart after three strikes, and back-to-back calls must count once. All use a stubbed manager, so no listener is spawned.

```suggestion
    qp.shutdownManager()


# ---------------------------------------------------------------------------
# Logging stall detection


class _FakeQueue(object):
    """ Serves a scripted sequence of qsize() readings, holding the last one. """

    def __init__(self, sizes):
        self.sizes = list(sizes)

    def qsize(self):
        return self.sizes.pop(0) if len(self.sizes) > 1 else self.sizes[0]


class _FakeCounter(object):
    """ Serves a scripted sequence of listener progress counts. """

    def __init__(self, counts):
        self.counts = list(counts)

    @property
    def value(self):
        return self.counts.pop(0) if len(self.counts) > 1 else self.counts[0]


class _AliveListener(object):
    def is_alive(self):
        return True


def _stallManager(backlogs, processed):
    """ A LoggingManager wired to scripted readings, with the restart stubbed out. """

    from RMS.Logger import LoggingManager

    mgr = LoggingManager()
    mgr.is_initialized = True
    mgr.listener_process = _AliveListener()
    mgr.logging_queue = _FakeQueue(backlogs)
    mgr._records_processed = _FakeCounter(processed)

    mgr.restarts = []
    mgr._restartLogging = lambda reason: mgr.restarts.append(reason)

    return mgr


def _sample(mgr):
    """ Take one strike-eligible sample (pretend enough time has passed since the last). """

    mgr._last_sample_time = None
    mgr.checkLoggingHealth()


def testLoggingStallIgnoresGrowingBacklogWhileListenerConsumes():
    """ A backlog that keeps GROWING is not a stall if the listener is still taking records
        off the queue - producers can legitimately outrun it during a burst, and restarting
        then discards everything queued. """

    mgr = _stallManager(backlogs=[1000, 5000, 12000, 20000, 25000],
                        processed=[500, 4000, 9000, 15000, 21000])

    for _ in range(5):
        _sample(mgr)

    assert mgr.restarts == []
    assert mgr._stall_strikes == 0


def testLoggingStallRestartsOnFrozenListener():
    """ A high backlog with a listener that consumes nothing is the real wedge. """

    mgr = _stallManager(backlogs=[5000], processed=[77])

    for _ in range(1 + 3):      # first sample only establishes the baseline
        _sample(mgr)

    assert len(mgr.restarts) == 1
    assert 'no listener progress' in mgr.restarts[0]


def testLoggingStallSamplesAreSpacedOut():
    """ Both the capture watchdog and the always-on health thread call the check, so
        back-to-back calls must not each count as a strike. """

    mgr = _stallManager(backlogs=[5000], processed=[77])

    _sample(mgr)                    # baseline
    for _ in range(10):             # no clock advance: all of these must be ignored
        mgr.checkLoggingHealth()

    assert mgr._stall_strikes == 0
    assert mgr.restarts == []

    _sample(mgr)
    assert mgr._stall_strikes == 1
```

Cybis320:
All items from the re-review are addressed in 3 commits (ce014a4e..62fa4953):

**Stall-detection patch (ce014a4e)**
- Applied as posted: the two manual `_listener_process` edits first (`records_processed` counter, incremented as each record leaves the queue), then all seven `Logger.py` suggestions and the test-file suggestion. All three stall tests pass alongside the existing suite (16 total).

**1. Unreachable SIGKILL escalation (26a6659e)**
- `_shutdownPool` reordered as suggested: graceful wait → per-worker SIGTERM → bounded 10 s wait → SIGKILL survivors → bounded reap wait → `terminate()`/`join()` only once every worker is confirmed dead. A worker that survives SIGKILL (uninterruptible I/O) is abandoned with a log line instead of wedging shutdown.
- `Compressor.stop()`: the bare `join()` after `terminate()` is now `join(5)` + SIGKILL escalation + reaping join.

**3. Smaller points**
- `_restartLogging` comment softened to best-effort: a feeder blocked on the wedged lock keeps its fds/thread/records, and the restart backoff is what bounds the leak (ce014a4e).
- Child drop counts are no longer invisible: `_DroppingQueueHandler` enqueues a warning record as soon as the queue has room again, so every producer (including forked children) self-reports; if the report itself doesn't fit, the count is re-banked rather than lost (ce014a4e).
- `StartCapture.py` watchdog diagnostic → `print(..., file=sys.__stderr__)` (62fa4953).
- `-h`/`--help` now bypasses the early station lock — argparse only prints and exits, so a running station no longer turns `--help` into an exit-1 refusal (62fa4953).
- `init_lock` carries the fork-hazard comment for future callers (ce014a4e).
- Coverage: added a backoff test that runs the real `_restartLogging` with only the spawn stubbed — a listener dying instantly on every respawn restarts once, then holds through the backoff window (ce014a4e). The `log_stdout` recursion fix remains untested; it needs a real listener with redirected streams, so I'd rather cover it in a follow-up integration test than block here.

`SocketHandler` transport and `oom_score_adj` remain deferred as agreed.

PR #937, opened 2026-07-23
Fix QueuedPool multiprocessing crash on Windows (spawn)

dvida:
## What was broken

On Windows, running SkyFit2 on a folder with no CALSTARS file tries to generate one automatically. This crashed with:

```
ERROR: Star extraction failed: cannot pickle 'weakref' object
OSError: [WinError 6] The handle is invalid
```

and no CALSTARS file was saved.

## Why

Windows and Linux start worker processes differently:

- **Linux** uses `fork` — the child is a memory copy of the parent, so nothing needs to be sent across.
- **Windows** uses `spawn` — the child is a fresh process, so Python has to *pickle* (serialize) everything the worker needs and ship it over.

`QueuedPool` starts its pool using a bound method (`self._workerFunc`) as the initializer. Pickling a bound method pickles the **entire `QueuedPool` object**. That object holds a `multiprocessing.Manager`, which internally contains a `weakref` — and weakrefs can't be pickled. So on Windows the pool never even started.

This was never seen before because the capture pipeline runs on Linux (Raspberry Pi), where `fork` skips pickling entirely. Only the Windows desktop tools hit it: SkyFit2 CALSTARS generation, FluxBatch, and TrackStack.

## The fix

Add `__getstate__` / `__setstate__` to `QueuedPool` that leave out the two things that can't (and don't need to) be pickled: the `Manager` object and the live pool handle. The workers never use either — they talk to the queue proxies, which pickle fine. Everything else the workers need serializes correctly during a real spawn.

**No effect on Linux:** `fork` never pickles, so these methods are never called there. The 24/7 capture pipeline is completely untouched.

## Testing

- Emulated Windows `spawn` on Linux (`set_start_method('spawn', force=True)`) and ran a `QueuedPool` end to end:
  - **Before:** `cannot pickle 'weakref' object`
  - **After:** pool starts, all jobs processed, results returned.
- Still to confirm on real Windows 11 / Python 3.8 via SkyFit2 (multi-file folder → CALSTARS generated), plus FluxBatch and TrackStack.

## Files

- `RMS/QueuedPool.py` — added `__getstate__` / `__setstate__`. One file, ~27 lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #938, opened 2026-07-23
Python 3.14 support: multiprocessing start-method safety and NumPy 2 audit

Cybis320:
## Motivation

Python 3.14 changes the default multiprocessing start method on Linux from `fork` to `forkserver`, which pickles everything sent to child processes instead of inheriting it. Much of the capture pipeline assumed `fork` semantics and breaks under pickling start methods — the same breakage Windows/macOS users already hit under `spawn` (e.g. the SkyFit2 CALSTARS failure fixed for `QueuedPool` in #937). This branch makes the pipeline safe under all three start methods and audits NumPy 2 compatibility for 3.13+.

## What's included

**Multiprocessing safety (fork/forkserver/spawn):**
- `QueuedPool`: `__getstate__`/`__setstate__` drop the SyncManager/Pool handles and reduce the logger to its name (superset of #937, resolved in the merge); workers re-attach logging via the shared queue in `_workerFunc`
- Frame and timestamp buffers shared via `multiprocessing.Array` base objects; numpy views rebuilt per-process (`frameBufferShape()` as single source of truth). Fixes blank FFs and a `RawFrameSaver.stop()` crash under forkserver/spawn
- `EventMonitor` opens its sqlite connection in `run()`; `UploadManager` drops its SyncManager on pickle; `RunExternalScript` re-imports the user module in the child
- Children re-attach logging and ignore SIGINT via `initChildProcess()`; `getLoggingQueue()` reads the root logger's QueueHandler so it works regardless of which LoggingManager configured logging
- `RecomputeCollectionAreas`: picklable `functools.partial` worker; consume the `imap_unordered` iterator so jobs actually dispatch (was silently a no-op)

**Start-method policy — no fleet behavior change:**
- `setMultiprocessingStartMethod()` pins each platform's existing default: **`fork` on Linux through 3.13** (preserves copy-on-write memory sharing — forkserver/spawn workers re-import the whole RMS stack, costing tens of MB of private RSS and seconds of import time per worker on small stations), `forkserver` on Linux from 3.14 (the upstream default there), `spawn` on Windows/macOS. Explicit `preferred=` override available for testing
- Pinned in all pool entry points: `StartCapture`, `Reprocess`, `ExtractStars`, `DetectStarsAndMeteors`, `FluxBatch`, `TrackStack`

**NumPy 2 / Python 3.13+:**
- `numpy>=2.3` required on Python 3.13+ (numpy 1.26 never supported 3.13+; on 3.14 its temporary-elision heuristic silently corrupts large-array arithmetic)
- Replaced removed `.tostring()` calls; `np.import_array()` added to Cython modules

**Tests:**
- `Tests/TestForkserverPickling.py`: spawn-context pickling tests for `QueuedPool` (the class regressed on spawn platforms once before with zero detection), worker logging re-attach, start-method selection matrix

## Merge with prerelease

`prerelease` is merged in. Notable resolutions:
- `RMS/QueuedPool.py`: this branch's `__getstate__` kept (superset of #937 — adds the logger-name reduction for Python <= 3.6)
- `RMS/RawFrameSave.py`: both sides' `run()` preambles composed — `setParentDeathSignal()` + parent-PID liveness probe (#926 orphan/OOM fix) AND `initChildProcess()` + `ensureViews()` (forkserver view rebuild). The fixes are complementary: PDEATHSIG fires on the wrong parent under forkserver, which the liveness probe covers
- `RMS/Misc.py`, `RMS/BufferedCapture.py`: both sides' additions kept (disjoint)

## Testing

- `Tests/TestForkserverPickling.py`: 6 passed, 2 skipped (Linux-only skips) on macOS/py3.12
- `QueuedPool` end-to-end under explicit `spawn` and `fork` contexts: all jobs processed, results correct
- Start-method selection verified for linux/3.8, linux/3.13 (`fork`), linux/3.14 (`forkserver`), darwin (`spawn`)
- Linux <= 3.13 stations see no start-method change by construction; the forkserver path would benefit from an overnight soak on a real station running 3.14 before that becomes common


dvida (commented):
Rigorous review of this branch against `prerelease`. Findings were verified by reading the merged branch state (not just the diff) and, where possible, by executing the affected code path locally.

## Verdict

The core design is right, and the risk to the currently deployed fleet is genuinely near-zero — Linux ≤ 3.13 stays on `fork` by construction, so the start-method policy is a no-op on every platform today. Two findings should be resolved before merge; the rest are follow-ups.

What this PR gets right:

- Passing `multiprocessing.Array` **base** objects and rebuilding the numpy views per-process is the correct fix, and `frameBufferShape()` as the single source of truth removes duplicated pad arithmetic between `StartCapture.runCapture` and `BufferedCapture.__main__`.
- `getLoggingQueue()` reading the root logger's `QueueHandler` is the right authority — `StartCapture`/`Reprocess` really do bypass the global `LoggingManager`, so the old lookup returned `None`.
- `RawFrameSaver.ensureViews()` correctly handles the cross-process case (`stop()` runs in the capture process, `run()` in the saver process).
- Both mechanical sweeps are complete: `np.import_array()` is in **all 7** `.pyx` files that `cimport numpy`, and `.tostring()` is gone from **all 4** call sites (`Formats/AST.py` ×3, `Formats/Vid.py` ×1).
- `RecomputeCollectionAreas`: correct diagnosis. The old code passed `tqdm(updateCollectionAreaNight)` as the callable and never consumed the iterator, so it really was a silent no-op.

## Blocking

### B1 — `dtype=bool` inside `@njit` breaks thresholding

`RMS/Routines/Image.py:288` changes `np.zeros((height, width), dtype=np.bool_)` to `dtype=bool`, inside `applyThresholdNumba`, which is `@njit`-decorated. Numba cannot type the Python builtin `bool` as a dtype. Reproduced locally on numpy 1.26.4 / numba 0.63.1:

```
During: resolving callee type: Function(<built-in function empty>)
  def impl(shape, dtype=float):
      arr = np.empty(shape, dtype=dtype)
During: Pass nopython_type_inference
```

`applyImgThreshold` is bound to `applyThresholdNumba` whenever numba imports (`Image.py:301`), and it is the only implementation `thresholdImg()` calls (`Image.py:342`). So on any machine with numba installed, thresholding raises a `TypingError` on first call. numba is not in `requirements.txt`, but it is present in developer/conda envs.

The change is also unnecessary — `np.bool_` still exists in NumPy 2. Only the `np.bool` alias was removed (1.24) and it was reinstated in 2.0. Suggest reverting this one line.

### B2 — `Extractor` still receives the frame buffer by value

`RMS/Compression.py:395` calls `extractor.start(frames, compressed, filename_millis)` where `frames` is `self.array1`/`self.array2` — a numpy view over shared memory. `Extractor.start()` (`VideoExtraction.py:255`) assigns `self.frames = frames` and then calls `Process.start()`. Under `forkserver`/`spawn` that pickles the entire 256-frame block by value through the pipe: `256 × (H+pad) × (W+pad)` bytes, roughly 236 MB at 1280×720, once per FF block (every ~10 s at 25 fps) whenever `enable_fireball_detection` is on.

This is the same bug class fixed here for `Compressor` and `RawFrameSaver`, left unfixed in the place carrying the largest payload. It is *correct* (the extractor only reads, and a snapshot is harmless) but under 3.14/forkserver it will stall the compression loop and spike RSS — likely fatal on an RPi. It is also live on Windows `spawn` today.

One wrinkle worth flagging for whoever fixes it: the `start_time1/2 = -1` handshake is released at `Compression.py:363-368` **before** the extractor is started, so the buffer is already eligible for refill. The current by-value pickle hides that. A naive switch to shared memory would introduce a real read/write race — the extractor needs to either copy up front or hold the handshake until it is done.

## Non-blocking

### N1 — QueuedPool workers lose `InRmsFilter`

`QueuedPool.py:340` calls `initChildProcess(self.logging_queue, None)`. With `config=None`, `initChildLogging` attaches a `QueueHandler` with no `InRmsFilter` after removing the inherited (filtered) one, and forces root level to `DEBUG`.

This changes behavior under `fork` too — i.e. on every Linux station today. Detection workers previously inherited the parent's filtered handler; now every third-party `DEBUG` record (matplotlib, PIL, paramiko) gets pickled and pushed across the mp queue only to be discarded by the listener's own `InRmsFilter` (`Logger.py:516-521`). No correctness impact, but wasted IPC and CPU on the smallest boxes. Threading a config into the worker would restore parity with `Compressor`/`BufferedCapture`.

### N2 — NumPy 2 needs a Cython floor, and the `.pyx` comment is inverted

`requirements.txt` now requires `numpy>=2.3.0` on Python ≥ 3.13 but leaves `cython>=0.27.3`. NumPy 2 requires **Cython ≥ 3.0.6** to compile anything that `cimport numpy` — all 7 `.pyx` modules. A fresh pip install resolves to Cython 3.x and is fine, but an env with an older pinned Cython fails with an obscure C error. Suggest adding `cython>=3.0.6 ; python_version>='3.13'`.

Relatedly, the comment added to all 7 files — *"required on NumPy 2.0+ when built with Cython < 3"* — has it backwards: Cython < 3 cannot build against NumPy 2 at all, and Cython 3 emits `import_array()` itself. The calls are harmless, but the stated reason is wrong.

### N3 — Stale comments and docstrings

- `StartCapture.py:1034-1036` and `Reprocess.py:800-802` still say *"'forkserver' where available, else 'spawn'"*, left over from 1801c06 and contradicted by e5a9b08, which pins `fork` on Linux ≤ 3.13.
- `Compression.py:61-64` and `BufferedCapture.py:138-141` still document `array1` as *"numpy array in shared memory"*. The constructor contract changed to an `mp.Array` base. `RawFrameSave.py` was updated; these two were missed.

### N4 — Test isolation, and nothing runs the tests

`Tests/TestForkserverPickling.py` calls `multiprocessing.set_start_method('spawn', force=True)` inside three tests, mutating process-global state for the whole pytest session (and `test_setMultiprocessingStartMethod_returns_valid_method` leaves it on `forkserver` on Linux). `multiprocessing.get_context('spawn')` with the context passed explicitly would keep the tests hermetic.

More importantly, the repo has no CI workflow (`.github/` does not exist). The stated rationale is that `QueuedPool` "regressed on spawn platforms once before with zero detection" — a test nobody runs cannot prevent that recurrence. A minimal GitHub Actions job running this file on Linux + Windows would turn it from documentation into a guard.

### N5 — Remaining forkserver gaps

- `Utils/LiveViewer.py` (a `multiprocessing.Process`) got no `initChildProcess`. It is picklable so it will not crash, but it only `print()`s, and under `forkserver` a child does not inherit the parent's `sys.stdout = LoggerWriter` — so its output leaves the log. Its `__main__` is also not pinned.
- More generally, `initChildProcess` re-establishes handlers but not the `sys.stdout`/`sys.stderr` redirection that `initLogging` installs (`Logger.py:426-429`). Under `forkserver`, uncaught child tracebacks and bare `print()`s go to the console/journald rather than the RMS log — a real debuggability loss on 3.14. Worth doing in `initChildLogging`.
- `Utils/RecomputeCollectionAreas.py` `__main__` creates a `Pool` but is not pinned, unlike the six entry points listed in the description. `RMS/Routines/GstreamerCapture.py` `GstCaptureTest` likewise (dev-only).

### N6 — The SIGINT escalation is now dead code

`initChildProcess(..., ignore_sigint=True)` installs `SIG_IGN` in every child, but `Compressor.stop()` (`Compression.py:220`) escalates with `os.kill(self.pid, signal.SIGINT)` before falling back to `terminate()`. That kill is now a guaranteed no-op, so shutdown always pays the extra `join(5)`.

Net behavior is roughly unchanged — under `fork` the child inherited `StartCapture.breakHandler`, which only set a flag the compressor loop never reads — but the path is now provably dead. Either drop the SIGINT step or pass `ignore_sigint=False` in `Compressor.run()`.

### N7 — EventMonitor closes its inherited sqlite connection under fork (nit)

`EventMonitor.run()` reassigns `self.db_conn`, dropping the last child-side reference to the connection inherited across `fork`, so sqlite3 closes it in the child. SQLite documents that a child should neither use *nor close* a connection carried across a fork. Risk is very low here (no transaction is open at `start()`, and fcntl locks are per-process), but it is a new behavior on the `fork` path this PR otherwise leaves alone. Opening a child-owned connection is the right call; the implicit close of the inherited one is the part worth avoiding.

## Suggested order

1. Revert `Image.py:288` to `np.bool_` (B1).
2. Convert `Extractor` to the base + rebuilt-view pattern, handling the handshake race noted above (B2). Fixes Windows `spawn` today as well.
3. Add the Cython floor and reword the `import_array()` comments (N2).
4. Thread a config into the QueuedPool worker's `initChildProcess` (N1).
5. Comment/docstring cleanup (N3) and drop the dead SIGINT step (N6).
6. Make the tests context-local and add a CI job to run them (N4).
7. Optional: `LiveViewer`, child stdout/stderr redirection, `RecomputeCollectionAreas` pin (N5).

For B2 the meaningful validation is an end-to-end capture under an explicitly forced `forkserver` context, checking non-blank FFs, non-zero star counts, an FR file actually produced by the extractor, and steady RSS across several FF blocks — the synthetic `mp.Array` test in `TestForkserverPickling.py` does not reach that path.

Cybis320:
All five findings addressed in dcb21ec0..07f0613b:

**B1** (dcb21ec0): reverted to `np.bool_` with a comment explaining why the builtin must not return — verified the numba thresholding path executes cleanly.

**B2** (07f0613b): went with a third option that fixes both of your concerns at once — the compressor snapshots the block into a fresh lock-free `mp.Array` **before** releasing the capture handshake (one memcpy while the buffer is still guaranteed stable), and `Extractor.start()` now takes the base + shape and rebuilds its view in `run()`. That eliminates the ~236 MB pickle under forkserver/spawn (the base crosses the pipe as a handle — verified under both fork and spawn) *and* closes the pre-existing fork race you flagged in the wrinkle: the old code released the handshake seconds before the extractor ever started reading. Cost: one 236 MB memcpy per FF block, only with fireball detection enabled, held ~100 ms on a Pi-class box while double-buffering absorbs it. `Grouping3DRunner` is unaffected (it sets attributes directly and never calls `start()`).

**N1** (f978b98d): `QueuedPool` takes an optional `config` and the detection call site passes it — workers get their `InRmsFilter` back.

**N2** (5f91118b): `cython>=3.0.6 ; python_version>='3.13'` added alongside the numpy marker; the `import_array` comment now states the dependency the right way around in all 7 files.

**N3** (7950c7b8): both start-method comments and both constructor docstrings refreshed.

These are also merged to our station test branch, so the B2 snapshot path runs on ~20 fork-based stations with fireball detection on starting tonight.

PR #941, opened 2026-07-25
SkyFit2: fix clipped tab labels on collapse, and make the keyboard reference findable

Cybis320:
Two SkyFit2 UI fixes from user reports (Pete).

## Collapsed tab panel clipped the tab labels

Clicking the already-selected tab collapses the right-hand panel. The collapsed width was
`2 * averageCharWidth`, but with `TabPosition.East` the tab bar's width is driven by the tab
*height* (rotated labels) — font ascent/descent plus the style's tab padding. Those two numbers
are unrelated, so whether the bar fit was luck; where it didn't, Qt lays the bar out at a negative
x and the panel boundary clips its left edge, so the labels are partially cut rather than shrunk.
That's why it reproduced on some systems and not others. Levels is the tab selected at startup, so
it's the one people hit first, but any tab clicked twice did it.

`minimizedWidth()` now measures the bar instead of guessing: `tabBar().sizeHint().width()` plus the
pane frame, floored at the old value. Measured offscreen with the same tab set (Fusion, default font):

```
OLD: panel= 20 px  bar x=-16 w=36  visible=20/36  cut=16 px
NEW: panel= 38 px  bar x=  2 w=36  visible=36/36  cut=0 px
```

## The keyboard reference was too hard to find

The shortcut list moved out of the on-image F1 overlay into the Help tab, but the only way to reach
it was Shift+F1 → Help home → scroll past ~15 links to the Reference section, last of five in a
32-char-wide panel. Meanwhile F1 — the key people actually press for shortcuts — shows the platepar
info panel, which contains no shortcuts and no pointer to them. Four changes:

- **`Ctrl+/` and a Help menu entry.** New `openKeyboardReference()` opens the keyboard reference for
  the current mode. It lives in the Help menu, the one menu not rebuilt on mode change.
- **The F1 panel points at it.** `updateLeftLabels` appends `SHIFT+F1 Help   CTRL+/ Shortcuts`
  (CMD on macOS), also shown in the "No data loaded" empty state. F1's behaviour is unchanged.
- **Pinned on the Help home**, directly under the intro and above the section list; it stays in the
  Reference section too.
- **A "Keys" button** in the Help nav row next to Home/Back, so the reference is one click from any
  topic page.

Both keyboard-reference topics also gained `Ctrl+/` and `Shift+F1` rows at the top of their
Navigation tables. A shared `shortcutsTopicId(gui)` helper in `SkyFitHelp.py` does the mode → topic
mapping for all call sites.

## Testing

Help content verified programmatically: renders in both modes, the pinned link precedes the first
section header, both shortcut topics render and contain `Ctrl+/` and `Ctrl+S`, search for
"keyboard" still finds the topic. `Ctrl+/` parses to a valid sequence with no collision against
`Ctrl+S`; the three nav buttons fit without eliding at the Help tab's 2x width; tab bar geometry
measured as above.

Not click-tested in a running SkyFit2 — the module doesn't import in my environment
(`RMS.Astrometry.CyFunctions`, numpy ABI mismatch, unrelated to this change). Worth a quick manual
pass on `Ctrl+/`: the app-level `eventFilter` intercepts every Ctrl+key and forwards it to
`keyPressEvent`, but Qt's shortcut map runs before application event filters, which is why the
existing `Ctrl+S`/`Ctrl+N` menu actions work — `Ctrl+/` takes the same path.


dvida:
Reviewed the diff against `prerelease` and traced each claim in the description. **Both fixes look correct — no blocking defects.** One history-hygiene item and three optional polish items below.

### Claims I checked against the code

| Claim | Result |
|---|---|
| `Ctrl+/` survives the app-level `eventFilter` | ✅ The filter acts on `QEvent.KeyPress` (`SkyFit2.py:9652`), which Qt only delivers *after* `ShortcutOverride`/`Shortcut` dispatch. The existing `Ctrl+N` carries the same note at `SkyFit2.py:2635`. |
| Help menu is not rebuilt on mode change | ✅ `changeMode` clears only `file_menu`/`view_menu` (`SkyFit2.py:3502-3506`, `3546-3550`); `help_menu` is untouched in both branches. |
| Help tab exists in both modes | ✅ Added unconditionally at `CustomPyqtgraphClasses.py:1334`; `removeTabText` is only ever called for `Debruijn`/`Fit Parameters`/`Station`, so `_raiseHelpTab` always succeeds. |
| Pinned home link doesn't disturb search | ✅ `buildHelpHome` returns early on the `query` path (`SkyFitHelp.py:1216-1232`); `pinned` is inserted after that return. |
| Panel state handled when opening from the menu | ✅ `openKeyboardReference` → `openHelpTopic` → `_raiseHelpTab` selects the tab, syncs `tab.index`, sets `maximized`, calls `applyTabWidth`. |
| East `tabBar().sizeHint().width()` is stable | ✅ For East/West shapes the bar's horizontal extent is the per-tab *height* (font + padding), not label length — so the mode switches that add/remove tabs don't change the measurement. |
| The measurement is valid where it's taken | ✅ `applyTabWidth` is reached only from `onTabBarClicked` and `_raiseHelpTab`, both after `show()` (`SkyFit2.py:3429`), so the font is resolved. |
| The appended F1 hint fits | ✅ `label1` is monospace 8pt with `setTextWidth(averageCharWidth*35)` (`SkyFit2.py:2787`); the new line is 32 chars (31 with `CMD`), so no wrap. |
| `ctrl_label` exists at both new sites | ✅ Set unconditionally at `SkyFit2.py:3081`, before line 3448 and before any `updateLeftLabels()` call. |

---

### 1. Please squash-merge — a 13k-line build artifact is in the branch history

Commit f4b0cea also added `RMS/ExtractStarsCy.c` (+13,081 lines), and 613d8d4 removes it. The **net diff is clean** (which is why only 3 files show up), but the blob is still reachable from the branch, so a merge commit would pull it into `prerelease` history permanently. Squash-merging avoids that.

Two details in 613d8d4's message don't hold on this base, if the branch gets rewritten anyway:

- "The generated C file is gitignored" — not here. `.gitignore:99-107` is an explicit per-file list with no `RMS/ExtractStarsCy.c` entry, and `git check-ignore --no-index RMS/ExtractStarsCy.c` exits 1. That ignore line only exists on `speed-up-star-extraction`, which is presumably where the artifact came from.
- The Python-3.7 build failure can't happen on `prerelease`: there's no `RMS/ExtractStarsCy.pyx` and no `ExtractStarsCy` `Extension` in `setup.py` here (`grep -r ExtractStarsCy` → 0 hits). The stale-`.c` hazard is real, but it belongs to the star-extraction branch.

Separate follow-up, not for this PR: because the ignore list is per-file, every new `.pyx` needs a manual `.gitignore` edit or its generated `.c` shows up untracked — exactly this accident. A `*.c`/`*.cpp` rule with a `!Native/Hough/*` exception (the only tracked C/C++ sources) would close it.

### 2. The maximized width has the same bug this PR just fixed

`applyTabWidth`'s other branch still uses a pure character guess:

```python
self.setFixedWidth(self.scaledWidth(self.maximizedWidthChars()))   # CustomPyqtgraphClasses.py:1360
```

`setFixedWidth` sizes the bar and the page together, so the *pane* ends up with `32*averageCharWidth − barWidth` — roughly 36 px short of the intended 32 characters, and ~72 px short of 64 for the Help tab. Same root cause as the collapsed case, and the measurement is now available. Suggest factoring it out:

```python
def _barWidth(self):
    frame = 2*self.style().pixelMetric(QtWidgets.QStyle.PM_DefaultFrameWidth, None, self)
    return self.tabBar().sizeHint().width() + frame

def minimizedWidth(self):
    return max(self._barWidth(), self.scaledWidth(self.TAB_MINIMIZED_CHARS))

def applyTabWidth(self):
    if self.maximized:
        self.setFixedWidth(self.scaledWidth(self.maximizedWidthChars()) + self._barWidth())
    else:
        self.setFixedWidth(self.minimizedWidth())
```

`__init__` sets the initial width directly at line 1326 rather than calling `applyTabWidth()`, so it needs the same treatment — or could just call `applyTabWidth()`. Your call whether to widen the scope here; it's a small addition on top of what's already in the PR.

### 3. `Ctrl+/` doesn't exist on some common keyboard layouts

On German (`Shift+7`), French (`Shift+:`) and similar layouts `/` is a shifted key, so `Ctrl+/` never matches — the shortcut is dead there while the F1 panel and Help pages still advertise it. Cheap mitigation:

```python
self.keyboard_shortcuts_action.setShortcuts([QtGui.QKeySequence('Ctrl+/'),
                                             QtGui.QKeySequence('Ctrl+?')])
```

`setShortcuts` keeps the first sequence as the one rendered in the menu, so the displayed label stays `Ctrl+/`. Low priority given most stations are US/EN, but it's a one-liner.

### 4. Nit: redundant `getattr`

`getattr(self, 'ctrl_label', 'CTRL')` at both new `SkyFit2.py` sites — `ctrl_label` is set unconditionally at `SkyFit2.py:3081`, earlier in the same `setupUI` and before both use sites. Plain `self.ctrl_label` reads better and would fail loudly if that assignment ever moves. (The `getattr` in `SkyFitHelp._ctrl` is a different case and correctly defensive, since it takes an arbitrary `gui`.)

---

### Checked and fine, for the record

- `QtWidgets.QStyle.PM_DefaultFrameWidth` is an unscoped enum and so PyQt6-incompatible, but this file already requires PyQt (`QtCore.pyqtSignal` at line 1301) and uses unscoped enums throughout (`QTabWidget.East`, `Qt.Key_Escape`) — no new portability constraint.
- The `TAB_MINIMIZED_CHARS` floor (~14 px) can never bind against the measured ~38 px. Dead in practice, but the new comment says as much, so keeping it is harmless.
- `Ctrl+/` with no data loaded is safe: it bypasses `keyPressEvent`'s `hasData()` gate (it's a QAction), and the shortcuts topics only read GUI state through guarded helpers, with `buildHelpTopic` wrapped in try/except.
- `Ctrl+/` from the Help search box still fires — `QLineEdit` doesn't accept `ShortcutOverride` for it.
- No key collision: there's no `Key_Slash`/`Key_Question` handling anywhere in `SkyFit2.py`.

### Suggested manual pass

Since this wasn't click-tested, one run would be worth it:

1. Click the selected **Levels** tab to collapse — the vertical labels should be fully visible, not cut on the left. Click again to expand.
2. `Ctrl+/` → Help opens on "Keyboard reference" at double width. `F1` → the info panel's last line reads `SHIFT+F1 Help   CTRL+/ Shortcuts`.
3. Switch to Manual Reduction, `Ctrl+/` → the *MR* reference (exercises the persistent Help menu and the `shortcutsTopicId` mapping).
4. Help home → pinned link sits under the intro; search "keyboard" still finds the topic; **Keys** works from a topic page.
5. Launch with no data → the empty-state label shows the hint, and `Ctrl+/` still works.

Step 1 on Windows if convenient, since that's where the clipping was reported.

**Approving** — please squash-merge. Items 2-4 are optional; 2 is the only one with user-visible effect and is a natural companion to what's already here.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

dvida:
Pushed fixes for items 2-4 straight to the branch (de1dac1, 865cc48) so this is ready to go. Item 1 is a merge-time action, still outstanding.

### 2. Maximized width — de1dac1

Split the bar measurement out as `barWidth()` and added it to the maximized branch as well. `__init__` no longer sets the width itself; it calls `applyTabWidth()` after the tabs exist and the bar is on the East side, which is the earliest point where the measurement means anything.

Measured offscreen, Fusion, default font, same tab set:

```
barWidth              = 38 px

collapsed panel w=38  bar x=2 w=36  cut=0        (unchanged by this commit)

OLD maximized (32 ch) = 320  -> page = 282
NEW maximized (32 ch) = 358  -> page = 320
OLD maximized Help    = 640  -> page = 602
NEW maximized Help    = 678  -> page = 640
```

So the page now actually gets the 32 (or 64) characters the constants ask for. The panel itself is 38 px wider than before in the expanded state — worth an eyeball on a small screen, though it only claims back the space the page was already supposed to have.

### 3. `Ctrl+?` alias — 865cc48

```python
self.keyboard_shortcuts_action.setShortcuts([QtGui.QKeySequence('Ctrl+/'),
                                             QtGui.QKeySequence('Ctrl+?')])
```

Verified both register and that the menu still renders the first one:

```
shortcuts   = ['Ctrl+/', 'Ctrl+?']
menu label  = Ctrl+/
```

### 4. `getattr` — 865cc48

Both sites now read `self.ctrl_label` directly.

### Checks run

Offscreen (`QT_QPA_PLATFORM=offscreen`), on top of the pushed branch:

- `Utils.SkyFit2` and `RMS.Routines.CustomPyqtgraphClasses` import; geometry numbers above measured on a `QTabWidget` with the real tab set and `TabPosition.East`.
- Help content renders in both modes; the pinned "Keyboard reference" link precedes the first section header (offset 276 vs 411 in SkyFit, 248 vs 379 in MR); both shortcut topics contain `CTRL + /`, `SHIFT + F1` and `CTRL + S`; searching "keyboard" still finds the topic and the pinned link correctly stays out of the search view; `CMD + /` renders when `ctrl_label` is `CMD`.
- `shortcutsTopicId` maps `skyfit` → `shortcuts_skyfit` and `manualreduction` → `shortcuts_mr`.

Still no click test in a live SkyFit2 — the manual pass from my previous comment is worth a couple of minutes, particularly step 1 on Windows.

### Item 1 is still open

f4b0cea's `RMS/ExtractStarsCy.c` blob (+13,081 lines) remains reachable from this branch. Nothing in a commit can fix that — **please squash-merge**, otherwise it lands in `prerelease` history permanently.

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

dvida:
Fixed in 2b3cb75 — and it turns out the black bar predates this PR, the width change just made it more obvious.

pyqtgraph's `HistogramLUTItem.__init__` does `self.vb.setMaximumWidth(152)`, so the item never grows past roughly 219 px: 35 px axis + 152 px plot + 30 px gradient bar. Everything to the right of that is empty `GraphicsView` background. Measured offscreen at the Levels tab's page width:

```
                 item w   plot w   gradient ends at   dead band
before this PR    219      152          218              64 px    (page was 282)
after de1dac1     219      152          218             102 px    (page now 320)
after 2b3cb75     320      253          319               0 px
```

The fix lifts the cap in our `HistogramLUTItem` subclass, so the plot uses whatever width the tab is given:

```python
self.vb.setMaximumWidth(QtWidgets.QWIDGETSIZE_MAX)
```

Verified by grabbing the widget to a pixmap and finding the rightmost non-background pixel column: 218 with the cap, 319 without it, at a 320 px page. Also checked the item fills correctly at 219 / 282 / 320 / 640 px pages, that `setLevels`/`getLevels` still round-trip, and that the region's `(0, None)` lower bound survives.

Net effect for the Levels tab: the histogram plot is now 253 px wide instead of 152, which is a bit easier to read as well.

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

PR #942, opened 2026-07-25
Always delete detection backups after a standalone reprocess (fixes #418)

### LINKED ISSUE #418: Incomplete cleanup after running reprocessing job?
Platform: 8GB Pi5 running Bookworm 64-bit, flashed from RMS_RPi5Bookworm_image_20240910.img.xz, 11-Sep-2024 16:29, 3GB, our newest image as of today and soon to be linked in the wiki. The new image has OS and RMS updated, chrony installed, and the latest default .config file.

I was bringing up a test station last night, and forgot to copy in the mask and platepar file for that camera.

I realized my error in the evening during capture, so in the morning, after processing was completed I moved the ArchivedFiles directory and tar.bz2 out of Archived files, copied in the correct mask and platepar to ~/source/RMS and to the CapturedFiles directory for last night.

I then restarted capture, expecting that the system would detect the missing ArchivedFiles directory and reprocess the data.

It did not detect the missing ArchivedFiles directory and re-create it, so I ran a manual reprocessing job, which completed successfully, using the correct platepar and mask.

I then restarted capture so it would be ready to run tonight, and to my surprise, the system detected unprocessed data, and is now proceeding to duplicate my reprocessing job.

My assumption is that something did not get cleaned up at the end of my reprocessing job that fooled RMS into thinking it needed to reprocess. I believe this problem should be easy to reproduce, as I have a distant recollection of having the same thing happen several months ago.

Cybis320:
Fixes #418.

## The bug

In `RMS/Reprocess.py`, the detection-backup cleanup sits **inside** the `if config.upload_enabled:`
block, so a standalone `python -m RMS.Reprocess <dir>` on a station with uploading disabled never
deletes the `rms_queue_bkup_*.pickle` files it wrote into the captured directory.

Every other caller cleans up unconditionally — `StartCapture.py:862` (end of night),
`StartCapture.py:1004` (auto-reprocess), `DetectStarsAndMeteors.py:560`. The standalone script is
the odd one out. The nesting dates back to f89771c3 (Oct 2018), whose intent was to delay the delete
until after upload; it just landed one level too deep.

The leftovers then trip the auto-reprocess check on the next start (`StartCapture.py:951-977`):

```python
pickle_files = glob.glob("{:s}/rms_queue_bkup_*.pickle".format(captured_dir_path))
...
if any_pickle_files:
    run_reprocess = True
```

Any leftover pickle means "partially processed" regardless of the FTPdetectinfo being present, so
RMS reprocesses the night the user had just reprocessed by hand.

It is self-concealing: the duplicate run goes through `StartCapture`, which *does* clean up, so the
pickles are gone afterwards and the night never triggers again. That is why it reads as
irreproducible and why the same reporter hit it twice, two years apart.

## Confirmed from a station log

`log_USTEST_20260725_172751_001.log`, RMS master f441ef15:

```
17:27:51 DownloadMask - Can't contact the server: RSA private key file not found.
17:27:53 StartCapture-line:983 - Found partially-processed data in .../USTEST_20260725_024335_309988
17:28:17 QueuedPool-line:204  - Loaded 3157 backed up results...
...
17:59:06 StartCapture-line:1008 - Folder .../USTEST_20260725_024335_309988 reprocessed with success!
```

- No `UploadManager` line anywhere in the run, and `StartCapture.py:1195-1200` only constructs one
  when `config.upload_enabled` is set — so uploading is off on this station.
- 3157 backups loaded is one per FF file for the full 9.021 hr night, i.e. the manual run left its
  *entire* backup set, not a partial failure.
- The manual run had finished cleanly 36 minutes earlier (`reprocess_log_USTEST_20260725_152745_001.log`,
  15:27:45 -> last entry 16:51:07), so this is not a race with a still-running job.
- The duplicate cost ~31 minutes of the pre-capture window. Detection was cheap (results came from
  the backups) but the ML filter, calibration, thumbnails, both stacks and archiving all re-ran.

## The change

Move the cleanup out of the `upload_enabled` branch and into a `finally`, so it runs whether or not
uploading is enabled and whether or not it raised. Ordering is preserved: the delete still happens
after the upload manager is stopped, which was the point of f89771c3.

## Testing

Verified structurally that the call is no longer gated on `upload_enabled` and now sits in
`Try.finalbody`, and that removing the `rms_queue_bkup_*.pickle` files flips `StartCapture`'s
`run_reprocess` decision from True to False with the FTPdetectinfo left in place. Not run on a live
station.

### DIFF
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -853,29 +853,35 @@ if __name__ == "__main__":
         # Run the external script
         runExternalScript(cml_args.dir_path[0], night_archive_dir, config)
 
-    # Upload the archive, if upload is enabled
-    if config.upload_enabled:
+    try:
 
-        # Add metadata archive first, so it might get uploaded first
-        files_to_add_list_unfiltered = [metadata_archive_name, imgdata_archive_name, archive_name]
-        files_to_add_list = [f for f in files_to_add_list_unfiltered if f is not None]
+        # Upload the archive, if upload is enabled
+        if config.upload_enabled:
 
-        # Init the upload manager
-        log.info('Starting the upload manager...')
+            # Add metadata archive first, so it might get uploaded first
+            files_to_add_list_unfiltered = [metadata_archive_name, imgdata_archive_name, archive_name]
+            files_to_add_list = [f for f in files_to_add_list_unfiltered if f is not None]
 
-        upload_manager = UploadManager(config)
-        upload_manager.start()
+            # Init the upload manager
+            log.info('Starting the upload manager...')
 
-        # Add file for upload
-        log.info(f'Adding files to upload list: {files_to_add_list}')
-        upload_manager.addFiles(files_to_add_list)
+            upload_manager = UploadManager(config)
+            upload_manager.start()
 
-        # Stop the upload manager
-        if upload_manager.is_alive():
-            upload_manager.stop()
-            log.info('Closing upload manager...')
+            # Add file for upload
+            log.info(f'Adding files to upload list: {files_to_add_list}')
+            upload_manager.addFiles(files_to_add_list)
 
+            # Stop the upload manager
+            if upload_manager.is_alive():
+                upload_manager.stop()
+                log.info('Closing upload manager...')
 
-        # Delete detection backup files
+    finally:
+
+        # Delete detection backup files. This has to happen whether or not uploading is enabled and
+        #   whether or not it succeeded: any rms_queue_bkup_*.pickle left in the captured directory
+        #   makes StartCapture treat the night as partially processed and reprocess it all over
+        #   again on the next start.
         if detector is not None:
             detector.deleteBackupFiles()
--- a/RMS/Reprocess.py
+++ b/RMS/Reprocess.py
@@ -849,11 +849,15 @@ if __name__ == "__main__":
     night_archive_dir, archive_name, imgdata_archive_name, metadata_archive_name, detector =  \
                                                     processNight(cml_args.dir_path[0], config)
 
-    if cml_args.run_extl_script:
-        # Run the external script
-        runExternalScript(cml_args.dir_path[0], night_archive_dir, config)
+    upload_manager = None
 
+    # The night is fully processed by this point, so the steps below all have to be followed by the
+    #   backup cleanup in the finally block, whether they succeed or not
     try:
+
+        if cml_args.run_extl_script:
+            # Run the external script
+            runExternalScript(cml_args.dir_path[0], night_archive_dir, config)
 
         # Upload the archive, if upload is enabled
         if config.upload_enabled:
@@ -872,16 +876,24 @@ if __name__ == "__main__":
             log.info(f'Adding files to upload list: {files_to_add_list}')
             upload_manager.addFiles(files_to_add_list)
 
-            # Stop the upload manager
-            if upload_manager.is_alive():
-                upload_manager.stop()
-                log.info('Closing upload manager...')
-
     finally:
+
+        # Stop the upload manager. It is a non-daemon process which loops until told to exit, so
+        #   leaving it running on an error path would hang the script at interpreter shutdown
+        if (upload_manager is not None) and upload_manager.is_alive():
+            log.info('Closing upload manager...')
+            upload_manager.stop()
+
 
         # Delete detection backup files. This has to happen whether or not uploading is enabled and
         #   whether or not it succeeded: any rms_queue_bkup_*.pickle left in the captured directory
         #   makes StartCapture treat the night as partially processed and reprocess it all over
         #   again on the next start.
         if detector is not None:
-            detector.deleteBackupFiles()
+
+            # Never let a cleanup failure mask an exception which is already propagating
+            try:
+                detector.deleteBackupFiles()
+
+            except Exception as e:
+                log.error('Deleting the detection backup files failed with message:\n' + repr(e))


dvida (commented):
Reviewed the diff against the code and the history. **The diagnosis is correct and the fix is the right shape** — nice bit of archaeology on a two-year-old intermittent. Everything in the writeup checks out:

- `f89771c3` really did land the delete one level too deep. Its `Reprocess.py` hunk adds `if detector is not None: detector.deleteBackupFiles()` literally inside the `if config.upload_enabled:` block, so the nesting has been wrong since 2018.
- Reprocess's `__main__` really is the only leaking caller. Every other `QueuedPool` with a non-`None` `backup_dir` deletes unconditionally — `DetectStarsAndMeteors.py:560`, `ExtractStars.py:944`, `StartCapture.py:862`, `StartCapture.py:1004` — and `Utils/FluxBatch.py:1019` and `Utils/TrackStack.py:272` pass `backup_dir=None`.
- `StartCapture.py:955-977` short-circuits on `any_pickle_files` before it ever looks at the FTPdetectinfo, so a single leftover pickle forces a full reprocess. Since the standalone run writes `FTPdetectinfo_*.txt` into the captured dir, removing the pickles does flip the decision to False.
- `detector` is always bound where the new `finally` uses it: `processNight` has a single `return`, and `detector = None` on both the pre-detected and `nodetect` paths.

Two suggestions while this block is being touched, then some notes.

### 1. Add an `except` so an upload failure doesn't kill the process

Right now a failure anywhere in the upload block exits the script with a raw traceback and a nonzero status — even though the night is fully processed and archived by that point, and `addFiles()` has already persisted the queue to `FILES_TO_UPLOAD.inf`. For a standalone CLI, logging it and exiting cleanly is the better contract:

```python
    except Exception:
        log.error("An error occurred while uploading the archive!", exc_info=True)

    finally:
        ...
```

(`exc_info=True` matches the style already used in this file in the `processFramesFiles` timelapse handler.)

### 2. Consider pulling `runExternalScript` inside the `try`

```python
    try:

        if cml_args.run_extl_script:
            # Run the external script
            runExternalScript(cml_args.dir_path[0], night_archive_dir, config)

        # Upload the archive, if upload is enabled
        if config.upload_enabled:
            ...
```

To be clear, this is defensive rather than a live bug — I checked, and `runExternalScript` (`RunExternalScript.py:38-101`) wraps its whole body in `try/except Exception` and logs, so it cannot currently propagate and leak the backups. But it is the one thing left sitting between `processNight` and the new cleanup guarantee, and it costs nothing to cover. It also matches the sibling path: `StartCapture.py:1004` deletes the backups *before* calling `runExternalScript` at `:1008`, so the net ordering ends up the same.

### 3. Worth a sentence in the PR description

The move to `finally` does two things, and the description really only covers one. Besides running when uploading is off, the cleanup now also runs when the upload block *raises*.

That is fine — `UploadManager.addFiles()` saves the queue to `FILES_TO_UPLOAD.inf` (`UploadManager.py:635`), so upload retries survive a restart independently of the pickles. The only residual case is `UploadManager(config)` or `.start()` raising *before* `addFiles`, where the archive never gets queued and there is no longer a leftover pickle to accidentally force an auto-reprocess. That accident was a bug rather than a design, and the `StartCapture` paths never had it either, so I am not asking for a change — just worth stating so the behavior change is on the record.

### 4. Notes, non-blocking

- The 23/17 diff is a two-line semantic change plus mechanical re-indentation. Might be worth saying so in the description so reviewers don't go hunting for hidden edits.
- Blank lines immediately after `try:` and after `finally:` aren't the style used elsewhere in `Reprocess.py`.
- The four-line comment could come down to two — the rationale is genuinely non-obvious and worth keeping, just more tersely.
- Only the Jul 2026 recurrence has a log confirming `upload_enabled` was off; the Sep 2024 original doesn't, so the two reports may not be strictly the same cause. Doesn't change anything about the fix.

### Unrelated, spotted while reading

`StartCapture.py:841-842` discards `processNight`'s detector into `_`, and `:862` then deletes via the capture-time detector created at `:561`. Two different objects sharing the same `backup_dir`, so it works — but `:561` is inside a conditional, so the binding at `:862` might be worth a look for the `nodetect` and frames-only paths. Nothing to do with this PR.

### Testing

Structural verification seems reasonable here, but it would be good to confirm on a station before merge:

```bash
# with upload_enabled: false
python -m RMS.Reprocess ~/RMS_data/CapturedFiles/XXNNNN_YYYYMMDD_...
ls ~/RMS_data/CapturedFiles/XXNNNN_.../rms_queue_bkup_*.pickle   # should be gone
ls ~/RMS_data/CapturedFiles/XXNNNN_.../FTPdetectinfo_*.txt       # should be present
```

then start capture and check that the log does not print `Found partially-processed data in` for that directory.

dvida:
Reviewed this and pushed three fixes on top as 770b3cc5. The diagnosis is right — I reproduced it end to end below — and all three findings are in the `try`/`finally` this PR adds, not in the reasoning behind it.

Fixes #418.

### Reproduced on a real reprocess run

Ran `python -m RMS.Reprocess <dir> -c .` against a station directory (`AU0001`, real `.config`, platepar, mask, FF file) with `upload_enabled: false`, on three versions of `Reprocess.py`:

```
                          exit  pickles_left  FTPdetectinfo
before this PR              0        1             1          <- issue #418
this PR (15cccf0c)          0        0             1
with 770b3cc5               0        0             1
```

One `rms_queue_bkup_*.pickle` survives on the old code, sitting next to a complete FTPdetectinfo — precisely the state `StartCapture.py:974` turns into a full reprocess. Invoking reprocess a second time on those same directories gives the signature from the report:

```
before this PR:  Loaded 1 backed up results...
with 770b3cc5:   Loaded 0 backed up results...
```

Also confirmed `QueuedPool.deleteBackupFiles()` removes only `rms_queue_bkup_*` and leaves the FF and FTPdetectinfo files alone.

### Finding 1 (hardening): the external script call sits outside the `try`

`runExternalScript` is called before the `try` is entered, so the cleanup does not cover it. In practice nothing escapes it — `RunExternalScript.py:101` catches `Exception` around everything it does — so this is **not** an actively reachable bug, and I do not want to oversell it. What it does mean is that the guarantee depends on a callee's blanket `except` rather than on the structure here, and a reasonable future tightening of that `except` would silently reintroduce #418. `KeyboardInterrupt` also escapes it today, and the night is fully processed by that point. Moving the call inside the `try` costs one line.

### Finding 2 (real, and it hangs the script)

`upload_manager` is started inside the `try` but only stopped on the happy path. It is a non-daemon `multiprocessing.Process` (`UploadManager.py:530`) whose `run()` loops `while not self.exit.is_set()` (`:865`). If `addFiles()` raises, the exception leaves the child looping, and interpreter shutdown joins non-daemon children — so the standalone reprocess hangs instead of exiting. Standalone reproduction of just that model:

```
without a finally stop:  exit=124   (timed out, hung at interpreter exit)
with a finally stop:     exit=0
```

Pre-existing, but this PR adds the exact construct that fixes it. `stop()` moves into the `finally`, which is why the `is_alive()` guard moves with it. `upload_manager` needs a `None` initialiser first — `stop()` is safe on an already-exited child, but `join()` asserts on one that was never started, which is every run with uploading disabled.

### Finding 3: `deleteBackupFiles()` can mask the real exception

It is a bare `os.remove()` loop (`QueuedPool.py:301-310`). Raising from inside a `finally` replaces whatever exception was already propagating, so one `PermissionError` during cleanup would hide the actual failure. Now caught and logged.

### Test harness

Some of these paths cannot be reached from a real run — you cannot easily make `addFiles` fail. So there is also a harness that slices the real `__main__` tail out of `Reprocess.py` by source text, wraps it in a function body and `exec`s it against stubs for `config`, `cml_args`, `detector`, `log`, `UploadManager` and `runExternalScript`. It runs the actual source, so it cannot drift from the file it tests. Scores:

| version | result |
| --- | --- |
| before this PR | 5 failed, 11 passed (includes the #418 case) |
| this PR (15cccf0c) | 4 failed, 12 passed (findings 1, 2, 3) |
| with 770b3cc5 | 16 passed |

The four failures on the current head are: external script raises -> backups not deleted; `addFiles` raises -> manager not stopped; cleanup raises -> swallowed-and-logged expected but it propagates; cleanup raises during an in-flight error -> the `OSError` replaces the original `RuntimeError`.

Harness in a `<details>` block below if you want it in the repo — I left it out of the commit since `Tests/` is Python scripts rather than a runner, and I did not want to invent a convention inside a one-file fix.

### Still needs a live station

Nothing here exercises the actual #418 loop, because that runs through `StartCapture`'s startup path: manual reprocess of a real captured night with uploading off, then a capture restart that does *not* re-run the night. Everything above stops at "no pickles survive the reprocess", which is the precondition, not the whole chain.

### Out of scope, worth a separate issue

`StartCapture.py:969-978` treats *any* leftover pickle as "partially processed" and checks that before the FTPdetectinfo, so a crash, a kill, or a future caller reproduces #418 from a different direction. And `StartCapture.py:1014` swallows a reprocess failure, so a deterministic failure retries on every start — roughly what #756 was aiming at.

<details>
<summary>Test harness</summary>

```python
""" Out-of-line harness for the post-processNight tail of RMS/Reprocess.py's __main__ block.

The block cannot be imported (it lives under `if __name__ == "__main__":`), so the real source is
sliced out of the file, wrapped in a function body and exec'd against stubs. This runs the actual
source rather than a copy of it, so it cannot silently drift from the file it is testing.

Covers the paths a real reprocess run cannot easily be made to take: an upload that raises, an
external script that raises, and a cleanup that raises.
"""

from __future__ import print_function

import os
import sys
import textwrap


SRC = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'reprocess_src.py')
if not os.path.isfile(SRC):
    SRC = '/home/dvida/source/RMS/RMS/Reprocess.py'

# Slice the tail: from the upload_manager init (or the try, on the unfixed version) to EOF
with open(SRC) as f:
    src = f.read()

markers = ['    upload_manager = None', '    if cml_args.run_extl_script:']
start = -1
for marker in markers:
    start = src.find(marker)
    if start != -1:
        break

if start == -1:
    sys.exit('FAIL: could not locate the post-processNight block in {:s}'.format(SRC))

block = src[start:]

# The block sits at one indent level inside `if __name__ == "__main__":`, so it drops straight into
#   a function body as-is
tail_src = 'def _tail():\n' + block + '\n'
tail_code = compile(tail_src, SRC + ' (__main__ tail)', 'exec')


class FakeDetector(object):
    def __init__(self, raises=False):
        self.raises = raises
        self.deleted = False

    def deleteBackupFiles(self):
        self.deleted = True
        if self.raises:
            raise OSError('simulated os.remove failure')


class FakeUploadManager(object):
    """ Mimics the bits of UploadManager the block touches, including the assertion a real
        multiprocessing.Process raises when join() is called before start().
    """

    instances = []

    def __init__(self, config, add_raises=False):
        self.started = False
        self.stopped = False
        self.alive = False
        self.files = None
        self.add_raises = add_raises
        FakeUploadManager.instances.append(self)

    def start(self):
        self.started = True
        self.alive = True

    def addFiles(self, files):
        self.files = files
        if self.add_raises:
            raise RuntimeError('simulated upload queue failure')

    def is_alive(self):
        return self.alive

    def stop(self, timeout=None):
        if not self.started:
            raise AssertionError('can only join a started process')
        self.stopped = True
        self.alive = False


class FakeConfig(object):
    def __init__(self, upload_enabled):
        self.upload_enabled = upload_enabled


class FakeArgs(object):
    def __init__(self, run_extl_script=False):
        self.dir_path = ['/fake/CapturedFiles/XX0001_20260101_000000_000000']
        self.run_extl_script = run_extl_script


class FakeLog(object):
    def __init__(self):
        self.errors = []

    def info(self, *args, **kwargs):
        pass

    def debug(self, *args, **kwargs):
        pass

    def error(self, msg, *args, **kwargs):
        self.errors.append(str(msg))

    def warning(self, *args, **kwargs):
        pass


def runTail(upload_enabled=False, run_extl_script=False, extl_raises=False, add_raises=False,
            detector=None, detector_raises=False):
    """ Exec the real block with stubs. Returns (namespace, raised_exception_or_None). """

    FakeUploadManager.instances = []

    if detector is None:
        detector = FakeDetector(raises=detector_raises)

    def runExternalScript(dir_path, night_archive_dir, config):
        ns['extl_called'] = True
        if extl_raises:
            raise RuntimeError('simulated external script failure')

    ns = {
        'config': FakeConfig(upload_enabled),
        'cml_args': FakeArgs(run_extl_script=run_extl_script),
        'detector': detector,
        'log': FakeLog(),
        'UploadManager': lambda config: FakeUploadManager(config, add_raises=add_raises),
        'runExternalScript': runExternalScript,
        'night_archive_dir': '/fake/ArchivedFiles/XX0001_20260101_000000_000000',
        'archive_name': '/fake/a_detected.tar.bz2',
        'imgdata_archive_name': '/fake/a_imgdata.tar.bz2',
        'metadata_archive_name': '/fake/a_metadata.tar.bz2',
        'extl_called': False,
    }

    exec(tail_code, ns)

    raised = None
    try:
        ns['_tail']()

    except BaseException as e:
        raised = e

    ns['detector_obj'] = detector
    ns['manager'] = FakeUploadManager.instances[0] if FakeUploadManager.instances else None

    return ns, raised


results = []


def check(name, cond, detail=''):
    results.append((name, bool(cond), detail))
    print('{:s} {:<62s} {:s}'.format('ok  ' if cond else 'FAIL', name, detail))


print('source: {:s}'.format(SRC))
print()

# --- the reported bug -------------------------------------------------------------------------
ns, raised = runTail(upload_enabled=False)
check('upload disabled -> backups deleted (issue #418)', ns['detector_obj'].deleted)
check('upload disabled -> no exception', raised is None, repr(raised) if raised else '')
check('upload disabled -> manager never constructed', ns['manager'] is None)

# --- happy path with uploading on -------------------------------------------------------------
ns, raised = runTail(upload_enabled=True)
check('upload enabled -> backups deleted', ns['detector_obj'].deleted)
check('upload enabled -> files queued', ns['manager'].files == ['/fake/a_metadata.tar.bz2',
                                                               '/fake/a_imgdata.tar.bz2',
                                                               '/fake/a_detected.tar.bz2'],
      str(ns['manager'].files))
check('upload enabled -> manager stopped', ns['manager'].stopped)
check('upload enabled -> no exception', raised is None, repr(raised) if raised else '')

# --- finding 2: addFiles raises ---------------------------------------------------------------
ns, raised = runTail(upload_enabled=True, add_raises=True)
check('addFiles raises -> backups still deleted', ns['detector_obj'].deleted)
check('addFiles raises -> manager stopped (no hang at exit)', ns['manager'].stopped)
check('addFiles raises -> original exception propagates', isinstance(raised, RuntimeError),
      repr(raised))

# --- finding 1: external script raises --------------------------------------------------------
ns, raised = runTail(upload_enabled=False, run_extl_script=True, extl_raises=True)
check('external script raises -> backups still deleted', ns['detector_obj'].deleted)
check('external script raises -> original exception propagates', isinstance(raised, RuntimeError),
      repr(raised))

ns, raised = runTail(upload_enabled=True, run_extl_script=True)
check('external script runs before the upload', ns['extl_called'] and ns['manager'].files is not None)

# --- finding 3: cleanup raises ----------------------------------------------------------------
ns, raised = runTail(upload_enabled=False, detector_raises=True)
check('cleanup raises with no in-flight error -> swallowed and logged',
      raised is None and len(ns['log'].errors) == 1,
      ns['log'].errors[0][:60] if ns['log'].errors else '')

ns, raised = runTail(upload_enabled=True, add_raises=True, detector_raises=True)
check('cleanup raises during an error -> original error survives',
      isinstance(raised, RuntimeError), repr(raised))

# --- degenerate case: nodetect / previously detected night gives detector = None ---------------
# runTail() substitutes a stub when detector is None, so exec directly for this one
FakeUploadManager.instances = []
ns = {
    'config': FakeConfig(False),
    'cml_args': FakeArgs(),
    'detector': None,
    'log': FakeLog(),
    'UploadManager': FakeUploadManager,
    'runExternalScript': lambda *a: None,
    'night_archive_dir': '', 'archive_name': '', 'imgdata_archive_name': '',
    'metadata_archive_name': '',
}
exec(tail_code, ns)
raised = None
try:
    ns['_tail']()
except BaseException as e:
    raised = e
check('detector is None -> no crash', raised is None, repr(raised) if raised else '')

print()
failed = [n for n, ok, _ in results if not ok]
if failed:
    print('{:d} FAILED, {:d} passed'.format(len(failed), len(results) - len(failed)))
    sys.exit(1)

print('all {:d} passed'.format(len(results)))
```

</details>

PR #948, opened 2026-07-28
Universal station tooling: multi-camera conversion in FirstRun, unified StartCapture

Cybis320:
This PR packages the multi-camera data structure work proposed by **Peter E.** on the forum, submitted on his behalf. It is the "proper pull request" promised in the June revert (c91ba9e4), updated with the improvements from Pete's revised proposal (his `Upgrade.zip`).

## What this does

**`RMS_FirstRun.sh`**
- Automatically checks whether the root file system fills the storage device. If not, one keypress runs `raspi-config --expand-rootfs` and reboots — no more navigating raspi-config menus. Warns if the device itself is below the 64 GB minimum.
- Prompts for station ID, latitude, longitude and elevation directly at the terminal, with validation, instead of opening the whole config file in an editor.
- At the end of setup, offers conversion to the multi-camera data structure (with a confirmation on decline), then chains to `add_Station.sh`.
- Assorted cleanups: never overwrites an existing SSH key (GMN has the public half on file), internet check falls back to HTTPS when ping is blocked, file-system expansion step is skipped on non-Pi systems.

**`Scripts/add_Station.sh`** (new, universal — replaces the Pi/Linux split)
- Creates stations under `~/source/Stations/<ID>`, interactively or in bulk from a CSV (`stations_template.csv` sample included).
- Converts a legacy single-camera install: `.config`, `mask.bmp` and `platepar_cmn2010.cal` are **copied** to the station directory and existing data is moved under `~/RMS_data/<ID>`. Nothing in `~/source/RMS` is modified — the conversion never deletes, rewrites or re-downloads files in the RMS root.
- The Pi camera-count limits (1, or 4 on a Pi5) are advisory with an override rather than hard exits.
- `add_Pi_Station.sh` and `add_GStation.sh` remain as thin compatibility wrappers, answering the naming concern Pete raised: the tool now lives at `Scripts/add_Station.sh`. Renaming the `MultiCamLinux/` directory itself is left for later so deployed entry points keep working.

**`Scripts/RMS_StartCapture.sh`**
- Now works on both data structures, as Pete's revised `RMS_StartCapture2.sh` proposed: if station directories exist under `~/source/Stations` it starts each one in its own terminal (staggered 70 s on multi-camera systems so concurrent first-start updates cannot collide, 2 s when there is only one station); otherwise it starts the legacy single camera as before.
- `RMS_StartCapture_MCP.sh` becomes a compatibility wrapper, and `add_Station.sh` no longer needs to retarget the Desktop shortcut after a conversion.

## Testing

- Pete tested his revised scripts end-to-end on a freshly flashed Bullseye image on a Pi4 (terminal transcripts posted on the forum).
- The reworked universal scripts were exercised in a sandboxed environment (isolated `$HOME` with shims) covering the legacy-to-multicam conversion, interactive station creation and the CSV bulk flow.
- A full interactive run of this exact branch on a freshly flashed Pi image is still outstanding — **testers with a spare SD card are very welcome.**

## Credits

- **Peter E.** — original conversion proposal, the revised scripts this PR incorporates, and Pi testing.
- **Ed Harman** — the original MultiCamLinux tooling this builds on.


markmac99:
I'll review this as soon as i can but not likely to be before 20th Aug as we're on holiday from 11th and i have a bunch of other GMN work to finish this week. 
In principle  as described all seems good stuff.

Scripts/RMS_StartCapture.sh
```diff
@@ -1,26 +1,74 @@
...
+    fi
+
+    echo " Starting all configured stations..."
+
+    loop=0
+    for Dir in ~/source/Stations/*/
+      do
+        Station=$(basename "$Dir")
+        echo " Starting camera ${Station}"
+        if [[ "$TERMINAL" == "gnome-terminal" ]]; then
```
markmac99, line 40: Earlier, we defined  TERMINAL, so shouldn't we do $TERMINAL --profile... here?  I realise the two terminals have different cmdline options. 
Maybe 
 [ $TERMINAL == gnome ]  && OPTS="--profile --title -- bash -c" || OPTS-"--title -e"
might work (pseudocode but you get the idea)

Also there's an assumption that the only WM options are Gnome or LXDE. This might potentially break with other windowmanagers.

Scripts/RMS_StartCapture.sh
```diff
@@ -1,26 +1,74 @@
...
+    # Detect the terminal emulator
+    if [[ "${XDG_CURRENT_DESKTOP:-}" == *GNOME* ]] && command -v gnome-terminal >/dev/null 2>&1; then
+        TERMINAL=gnome-terminal
+    elif command -v lxterminal >/dev/null 2>&1; then
+        TERMINAL=lxterminal
+    else
+        TERMINAL=gnome-terminal
+    fi
+
+    echo " Starting all configured stations..."
```
markmac99, line 33: I like the principle of starting all cameras from one script, though i remain of the view that using interactive windows is unnecessary (cron is easier...)!

Scripts/RMS_StartCapture.sh
```diff
@@ -1,26 +1,74 @@
...
+    echo " All cameras started"
+
+else
+
+    # Legacy single-camera data structure
+    source ~/vRMS/bin/activate
+    cd ~/source/RMS
 
-mkdir -p $LOGPATH
+    mkdir -p ~/RMS_data/logs/
```
markmac99, line 60: hardcoded log location - is this actually used?

Scripts/RMS_StartCapture.sh
```diff
@@ -1,26 +1,74 @@
...
+dircount=$(find ~/source/Stations -mindepth 1 -maxdepth 1 -type d 2>/dev/null | wc -l)
+
+if [[ $dircount -gt 0 ]]; then
+
+    # On a single camera system concurrent first-start updates are no
+    # risk, so the long stagger delay is skipped
+    if [[ $dircount -le 1 ]]; then
+        seconds=2
+    else
+        seconds=70
```
markmac99, line 21: should we use the capture_wait_seconds value from the CSV file here?  (if available)

Scripts/MultiCamLinux/Pi/RMS_StartCapture_MCP.sh
```diff
@@ -1,19 +1,6 @@
...
-	lxterminal --title=${Station} -e "$HOME/source/RMS/Scripts/MultiCamLinux/StartCapture.sh ${Station}"  &
-	echo "  waiting $seconds seconds..."
-	sleep ${seconds}
-	if [[ $loop = 0 ]] ; then
-	    seconds=10
-	fi
-	let loop++
-  done
-echo " All cameras started"
+# Kept for compatibility with Desktop shortcuts created by older
```
markmac99, line 3: I would recommend retiring these scripts in a couple of releases, to avoid potential confusion and accidental changes being introduced. Good idea to proxy them out for now tho.

markmac99 (commented):
generally looks good, Main points are noted inline.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+if [[ ! -d ~/source/Stations && -n "$DefStation" && "$DefStation" != "XX0001" ]]; then
+    migrate_legacy
+elif [[ -f ~/.rmsautorunflag && "$DefStation" == "XX0001" && ! -d ~/source/Stations ]]; then
+    echo "Please run RMS_FirstRun and configure your 1st station,"
+    echo "then run add_Station again to convert it or add more cameras."
+    exit
+fi
+
+# Ensure required directories exist
+mkdir -p ~/source/Stations
```
dvida, line 281: **Blocking (1/12): declining the conversion makes it permanently unreachable.**

This `mkdir -p ~/source/Stations` runs unconditionally — including when the user just declined `migrate_legacy` above and added no stations. But both entry points gate on the directory merely *existing*:

* `RMS_FirstRun.sh:386` — `if [[ ! -d ~/source/Stations ]]` → step 5 is never offered again
* `add_Station.sh:272` — `if [[ ! -d ~/source/Stations && ... ]]` → `migrate_legacy` is never reached again

So FirstRun's own advice at line 427 ("If you decide to convert in the future, run this command from a terminal") is wrong: after a single decline, the legacy camera can never be converted by either script. Same trap for anyone who runs `add_Station.sh` once and backs out at the prompt.

Suggested fix: create `~/source/Stations` lazily inside `create_station`, and gate on `$(count_stations) -eq 0` instead of on directory existence — the `count_stations` helper already exists at line 51.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+
+    echo ""
+    echo "Conversion of station ${DefStation} complete"
+}
+
+# ------------------------------------------------------------------ main
+
+# Parse arguments: [RMS_data_path] [stations.csv]
+# Smart detection: if single arg ends in .csv, treat as CSV file
+RMS_data=~/RMS_data
```
dvida, line 204: **Blocking (2/12): the migration hardcodes `~/RMS_data` and ignores the configured `data_dir`.**

`RMS_FirstRun.sh:419` invokes this script with no arguments, so `RMS_data` stays at this default. On a legacy install whose `.config` has e.g. `data_dir: /mnt/ssd/RMS_data` (common on PC stations, and on Pis with external storage):

* the banner at lines 161-164 promises to move data from the wrong path;
* `find "${RMS_data}" ... -exec mv -t ...` at line 180 moves nothing;
* `create_station` still rewrites the station config to `data_dir: <this default>/<ID>` (line 137).

Net effect: the existing archive is orphaned and capture silently switches to the system drive, which then fills up. This is a silent data-and-space hazard on exactly the systems least able to absorb it.

Suggested fix: default `RMS_data` from the root config — `awk '/^data_dir:/ {print $2}' ~/source/RMS/.config`, with `~` expanded — falling back to `~/RMS_data` only when that is absent.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+    # scripts - their different file names would otherwise double-launch
+    # the station at login
+    rm -f ~/.config/autostart/"${station}_StartCap.desktop" \
+          "${Desktop}/${station}_StartCap.desktop" \
+          "${Desktop}/Show_LiveStream-${station}.desktop" \
+          "${Desktop}/${station}-Show_LiveStream.desktop"
+
+    if [[ $total -gt 1 ]]; then
+        # scale the reserved free space with the number of stations
+        sed -i "s/^extra_space_gb:.*$/extra_space_gb: $(( total * 20 ))/g" "${dir}/.config"
```
dvida, line 451: **Blocking (3/12): every run clobbers operator-tuned `.config` values.**

This loop rewrites `extra_space_gb` and `reboot_after_processing` across *all* station configs on *every* invocation. Two separate problems:

1. Line 456 force-writes `reboot_after_processing: true` for single-station systems. The old scripts only ever set it to `false`. Anyone who deliberately disabled the nightly reboot gets it silently re-enabled the next time they add a camera.
2. `extra_space_gb` becomes `total * 20`, down from `add_Pi_Station.sh`'s `total * 30`. On a Pi5 with 4 cameras that is 80 GB reserved instead of 120 GB — a real change in free-space headroom that deserves to be a deliberate decision rather than a side effect of merging the two scripts.

Suggested fix: write these only at station-creation time, or put the recompute behind an explicit flag; and reconcile the 20-vs-30 factor on purpose.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+            header="${HEADERS[$i]}"
+            value="${VALUES[$i]}"
+
+            # Skip station_id, camera_ip (already handled), and empty values
+            [[ "$header" == "station_id" || "$header" == "stationID" || "$header" == "camera_ip" ]] && continue
+            [[ -z "$value" ]] && continue
+
+            # Check if this header exists as a key in .config
+            if grep -q "^${header}:" "${config_file}"; then
+                sed -i "s|^${header}:.*$|${header}: ${value}|g" "${config_file}"
```
dvida, line 369: **Blocking (4/12): CSV values containing `&` corrupt the config.**

An unescaped `&` in a `sed` replacement expands to the whole match. The `device:` line is the obvious victim — the stock value in `.config:101` is `rtsp://.../user=admin&password=&channel=1&stream=0.sdp`, and `device` is a perfectly reasonable column for a bulk deploy. Demonstrated:

```
in:  device: rtsp://192.168.42.10:554/user=admin&password=&channel=1
out: device: rtsp://1.2.3.4:554/user=admindevice: rtsp://192.168.42.10:554/...password=device: ...channel=1
```

`|` and `\` in a value break it too, and `${header}` is interpolated into the regex side unescaped.

Suggested fix: escape the replacement before substituting —

```bash
value=${value//\\/\\\\}; value=${value//&/\\&}; value=${value//|/\\|}
```

or rewrite the line with `awk` instead of `sed`.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+    TERMINAL=gnome-terminal
+else
+    TERMINAL=lxterminal
+    echo "Warning: neither lxterminal nor gnome-terminal was found, launch entries will use lxterminal"
+fi
+
+# Get user's desktop directory
+Desktop=$(xdg-user-dir DESKTOP 2>/dev/null || echo "$HOME/Desktop")
+
+DefStation=$(uppercase "$(awk '/^stationID:/ {print $2}' ~/source/RMS/.config 2>/dev/null)")
```
dvida, line 269: **(5/12) `DefStation` is the one station ID that is never validated.**

This is read straight from the root `.config` and then passed unquoted to `create_station` (`mkdir -p ~/source/Stations/${DefStation}`) and to `find ... ! -name "${DefStation}"` at line 180.

The CSV path (line 363) and the interactive path (line 396) both validate against `^[A-Za-z0-9_-]+$`; the migration path does not. A hand-edited `stationID` containing a space or a slash will corrupt the migration — wrong directories created, and the `find`/`mv` exclusion silently failing to match.

Suggested fix: apply the same regex check here and bail with a clear message if it fails.

Scripts/RMS_FirstRun.sh
```diff
@@ -208,50 +299,148 @@ echo "Updating to the latest version of RMS..."
...
+    flushInput
+    read -r -p "Station ID (e.g. US01AB): " statID || exit 1
+    statID=$(trimSpaces "$statID" | tr '[:lower:]' '[:upper:]')
+    if [[ "$statID" =~ ^[A-Z]{2}[A-Z0-9]{4}$ && "$statID" != "XX0001" ]]; then
+      break
+    fi
+    echo "  A station code has 2 country letters followed by 4 characters, e.g. US01AB."
+    # Only offer to keep a nonstandard entry when it is still safe to use
+    # as a directory name and a .config value
+    if [[ "$statID" =~ ^[A-Z0-9_-]+$ ]]; then
```
dvida, line 326: **(6/12) the override lets the user accept `XX0001`, creating a dead end.**

Line 320 correctly rejects `XX0001` on the happy path, but this "use it anyway" branch matches `^[A-Z0-9_-]+$`, which `XX0001` satisfies. If the user presses Y, step 5 chains to `add_Station.sh`, which hits its line 274 and prints *"Please run RMS_FirstRun and configure your 1st station"* — sending the user back to the script they are currently inside.

Suggested fix: exclude `XX0001` from this branch too, e.g. `[[ "$statID" =~ ^[A-Z0-9_-]+$ && "$statID" != "XX0001" ]]`.

Scripts/RMS_FirstRun.sh
```diff
@@ -208,50 +299,148 @@ echo "Updating to the latest version of RMS..."
...
+echo "Enter the station ID and the geo coordinates of your camera."
 echo ""
 
 echo "
-If you need to make changes to the configuration file in the future, you can
-find a shortcut to it on desktop (RMS_config), or open it directly in
-$RMSCONFIG.
+If you need to change any other setting in the future, you can find a
+shortcut to the configuration file on desktop (RMS_config), or open it
+directly in $RMSCONFIG.
```
dvida, line 310: **(7/12) after conversion, this text points users at a config file that is no longer live.**

Step 4 tells the user to edit `$RMSCONFIG` (`~/source/RMS/.config`) or use the desktop `RMS_config` shortcut. Once step 5 converts, neither is true: capture reads `~/source/Stations/<ID>/.config`, and `migrate_legacy` deletes `~/Desktop/RMS_config.txt` (`add_Station.sh:192`).

The replacement — the `Edit_configs` helper created by `add_Station.sh` — is never mentioned to the user.

Suggested fix: print a short closing note after a successful conversion saying where the live config now lives and which shortcut edits it.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+}
+
+# Build the Exec= line for a .desktop entry using the detected terminal.
+# The gnome-terminal StartCapture profile is optional - create one to
+# customise the capture windows; without it the default profile is used.
+term_exec() {   # $1 = window title, $2 = command
+    if [[ "$TERMINAL" == "gnome-terminal" ]]; then
+        printf 'gnome-terminal --profile=StartCapture --title=%s -- bash -c "%s"' "$1" "$2"
+    else
+        printf 'lxterminal --title=%s -e "%s"' "$1" "$2"
```
dvida, line 66: **(8/12) `.desktop` `Exec=` lines use `~`, which the desktop-entry spec does not expand.**

`term_exec` emits `~/source/RMS/Scripts/MultiCamLinux/StartCapture.sh ...` into the `Exec=` line. The gnome-terminal branch survives by accident because `bash -c` re-expands the tilde; this lxterminal branch passes the literal string to `lxterminal -e`, which `execvp`s it verbatim and fails.

This is pre-existing in both old scripts, but since this PR rewrites the generator it is a good moment to fix it: use `${HOME}`, exactly as the runtime call in `RMS_StartCapture.sh` already does.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+Exec=${HOME}/source/RMS/Scripts/CMNbinViewer_env.sh
+Hidden=false
+NoDisplay=false
+Icon=${HOME}/source/RMS/Scripts/MultiCamLinux/icon.png
+EOF
+gio set "${Desktop}/CMNbinViewer.desktop" metadata::trusted true 2>/dev/null
+chmod +x "${Desktop}/CMNbinViewer.desktop"
+
+# Create Desktop shortcut for editing the station .config files
+cat << 'EOF' > "${Desktop}/Edit_configs"
```
dvida, line 504: **(9/12) `Edit_configs` will not launch for most users.**

This writes a bare executable shell script to the Desktop — no `.desktop` entry, no `Terminal=true`. Double-clicking it on GNOME or LXDE will not open a terminal, so its `read` prompts have nowhere to go. Every other shortcut this script creates is a proper `.desktop` entry.

Two smaller issues inside it:

* `editor=""` if none of the five candidates is installed, so `"$editor" ...` becomes "command not found";
* `for Dir in ~/source/Stations/*/` has no nullglob guard, so on an empty `Stations` directory it tries to edit a literal `*`.

Suggested fix: ship it as a `.desktop` entry that runs through `term_exec`, like the others.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+    ssh-keygen -t rsa -f ~/.ssh/id_rsa -q -P ""
+    cp ~/.ssh/id_rsa.pub "${Desktop}/"
+    echo "SSH keys generated in ~/.ssh"
+    echo "Your new id_rsa.pub public key file is now placed on the Desktop."
+    echo "Be sure to send a copy of this file to Denis"
+fi
+
+# Add RMS activation to .bashrc if not present
+if ! grep -q "source ~/vRMS/bin/activate" ~/.bashrc; then
+    cat <<EOF >> ~/.bashrc
```
dvida, line 482: **(10/12) `~/.bashrc` is now modified on the Pi too.**

`add_Pi_Station.sh` never touched `.bashrc`, and the `add_GStation.sh` version of this block was indented inside its heredoc (so the appended lines were inert). This version is un-indented and actually works — which makes it effectively new behavior on every platform: every new interactive shell will now `cd ~/source/RMS`.

Worth a deliberate decision rather than inheriting it from the merge. At minimum, dropping the `cd` and keeping only the venv activation would be far less surprising.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+
+# Make sure the Desktop StartCapture entry points at RMS_StartCapture.sh,
+# which handles both the legacy and the multi-camera data structure (the
+# entry may be a stale copy rather than a symlink on older installs).
+# On the Pi images FirstRun also runs this launcher at boot.
+ln -sf ~/source/RMS/Scripts/RMS_StartCapture.sh "${Desktop}/RMS_StartCapture.sh"
+
+# Generate SSH keys if not present
+if [[ ! -f ~/.ssh/id_rsa ]]; then
+    ssh-keygen -t rsa -f ~/.ssh/id_rsa -q -P ""
```
dvida, line 473: **(11/12) SSH key format differs between the two paths.**

`RMS_FirstRun.sh` generates with `ssh-keygen -t rsa -m PEM`; this call omits `-m PEM` and produces an OpenSSH-format key. Which format a machine ends up with depends on which script happened to run first.

Suggested fix: pick one and use it in both places.

Scripts/add_Station.sh
```diff
@@ -0,0 +1,539 @@
...
+    printf '%s' "$s"
+}
+
+uppercase() {
+    printf '%s' "$*" | tr '[:lower:]' '[:upper:]'
+}
+
+count_stations() {
+    find ~/source/Stations -mindepth 1 -maxdepth 1 -type d 2>/dev/null | wc -l
+}
```
dvida, line 57: **(12/12) `lxterminal` is no longer installed by the station tooling.**

The old `add_GStation.sh` installed it and tuned its scrollback/geometry. `RMS_Installer.sh:38` still installs it, so the `GRMSL_Install.sh` path stays covered — but a system that has neither terminal now gets this warning plus launch entries that reference a binary that is not there.

Worth either installing the chosen terminal here, or making this warning louder about the consequence (the entries created in this run will not work).

dvida (commented):
Reviewed statically, with targeted experiments for the assumptions that could fail silently. No Pi hardware, so the expand-rootfs path is unverified — see the testing note at the end.

Overall: the consolidation is the right direction and the copy-only conversion is a genuine improvement over the previous `rm .config && wget` behavior. Four findings look blocking to me, eight are worth fixing before merge, and the rest are nits. All are posted inline.

## Verified correct

These were the highest-risk assumptions in the diff, so I checked them rather than trusting them:

* **The `^`-anchored seds match.** `.config` keys sit at column 0 (lines 15, 18, 21, 26, 90, 266, 288), so moving from `s/D:.*$/.../` to `s/^stationID:.*$/.../` works. This was the single biggest silent-failure risk in the PR and it is correct.
* **`camera_settings.json` really is shared from the RMS root.** `camera_settings_path: ./camera_settings.json` resolves against the CWD (`RMS/ConfigReader.py:1129`), and `MultiCamLinux/StartCapture.sh` does `cd "$HOME/source/RMS"` before launching. The copy-only claim in the description holds.
* **`gnome-terminal --profile=StartCapture` degrades gracefully** when the profile does not exist — tested: it warns "Attempting to fall back to the default profile" and exits 0.
* **The CSV trailing-newline idiom does not loop forever.** `while IFS=, read -r -a VALUES || [[ ${#VALUES[@]} -gt 0 ]]` terminates correctly; `read -a` clears the array at EOF (tested).
* **Both wrapper relative paths resolve.** `Pi/add_Pi_Station.sh` → `../../add_Station.sh` and `MultiCamLinux/add_GStation.sh` → `../add_Station.sh` both land on `Scripts/add_Station.sh`, and `GRMSL_Install.sh:46` calls the wrapper from its own directory, so that flow still works.
* **The GRMSUpdater regex change targets the right site.** The other two station patterns (`GRMSUpdater.sh:106` and `:318`) interpolate the ID literally; only `:530` hardcoded `{6}`.
* **No double-launch.** Pi images get a Desktop entry only (FirstRun starts captures at boot), everything else gets XDG autostart, and the recompute loop removes the old `_StartCap.desktop` / `Show_LiveStream-*` names.
* **This converges the Pi and PC Desktop shortcut.** `GenerateDesktopLinks.sh` already symlinks `~/Desktop/RMS_StartCapture.sh` to `Scripts/RMS_StartCapture.sh`, so making that script dual-mode is the right call and removes the old retarget hack.
* All three new/changed scripts pass `bash -n`.

## Nits (not worth inline comments)

* `.gitignore` — the new comment names `Scripts/MultiCamLinux/add_GStation.sh`, but the reader is now `Scripts/add_Station.sh` and the file it looks for is `~/source/RMS/stations.csv`.
* `Scripts/RMS_Installer.sh:21` still points at `~/source/RMS/Scripts/MultiCamlinux/add_GStation.sh` (also misspelled). This PR touches the file, so it is an easy one to sweep up.
* `stations_template.csv` uses `XX0001`–`XX0006` as sample IDs, but `XX0001` is the sentinel "unconfigured" value both scripts gate on. Non-sentinel placeholders would be less confusing. Also worth considering `Scripts/` rather than the repo root for the template.
* Pi model parsing (`add_Station.sh:244`) takes the first integer in the model string, so "Raspberry Pi 400" reads as 400 and is treated as Pi5-class (4 cameras).
* `RECOMMENDED -gt 0` doubles as an "is this a Pi" test at `add_Station.sh:99`; the comment there already acknowledges the conflation. A separate `IS_PI` flag would survive future edits to the limit logic.
* The multi-camera branch of `RMS_StartCapture.sh` silently drops `"$@"`, and `RMS_StartCapture_MCP.sh` forwards arguments that then go nowhere.
* The overview at `RMS_FirstRun.sh:129` still lists "0. Expand the file system" on non-Pi systems, where the entire block is skipped.
* `flushInput()` is defined in `RMS_FirstRun.sh` but two call sites still inline `while read -t 0.01; do :; done`.
* `Desktop=$(xdg-user-dir DESKTOP 2>/dev/null || echo "$HOME/Desktop")` (`add_Station.sh:267`) leaves `Desktop=""` if `xdg-user-dir` succeeds with empty output, after which every path becomes `/...`. A non-empty check would close that.
* The wrapper scripts use `$(dirname "$0")`, which breaks if they are ever invoked through a symlink. Nothing symlinks them today, but `GenerateDesktopLinks.sh` already has the `readlink -f` idiom worth copying.
* `camera_ip` splicing (`add_Station.sh:354`) only matches an IPv4 literal in the URL, so a config already using a hostname is silently left alone.

## Suggested testing before this lands on `prerelease`

The description notes a full interactive run on a fresh Pi image is still outstanding, and that is the main remaining risk. What I would want to see:

1. **A sandboxed matrix run** — isolated `$HOME` with `RMS_PI_MODEL_FILE` shimmed (the script already supports that override at line 241), covering: fresh Pi plus convert; fresh Pi, decline, then re-run (finding 1); a legacy install with a non-default `data_dir` (finding 2); a CSV containing a `device` column (finding 4); and re-running `add_Station.sh` on an already-converted system (finding 3).
2. **A fresh Pi image, full interactive run**, including the expand-rootfs reboot path — nothing static can validate `raspi-config --expand-rootfs`, the `resize2fs` wait branch, or the re-entry into FirstRun after the reboot.
3. **One generic Ubuntu run** through `GRMSL_Install.sh` into the `add_GStation.sh` wrapper, confirming the gnome-terminal autostart entries actually start capture at login.
4. `shellcheck -S warning` over the three new/changed scripts.

PR #951, opened 2026-07-29
Honor fixed_scale in the radial fitAstrometry paths

GlassOnTin:
## Summary
`fitAstrometry` accepts a `fixed_scale` argument (SkyFit2 exposes it as the "Fixed scale" checkbox), but for **radial** distortion types the scale is always re-fit: the radial joint fit (matched-pair LM and NN/RANSAC paths) carries `F_scale` as a free parameter and overwrites `self.F_scale`, silently ignoring `fixed_scale`. `fixed_scale=True` and `False` produce identical results on radial platepars before this change.

## Fix
Hold `F_scale` at its input value when `fixed_scale` is set, at every site where the radial fit otherwise takes it from the fitted parameter vector: the five radial residual functions, the two final extractions, and the scoring / RANSAC warm-start temporaries. `F_scale` is *held* rather than *dropped* from the parameter vector to keep the diff small across the multi-stage RANSAC; it then has no effect and the optimizer leaves it at its seed. (Happy to switch to true removal from the vector if preferred.)

## Testing
Synthetic recovery on the bundled `AU000A` example (radial9-odd), perturbing RA +0.5°, Dec +0.3°, F_scale ×1.05:

| | F_scale (truth 24.087, perturbed 25.29) | reproj RMS |
|---|---|---|
| `fixed_scale=False` | **24.087** (recovered) | **0.000 px** (no regression) |
| `fixed_scale=True` | **25.29** (held, as documented) | 1.94 px |

## Note on the earlier (master) version
The first version of this PR also normalized `F_scale` in the Nelder-Mead simplex to address an F_scale runaway on a fast fisheye plate. Per the review discussion that runaway is a genuinely degenerate scale-vs-distortion direction (not a conditioning issue), and `fixed_scale=True` is the actual cure — so the normalization is dropped, and prerelease's LM matched-pair fit makes it moot anyway. A self-contained reproducer of the divergence is linked in the comments.

### DIFF
--- a/RMS/Formats/Platepar.py
+++ b/RMS/Formats/Platepar.py
@@ -923,7 +923,7 @@ class Platepar(object):
             pp_copy.RA_d = (360 * ra_ref) % (360)
             pp_copy.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
             pp_copy.pos_angle_ref = (360 * pos_angle_ref) % (360)
-            pp_copy.F_scale = abs(F_scale)
+            pp_copy.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
 
             # Assign distortion parameters
             pp_copy.x_poly_rev = params[4:]
@@ -947,7 +947,7 @@ class Platepar(object):
             pp_copy.RA_d = (360 * ra_ref) % (360)
             pp_copy.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
             pp_copy.pos_angle_ref = (360 * pos_angle_ref) % (360)
-            pp_copy.F_scale = abs(F_scale)
+            pp_copy.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
             pp_copy.x_poly_rev = np.array(params[4:])
 
             img_x, img_y, _ = img_stars.T
@@ -973,7 +973,7 @@ class Platepar(object):
             pp_copy.RA_d = (360 * ra_ref) % (360)
             pp_copy.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
             pp_copy.pos_angle_ref = (360 * pos_angle_ref) % (360)
-            pp_copy.F_scale = abs(F_scale)
+            pp_copy.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
 
             # Assign distortion parameters
             pp_copy.x_poly_fwd = np.array(params[4:])
@@ -1009,7 +1009,7 @@ class Platepar(object):
             pp_copy.RA_d = (360 * ra_ref) % (360)
             pp_copy.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
             pp_copy.pos_angle_ref = (360 * pos_angle_ref) % (360)
-            pp_copy.F_scale = abs(F_scale)
+            pp_copy.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
             pp_copy.x_poly_fwd = np.array(params[4:])
 
             img_x, img_y, _ = img_stars.T
@@ -1048,7 +1048,7 @@ class Platepar(object):
             pp_copy.RA_d = (360 * ra_ref) % (360)
             pp_copy.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
             pp_copy.pos_angle_ref = (360 * pos_angle_ref) % (360)
-            pp_copy.F_scale = abs(F_scale)
+            pp_copy.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
 
             # Assign distortion parameters
             pp_copy.x_poly_fwd = np.array(params[4:])
@@ -1187,6 +1187,11 @@ class Platepar(object):
 
                 # Fitting the pointing direction below! - if used, it should be put BEFORE the reverse fit!
                 # Initial parameters for the pointing and distortion fit (normalize to the 0-1 range)
+                # Hold F_scale when fixed_scale is set. The radial joint fit below otherwise carries F_scale
+                # as a free parameter; the patched residual functions make the objective independent of it,
+                # and every extraction/temporary below reads this held value instead of the (drifting under
+                # Nelder-Mead) parameter. fixed_scale is forwarded through the recursive matched-pair fit too.
+                f_scale_fixed = abs(self.F_scale)
                 p0 = [self.RA_d / 360, self.dec_d / 90, self.pos_angle_ref / 360, abs(self.F_scale)]
                 p0 += self.x_poly_fwd.tolist()
 
@@ -1315,7 +1320,7 @@ class Platepar(object):
                                 self.RA_d = (360 * ra_ref) % (360)
                                 self.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
                                 self.pos_angle_ref = (360 * pos_angle_ref) % (360)
-                                self.F_scale = abs(F_scale)
+                                self.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
                                 # Update x_poly_fwd with best radial3-odd coefficients
                                 self.x_poly_fwd = np.array(best_res.x[4:])
                                 self.x_poly_rev = self.x_poly_fwd.copy()
@@ -1347,7 +1352,7 @@ class Platepar(object):
                                 self.RA_d = (360 * ra_ref) % (360)
                                 self.dec_d = -90 + (90 * dec_ref + 90) % (180.000001)
                                 self.pos_angle_ref = (360 * pos_angle_ref) % (360)
-                                self.F_scale = abs(F_scale)
+                                self.F_scale = f_scale_fixed if fixed_scale else abs(F_scale)
                                 # Update x_poly_fwd with best radial5-odd coefficients
                                 self.x_poly_fwd = np.array(best_res.x[4:])
                                 self.x_poly_rev = self.x_poly_fwd.copy()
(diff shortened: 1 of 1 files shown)


Cybis320:
Thanks for digging into this — and for the writeup. I ran your synthetic-recovery test on the bundled AU000A radial9-odd platepar and your numbers reproduce almost exactly (free fit recovers F_scale to 24.089 with ~0.07 px RMS; fixed_scale=True holds 25.29). The fixed_scale half of this is a real bug and worth fixing: on master, the SkyFit2 "Fixed scale" checkbox is silently ignored for radial distortion types unless "fit only pointing" is also enabled — I confirmed fixed_scale=True and False produce bit-identical results before your patch.

Two asks:

1. Could you retarget this to prerelease? All development goes through prerelease (master only receives tested releases), so a PR against master can't be merged as-is. Note it won't be a clean rebase: the radial fit path was substantially rewritten on prerelease — the matched-pair joint fit now uses Levenberg-Marquardt least_squares (_lstsqFit) instead of Nelder-Mead, so the divergence concern is mostly moot there. What does carry over is the fixed_scale bug: prerelease's radial path (both the LM matched-pair fit and the NN/RANSAC path) still never honors it. A port that drops F_scale from the parameter vector in those spots would be very welcome.

2. Can you share how to reproduce the runaway? I couldn't trigger it even with sparse stars (n=15) and a large perturbation (F_scale ×1.2, +2° rotation, +1° pointing) — master converges fine there. I'm also a bit skeptical of the conditioning explanation: scipy's Nelder-Mead builds its initial simplex from 5% relative perturbations and the algorithm is affine-equivariant, so rescaling one coordinate should give an essentially identical search trajectory — the only real differences are the absolute xatol termination (which raw F_scale effectively can't satisfy, so the fit always runs to maxiter) and floating-point noise. My guess is your fisheye plate has a genuinely degenerate scale-vs-distortion direction, and which side of the cliff the fit lands on is chance — in which case fixed_scale=True (your other fix) is the actual cure. If you can share the platepar + star pairs from the 3.2 mm plate that diverged, I'd like to check whether the normalization reliably prevents it or whether it just happened to on that run.

GlassOnTin:
Thanks @Cybis320 — really helpful review, and you're right on the key point.

**On the conditioning/normalization explanation — you've convinced me, and I've dropped it.** My own testing agrees with your affine-equivariance argument: on the 3.2 mm fisheye plate, normalizing F_scale *alone* only moved the runaway from ~3e6 to ~1300 px/deg — it still diverged. `fixed_scale=True` is what actually held it. So this PR is now **purely the `fixed_scale` fix**, and since prerelease replaced Nelder-Mead with LM `least_squares` on the matched-pair path, the divergence angle is moot there anyway.

**Retargeted to `prerelease` and ported the fix.** For radial types the joint fit still carried F_scale as a free parameter and overwrote `self.F_scale`, ignoring `fixed_scale`, on both the matched-pair and NN/RANSAC paths. I hold F_scale at its input value at every site where the fit takes it from the parameter vector (the 5 radial residual functions + the 2 extractions + the scoring / RANSAC warm-start temporaries). I *held* it rather than *dropped* it from the vector to keep the diff small through the multi-stage RANSAC — the parameter then has no effect and the optimizer leaves it at its seed. Happy to switch to true removal if you'd prefer.

**Validated on the bundled AU000A (radial9-odd)** — synthetic recovery (RA +0.5°, Dec +0.3°, F_scale ×1.05):
- `fixed_scale=False` → F_scale recovered 25.29 → 24.087 (truth 24.087), **0.000 px** reprojection RMS (no regression)
- `fixed_scale=True` → F_scale **held at 25.29**, as documented

**Divergence reproducer** (you asked for the platepar + star pairs): https://gist.github.com/GlassOnTin/77ebb1efc3f8b5a3874a1d76b049f50d — the 3.2 mm plate + 88 time-frozen star pairs + a `repro.py`. On master it drives F_scale 15.82 → **3,093,444** px/deg with `fixed_scale=False`. It's a genuinely degenerate scale-vs-distortion direction on this cheap fisheye, exactly as you suspected — so `fixed_scale=True` is the real cure. Moot on prerelease's LM path, but there in case it's useful.

Cybis320 (changes requested):
Thanks for the retarget and the reproducer — the gist reproduces on my end too, and I agree with your read: it's a genuinely degenerate scale-vs-distortion direction, so `fixed_scale=True` is the right cure and dropping the normalization was the right call.

I verified the ported fix. The **matched-pair LM path is correct**: with the patched residuals making the objective exactly independent of the scale parameter, the finite-difference Jacobian column in `least_squares` is exactly zero, so LM leaves `x[3]` at its seed and the (unpatched) extractions at lines 1694/1750 read back the held value. Your AU000A numbers reproduce for me on this path.

However, the **NN/RANSAC path is still incomplete** — the hold-in-vector approach breaks down under Nelder-Mead, and two sites were missed:

1. **Line 1605** ("Apply best RANSAC params"): `self.F_scale = abs(F_scale)` still comes from `best_res.x`. Unlike LM, Nelder-Mead does *not* leave a flat coordinate at its seed — simplex reflections/contractions move all coordinates, so `x[3]` random-walks through the RANSAC iterations and a drifted, meaningless scale lands in the platepar. This is the same reason the extractions at 1322/1354 and the scoring temporaries at 1433/1480 needed patching; this final extraction is the one that actually persists.
2. **Line 1661**: the final recursive matched-pair fit (`self.fitAstrometry(..., use_nn_cost=False)`) doesn't forward `fixed_scale`, so even if RANSAC held the scale, the final fit re-fits it freely. Pre-existing omission, but it defeats the NN-path half of the fix end-to-end.

(Cosmetic: the `pp_iter` debug-callback copy at line 1548 will also display the drifted value.)

Since both misses are symptoms of the same fragility — relying on the optimizer to leave a flat parameter at its seed — I'll take you up on your offer: please **drop F_scale from the parameter vector when `fixed_scale` is set**, rather than holding it. That eliminates the Nelder-Mead drift problem outright and removes the implicit LM-seed assumption on the matched-pair path. If the conditional vector length proves too invasive across the multi-stage RANSAC (all the `res.x[4:]` slicing and warm-starts), completing the hold is an acceptable fallback — patch line 1605 the same way as the others and pass `fixed_scale=fixed_scale` (and arguably `fit_only_pointing`) through the recursive call at 1661 — but removal is the shape I'd prefer to merge.

One mitigating note for context: SkyFit2 currently never passes `fixed_scale` into the `use_nn_cost=True` calls, so the user-facing checkbox bug is fixed by this PR as-is. But the API claims to honor `fixed_scale` on the NN path and mostly patches it, so let's close the gap rather than ship it half-held.

GlassOnTin:
Thanks @Cybis320 — both misses fixed, and tested on the NN path this time.

**Changes:**
- The final *"apply best RANSAC params"* extraction now reads the held `f_scale_fixed`, so the drifted `best_res.x[3]` no longer persists into `self.F_scale`.
- `fixed_scale` is forwarded through the recursive final matched-pair fit (`use_nn_cost=False`).
- Also swept the scoring / warm-start / iteration-callback temporaries to read the held value, for consistency.

I went with the **completed hold** (your accepted fallback) rather than the removal. The removal threads a conditional 3-vs-4 vector length through ~15 sites — the 5 residual functions, ~6 extractions, the warm-start seeding, and the recursive-loop `p0` rebuild — which is exactly the multi-stage-RANSAC invasiveness you flagged. Since the patched residuals already make the objective *exactly* independent of `x[3]`, the only requirement for correctness is that no extraction reads `x[3]` and that `fixed_scale` flows through the recursion — both now hold, so the Nelder-Mead drift has no effect on the persisted result.

**Verified on AU000A (radial9-odd), both cost paths** (synthetic recovery, RA +0.5°, Dec +0.3°, F_scale ×1.05):

| | `fixed_scale=False` | `fixed_scale=True` |
|---|---|---|
| `use_nn_cost=False` (LM) | recovers → **0.000 px** | held |
| `use_nn_cost=True` (NN/RANSAC) | recovers → **0.001 px** | **held, no drift** |

Happy to switch to the full vector removal if you would still prefer that shape for merge — just wanted to avoid the conditional-length threading unless you want it.

dvida:
Pushed a small follow-up commit (7f355ed) to finish honoring `fixed_scale` on the last radial path.

**In plain terms:** this PR pins `F_scale` at every place the radial fit reads it back — except two spots on the *matched-pair* path (`use_nn_cost=False`): the fwd-rev loop and the final joint refinement. Those still wrote `self.F_scale` straight from the fitted vector.

Today they happen to hold the right value: the patched residual makes the objective independent of the `F_scale` parameter, so `least_squares` (LM) leaves that coordinate sitting at its seed. So there's no bug right now.

But that's exactly the path the **"Fixed scale" checkbox** actually runs through — `fitPickedStars` calls `fitAstrometry` with `use_nn_cost=False`. So the main user-facing case was correct only by relying on optimizer null-space behavior, not by an explicit rule like everywhere else in the PR. If someone later reintroduces an `F_scale` dependence in the residual, or swaps the optimizer, the scale would silently start drifting on that path.

The commit just gates those two lines with the same `f_scale_fixed if fixed_scale else abs(...)` pattern you used everywhere else, so the hold is explicit.

**No behavior change** — verified on a radial7-odd synthetic recovery through the matched-pair path:
- `fixed_scale=True`  → `F_scale` held at the seed (does not re-fit)
- `fixed_scale=False` → `F_scale` recovers truth, reproj RMS 0.000 px

Also worth flagging (separate, pre-existing, **not** touched here): the NN auto-recalibration callers (`SkyFit2.py:12699` / `:13674`, `AutoPlatepar.py:600/693`) save/restore `user_fixed_scale` but never forward it into their `use_nn_cost=True` calls, so the checkbox has no effect on automatic recalibration. Likely intentional, but the dangling save/restore reads like unfinished wiring.

PR #953, opened 2026-07-29
SkyFit2: use scoped Qt enum names for PyQt proxy / Qt 5.15 compatibility

GlassOnTin:
## Summary
`SkyFit2` imports `QtWidgets`/`QtGui` via `pyqtgraph.Qt`, whose proxy no longer exposes the old flat/unscoped enum aliases on newer stacks (e.g. **Qt 5.15 on Debian Trixie**). Accesses like `QtWidgets.QMessageBox.Yes` then raise `AttributeError` and crash the GUI (first hit is the new-platepar "Pointing information" dialog, but there are more throughout).

## Fix
Sweep **all 40** flat `QtWidgets`/`QtGui` enum accesses in `Utils/SkyFit2.py` to their fully-scoped names (valid on both PyQt5 and PyQt6 and through the proxy):

- `QMessageBox.{Yes,No,Cancel,Save,Discard}` → `.StandardButton.*`, `.ActionRole` → `.ButtonRole.ActionRole`, `.Warning` → `.Icon.Warning`
- `QFileDialog.ShowDirsOnly` → `.Option.ShowDirsOnly`
- `QDialogButtonBox.{Ok,Cancel}` → `.StandardButton.*`
- `QDialog.Accepted` → `.DialogCode.Accepted`
- `QPainter.CompositionMode_SourceOver` → `.CompositionMode.*`
- `QDoubleValidator.StandardNotation` → `.Notation.*`
- `QFont.Bold` → `.Weight.Bold`

Method calls (`QApplication.processEvents`, `QFileDialog.getOpenFileName`, `QMessageBox.warning`/`question`, …) are untouched, and `QtCore.Qt` enums come from raw PyQt5 so they were already fine. Verified every scoped name resolves and equals its flat value in PyQt5 5.15.

### DIFF
--- a/Utils/SkyFit2.py
+++ b/Utils/SkyFit2.py
@@ -1263,7 +1263,7 @@ class CalibrationFilesDialog(QtWidgets.QDialog):
         # Pick folder
         dir_path = str(QtWidgets.QFileDialog.getExistingDirectory(
             self, "Select folder with FF/image files",
-            pt.dir_path, QtWidgets.QFileDialog.ShowDirsOnly))
+            pt.dir_path, QtWidgets.QFileDialog.Option.ShowDirsOnly))
         if not dir_path:
             return
 
@@ -1292,11 +1292,11 @@ class CalibrationFilesDialog(QtWidgets.QDialog):
             reply = QtWidgets.QMessageBox.question(
                 self, "Unsaved Changes",
                 "The current platepar has unsaved changes.\nSave before changing station?",
-                QtWidgets.QMessageBox.Save | QtWidgets.QMessageBox.Discard | QtWidgets.QMessageBox.Cancel,
-                QtWidgets.QMessageBox.Save)
-            if reply == QtWidgets.QMessageBox.Cancel:
+                QtWidgets.QMessageBox.StandardButton.Save | QtWidgets.QMessageBox.StandardButton.Discard | QtWidgets.QMessageBox.StandardButton.Cancel,
+                QtWidgets.QMessageBox.StandardButton.Save)
+            if reply == QtWidgets.QMessageBox.StandardButton.Cancel:
                 return
-            if reply == QtWidgets.QMessageBox.Save:
+            if reply == QtWidgets.QMessageBox.StandardButton.Save:
                 pt.savePlatepar()
 
         # Reload via changeStation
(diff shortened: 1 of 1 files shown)


Cybis320:
Thanks for the fix — the diagnosis is right and the scoped enum names are the correct solution (they resolve on both PyQt5 and PyQt6, so there's no risk on older stacks).

Two issues with the PR as it stands, though:

1. It targets master, but development happens on prerelease (same note as on #951). master only gets updated on releases, so a fix merged there would be overwritten. Could you rebase/retarget this onto prerelease?

2. On prerelease the patch is incomplete. Utils/SkyFit2.py has diverged a lot since master, and it now has more instances of the same flat enums, plus other enum classes that would crash the same way on your stack as soon as the next dialog opens.

If you could redo this against prerelease and sweep all the flat QtWidgets/QtGui enum accesses in SkyFit2 in one pass, that would be great.

GlassOnTin:
Thanks @Cybis320 — both addressed:

1. **Retargeted to `prerelease`.**
2. **Full sweep done.** All 40 flat `QtWidgets`/`QtGui` enum accesses in `SkyFit2.py` are now scoped — not just the original four:
   - `QMessageBox.{Yes,No,Cancel,Save,Discard}` → `.StandardButton.*`, `.ActionRole` → `.ButtonRole.ActionRole`, `.Warning` → `.Icon.Warning`
   - `QFileDialog.ShowDirsOnly` → `.Option.*`, `QDialogButtonBox.{Ok,Cancel}` → `.StandardButton.*`, `QDialog.Accepted` → `.DialogCode.Accepted`
   - `QPainter.CompositionMode_SourceOver` → `.CompositionMode.*`, `QDoubleValidator.StandardNotation` → `.Notation.*`, `QFont.Bold` → `.Weight.Bold`

Method calls (`processEvents`, `getOpenFileName`, `warning`/`question`, …) are untouched, and `QtCore.Qt` enums come from raw PyQt5 so they were already fine. I verified every scoped name resolves and equals its flat value in PyQt5 5.15.

Cybis320 (approved):
Verified and approving. I enumerated every `QtWidgets`/`QtGui` attribute chain on the PR head: nothing flat remains — only static method calls (correctly untouched) and scoped enums, summing to exactly the 40 sites claimed. All 14 distinct scoped names resolve and equal their flat values on PyQt5 5.15.11 here, and they're the required forms on PyQt6, so this is forward-compatible rather than just a proxy workaround. The diff is a pure mechanical substitution with no logic changes, and the flag OR-ing / result comparisons stay valid on both bindings. The bare-name PyQt5 imports in the ASTRA block are unaffected (raw PyQt5, no flat enum accesses there anyway).

Issue #954, opened 2026-08-01
On loss of network ObservationSummary generation can hang indefinitely


g7gpr:
If the connection to the github server is lost, then a git operation hangs indefinitely.

PR #955, opened 2026-08-01
Prevent observation summary hanging on network loss and handle symbolic references

g7gpr:
This pull request addresses issue #954 

This patch prevents a network loss causing observation summary generation to hang, by adding a 120 second timeout to the git operation. 

It also prevents a symbolic reference in the list of remotes causing an exception, returning that the remote branch time difference is not determined. 

At present

```
repository_lag_remote_days                : Not determined 
```

With this PR

```
remote_branch                             : systemd 
repository_lag_remote_days                : 0.302 
```

Also minor changes to commenting.

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -1088,7 +1088,7 @@ def updateCommitHistoryDirectory(remote_urls, target_directory):
     """ Clone only the commit history of a remote repository.
 
     Arguments:
-        remote_urls: [url] the remote url to be cloned/
+        remote_urls: [url] the remote url to be cloned
         target_directory: [path] the directory into which to clone.
 
     Return:
@@ -1108,14 +1108,24 @@ def updateCommitHistoryDirectory(remote_urls, target_directory):
             first_remote = False
             p = subprocess.Popen(["git", "clone", url, "--filter=blob:none", "--no-checkout"], cwd=target_directory,
                              stdout=subprocess.PIPE)
-            p.wait()
-            # this first remote might have been pulled in with the wrong local_name so rename it
+            try:
+                p.wait(timeout=120)
+            except subprocess.TimeoutExpired:
+                p.kill()
+                raise
+
+            # this first remote might have been pulled in with the wrong local_name so check and rename if required
             commit_repo_directory = os.path.join(target_directory, os.listdir(target_directory)[0])
             downloaded_remote_name = subprocess.check_output(["git", "remote"], cwd = commit_repo_directory).strip().decode('utf-8')
 
             if downloaded_remote_name != local_name:
                 p = subprocess.Popen(["git", "remote", "rename", downloaded_remote_name, local_name], cwd = commit_repo_directory)
-                p.wait()
+
+                try:
+                    p.wait(timeout=120)
+                except subprocess.TimeoutExpired:
+                    p.kill()
+                    raise
 
         else:
             # this is not the first remote so add another remote
@@ -1247,9 +1257,8 @@ def getRemoteBranchNameForCommit(repo, commit):
         if branch_stripped.startswith("remotes/"):
             remote_branch_name = branch_stripped
 
-    # If we are not at the HEAD, then get all the branches which contain the commit.
+    # Get all the branches that contain the commit and pick the most likely
 
-    # 2. Branches that *contain* the commit
     try:
         contains = subprocess.check_output(
             ["git", "branch", "-r", "--contains", commit],
@@ -1258,7 +1267,8 @@ def getRemoteBranchNameForCommit(repo, commit):
     except Exception:
         contains = []
 
-    contains = [c.strip() for c in contains if c.strip()]
+    # Eliminate any symbolic references
+    contains = [c.strip() for c in contains if c.strip() and "->" not in c.strip()]
 
     if contains:
         # If the branch is origin/main or origin/pre-release; then that is almost certainly where we are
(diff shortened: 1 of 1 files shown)


dvida (commented):
Reviewed for correctness. The diagnosis is right and both changes point the right way — I reproduced the symref failure locally, so that half is fixing something real: `git branch -r --contains <a commit on master>` really does emit `  origin/HEAD -> origin/master`, and that arrow line then reaches `getDateOfCommit`, where `git show` fails and the bare `except` at line 1892 turns the whole thing into `Not determined`. Nice catch.

A few gaps, though. Line numbers below are against this branch.

### The `git fetch` is still unguarded

`ObservationSummary.py:1135-1136` — in the non-first-remote branch, `git fetch --filter=blob:none` is left on a bare `p.wait()`. That is the same class of network operation as the clone you did guard, so any station with more than one remote (a dev box, or an operator who added `upstream`) still hangs exactly as #954 describes. `git remote add` at :1133 is local so it is low risk, but it may as well follow the same pattern.

This one matters most — as it stands the patch fixes the single-remote case only.

### `p.kill()` without a reap

:1114 and :1127 — after `kill()` the child is never `wait()`ed, so it is left defunct, and killing `git` does not necessarily take down its `git-remote-https` helper. There is already a correct teardown pattern in the tree at `Utils/GenerateTimelapse.py:836-846` (terminate → `wait(10)` → kill).

Simpler than either: `subprocess.run(cmd, cwd=..., timeout=...)` kills *and* reaps internally, and this file already uses `subprocess.run` at :95 and :501. That also lets you drop the `stdout=subprocess.PIPE` at :1110, which is worth dropping on its own — piping stdout and never draining it is a latent deadlock (pipe buffer fills, `wait()` blocks forever), independent of the network.

### The timeout leaves no trace

The re-raise lands in the bare `except:` at :1892 and becomes `repository_lag_remote_days: Not determined` with nothing logged. Since the point of #954 is a network stall that is hard to diagnose, an operator seeing `Not determined` should be able to find out why from the log. `log` is available at :70 and `log.error` is used a few lines up at :1885.

### The temp clone leaks on the new raise path

`daysBehind()` at :1296-1308 only calls `target_directory_obj.cleanup()` on the two `return` paths, so the new `raise` skips it. The `TemporaryDirectory` finalizer does reclaim it eventually, but in a long-lived `Reprocess` process a partial blobless clone of ~11800 commits accumulates per failure. `with tempfile.TemporaryDirectory() as target_directory:` handles it.

### Is 120 s enough?

A blobless clone of ~11800 commits over a slow rural station uplink can plausibly exceed 120 s, and a ceiling that is too tight converts a healthy station into a permanent `Not determined` — arguably worse than the hang, because it looks like a real answer. I would go to 300 s, and hoist it into a single module constant rather than repeating the literal at each call site (four sites once the fetch is covered).

### On the comments

`:1260` — replacing `# If we are not at the HEAD, then get all the branches which contain the commit.` was the right call, because the old comment described code that does not exist: the `--contains` query is unconditional, and its result *overrides* the `--points-at` result computed at :1245-1258, which now survives only as a fallback for when `contains` is empty. Two small things: the comment reads better flush against the `try` it describes, and the `--points-at` block above deserves a line saying it is the fallback, so that precedence reads as deliberate rather than accidental.

`:1270` — `# Eliminate any symbolic references` says the what but not the why. Worth naming the actual offending output (`origin/HEAD -> origin/master`), since that is the non-obvious part.

`:1271` — `"->" not in c` is looser than it needs to be. Git forbids spaces in ref names but not `>`, so a branch literally named `a->b` is legal and would be silently discarded. Matching `" -> "` costs nothing and is strictly safer.

The docstring fix at :1091 and dropping the orphaned `# 2.` marker are both good.

---

I have the above on a local branch on top of your commit — happy to push it or open a follow-up PR if you would rather take it than redo it, whichever you prefer.

### Separate pre-existing issues, just noting them — not for this PR

- `daysBehind()` returns a 2-tuple on success but a bare string at :1308, so the caller's `days_behind, remote_branch = daysBehind()` unpack raises `ValueError` and gets swallowed into `Not determined`. `(None, None)` would be more honest.
- `os.path.basename(remote_branch)` at :1891 mangles branch names containing a slash. Testing here, `g7gpr/fix/observationsummary/symrefs` is reported as `remote_branch: symrefs`. Flat names like your `systemd` example are unaffected, which is probably why this has not been noticed.
- The preferred-branch list at :1275 hardcodes the remote name `origin`, though the surrounding code deliberately supports arbitrary remote local names.
- Duplicate `import subprocess` at :28 and :59.

dvida:
Pushed the fixes to this branch as db22e994 (fast-forward on top of your commit, nothing rewritten) so you do not have to redo them. Covers everything from the review above: all four git calls now go through a `runGitCommand()` helper on `subprocess.run`, `GIT_TIMEOUT_SEC = 300` as a single constant, timeouts logged before re-raising, the temp clone held in a `with` block, and the symref filter matching `" -> "`.

One thing beyond the review: neither version checked the git exit code, so a failed clone fell through to `os.listdir(target_directory)[0]` and surfaced as a bare `IndexError` with no reason logged. The helper now captures stderr and logs it on a non-zero return code.

Verified locally in the `rms` env: `daysBehind()` resolves through the previously-unguarded `git fetch` path (this tree has two remotes); a commit on master resolves to `origin/master` rather than the symref line; and with the ceiling squeezed to 1 s the timeout logs which command stalled, propagates, and leaves no temp directory or defunct git process behind. Your call on the four pre-existing items at the end of the review — I left all of them alone.

g7gpr:
Taking only the basename is intentional, as is preferring origin as the branch name. Most operators will be on master, a few will be on prerelease. The purpose of this code is to identify if a station is on one of these, or something else entirely. 

I'll look at the other pre-existing issues, and if necessary raise a pr.

PR #961, opened 2026-08-10
Bound the fireball extractor with a reusable snapshot pool (fixes mid-night OOM)

Cybis320:
## Problem

The extractor snapshot introduced in the #935/#938 review round (finding B2) allocates a fresh ~236 MB `multiprocessing.Array` for every FF block and hands it to a fire-and-forget Extractor process. Extractors have no concurrency bound, so whenever extraction runs longer than the ~10 s block cadence (noisy sky, long `find3DLines`), extractors pile up, each pinning its own private snapshot. Memory grows without bound and the OS OOM-kills capture mid-night. Observed in production within days of deploying the squashed PRs; any station on `prerelease` with `enable_fireball_detection` is exposed.

Reverting B2 outright would restore the old unbounded-correctness trade-off instead: extractors reading the live ping-pong buffers race the refill whenever they outlive the ~10 s grace, silently corrupting FR clips - and the OOM frequency suggests that overrun condition is common, so the old race was likely firing regularly and going unnoticed.

## Fix

Replace the per-block allocation with a small reusable snapshot pool that keeps B2's correctness properties and master's boundedness:

- At most `MAX_EXTRACTOR_SNAPSHOTS = 2` buffers are ever allocated, lazily - on a healthy night one buffer exists and is reused for every block. Hard memory ceiling: 2 x ~236 MB per station.
- The block is still copied before the capture handshake is released, and the extractor still receives the base + shape - no refill race, no by-value pickle under `forkserver`/`spawn`.
- A buffer returns to the pool only after its extractor has exited, so reuse can never race a live reader.
- Two buffers absorb one over-cadence extraction with no loss (a bright fireball whose own extraction runs long, followed by its continuation block). Only a 3-deep backlog skips a block, with log visibility: an info line when the second buffer is allocated, a warning per skip with a running count, and an info line on catch-up. Previously both failure modes (the race on master, the pile-up after B2) were silent.
- The compressor's exit paths now join live extractors (bounded, 20 s) so a dawn shutdown does not kill an extractor mid-FR-write, and `Extractor.run()` sets the parent-death signal (same pattern as `RawFrameSaver`) so a hung extraction cannot be orphaned pinning its snapshot forever.

## Testing

- `Tests/test_DeadlockFixes.py` passes (16/16).
- A simulation of the pool state machine (mirroring the loop logic) asserts, across healthy-night / single-slow-fireball / sustained-overrun scenarios: the single buffer is reused, the cap is never exceeded, no buffer is ever owned twice, skips occur only at pool exhaustion, and extraction resumes when a buffer frees.

### DIFF
--- a/RMS/Compression.py
+++ b/RMS/Compression.py
@@ -45,6 +45,15 @@ from RMS.CompressionCy import compressFrames
 log = getLogger("rmslogger")
 
 
+# Maximum number of shared-memory snapshot buffers for the fireball extractor. Each holds a full
+# raw frame block (~236 MB at 720p), so this bounds the extractor's total memory. Two buffers
+# tolerate one over-cadence extraction with no loss - the case of a bright fireball whose own
+# extraction runs long, immediately followed by the block with its continuation. A backlog deeper
+# than that means extraction has fallen behind capture for a sustained stretch (in practice a
+# noise night - clouds, rain, insects), and further blocks are skipped to keep memory bounded.
+MAX_EXTRACTOR_SNAPSHOTS = 2
+
+
 class Compressor(multiprocessing.Process):
     """Compress list of numpy arrays (video frames).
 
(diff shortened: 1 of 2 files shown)


dvida:
Thanks for this, Luc — the diagnosis here is excellent, and it's what made the fix tractable: the unbounded per-block snapshot, the pile-up under over-cadence extraction, and the `forkserver`/`spawn` pickle-by-value trap all landed exactly right.

After weighing it for the memory-limited Raspberry Pis, I've gone a different route in #962. Rather than bounding the snapshots to a 2-buffer pool (which still adds up to ~472 MB on a 2 GB Pi), #962 restores the original pre-#935 design: the extractor reads the **live** ping-pong buffer directly, so memory stays fixed at the two existing frame buffers — no per-block allocation at all.

The trade-off is that this re-accepts the rare overrun→corrupt-clip race the snapshot was added to avoid. On a Pi that's the right call: worst case is one lost fireball clip on a slow night (FF files and detection are unaffected), which is cheaper than spending memory we don't have. And the corruption threshold (extraction outrunning the ~2-block buffer-recycle window) is looser than the OOM threshold was (extraction merely outrunning the ~10 s cadence), so it should fire less often than the pile-up did.

Your analysis fed directly into #962 — in particular passing the `mp.Array` **base** by handle (not the numpy view) to stay zero-copy under `spawn`, which is the key insight from this PR. Credited there.

Closing as superseded by #962. Thank you again for catching the OOM and working it through so thoroughly.

Cybis320:
Issue with the original approach we are now back at is that the fireball detector now fails silently due to a race condition under load.

dvida:
It has to be this way due to severe memory limitations on older systems. And I haven't seen fireballs getting missed since we dumped pi2s.

PR #962, opened 2026-08-10
Fix mid-night OOM (fixed-memory live-buffer extractor) and NumPy 2.x binary readers (#959)

### LINKED ISSUE #959: FRbinViewer stopped working after increasing  version of numpy on Trixie image
Hi, after recent update of requirements.txt in **prerelease** branch (which valid only for Trixie as it uses python 3.13.5):

`numpy>=2.3.0,<3.0.0 ; python_version>='3.13'`

and increasing numpy to version : 2.5.1

FRbinViewer fails with error:
``` python
  File "/home/rms/source/RMS/Utils/FRbinViewer.py", line 767, in <module>
    retval = view(dir_path, ff_match, fr, config, save_frames=cml_args.extract,
        extract_format=cml_args.extractformat, hide=cml_args.hide, avg_background=cml_args.avg,
        split=cml_args.split, add_timestamp=cml_args.timestamp, add_frame_number=cml_args.framenumber,
                  append_ff_to_video=cml_args.append_ff_to_video, add_shower_name=add_shower_name,
                  associations=associations)
  File "/home/rms/source/RMS/Utils/FRbinViewer.py", line 291, in view
    fr = FRbin.read(dir_path, fr_path)
  File "/home/rms/source/RMS/RMS/Formats/FRbin.py", line 187, in read
    yc.append(int(np.fromfile(fid, dtype=np.uint32, count=1)))
              ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

dvida:
Two independent reliability fixes for `prerelease`.

## 1. Mid-night OOM: revert the fireball extractor to a fixed-memory live buffer

The extractor snapshot introduced in the #935/#938 round (finding B2) hands each Extractor its own ~236 MB `multiprocessing.Array` copy of the frame block, with no concurrency bound. On noisy nights (slow `find3DLines`) extractions outrun the ~10 s block cadence, pile up, each pinning a copy, and the OS OOM-kills capture mid-night on memory-limited Raspberry Pis.

This is an **alternative to #961**. Rather than bounding the copies to a 2-buffer pool (which still costs up to +472 MB on the Pi), this restores the **original pre-#935 design**: the extractor reads the **live** ping-pong buffer directly, so memory stays fixed at the two existing frame buffers — no per-block allocation, OOM impossible.

- Pass the `mp.Array` **base** (not a numpy view) to the extractor, so it crosses the process boundary **by handle** under `fork` and `forkserver`/`spawn` alike. Passing the view is what pickled the ~236 MB block by value under `spawn`; passing the base does not. Verified a locked `Array` base rebuilds correctly in a spawned child.
- The extractor rebuilds its view via `.get_obj()` (the live base is a locked `Array`), matching the pattern in `Compression.py`/`BufferedCapture.py`.
- Trade-off: a rare extraction that outruns the buffer-recycle window (~2 blocks) can corrupt its own FR clip — the original, accepted behavior. FF files and meteor detection are unaffected. Worst case is a lost fireball clip, not a dead night.

Files: `RMS/Compression.py`, `RMS/VideoExtraction.py`.

## 2. Binary readers broken under NumPy 2.x — fixes #959

NumPy 2.0 removed the implicit conversion of a size-1 array to a Python scalar via `int()`/`float()`. `np.fromfile(..., count=1)` returns a length-1 array, so `int(np.fromfile(...))` raises `TypeError: only 0-dimensional arrays can be converted to Python scalars` on NumPy ≥ 2.0. This surfaced when Trixie (Python 3.13) pulled `numpy>=2.3`, breaking `FRbinViewer` via `FRbin.read` (#959).

Index the size-1 array before converting — `int(np.fromfile(...)[0])` — the form already used at `FRbin.py:176` and throughout `StarCatalog.py`. Applied to all **39 affected sites** across the binary readers: `Vid.py`, `FFbin.py`, `FTfile.py`, `FieldIntensities.py`, `FRbin.py`. Verified with an `FRbin.read` round-trip returning correct Python `int`s.

Files: `RMS/Formats/{Vid,FFbin,FTfile,FieldIntensities,FRbin}.py`.

The two fixes are unrelated (extractor memory vs a NumPy API change); bundled here for one review round.

### DIFF
--- a/RMS/Compression.py
+++ b/RMS/Compression.py
@@ -383,21 +383,6 @@ class Compressor(multiprocessing.Process):
             # Run the compression
             compressed, field_intensities = self.compress(frames)
 
-            # Snapshot the raw block for the extractor BEFORE releasing the capture
-            # handshake: once start_time returns to 0 the buffer is eligible for refill,
-            # and the extractor reads it seconds later - the old code raced the refill
-            # under 'fork' and pickled the whole ~236 MB block by value under
-            # 'forkserver'/'spawn' (stalling compression and spiking RSS). A shared
-            # mp.Array snapshot costs one memcpy while the handshake is held and crosses
-            # the process boundary without pickling.
-            frames_snapshot_base = None
-            if self.config.enable_fireball_detection:
-                frames_snapshot_base = multiprocessing.Array(
-                    ctypes.c_uint8, int(frames.size), lock=False)
-                snapshot_view = np.frombuffer(frames_snapshot_base,
-                    dtype=np.uint8).reshape(frames.shape)
-                np.copyto(snapshot_view, frames)
-
             # Once the compression is done, tell the capture thread to keep filling the buffer
             if buffer_one:
                 self.start_time1.value = 0
@@ -428,11 +413,24 @@ class Compressor(multiprocessing.Process):
             # Save the extracted intensities per every field
             FieldIntensities.saveFieldIntensitiesBin(field_intensities, self.data_dir, filename_micros)
 
-            # Run the extractor (on the pre-handshake snapshot, never the live buffer)
+            # Run the extractor on the LIVE ping-pong buffer (fixed memory: no per-block
+            # copy). With two buffers the extractor has at most about one block before
+            # capture, having filled the other buffer, wraps back and starts overwriting
+            # this one - and less than that in practice, since the handshake was released
+            # (start_time -> 0) and the FF/JPG/intensities saved above before we get here.
+            # A normal-night extraction finishes inside that window; a slow one (noisy sky,
+            # long find3DLines) can outrun it, and then capture overwrites the block
+            # mid-read and that single FR clip is corrupted. That is the original pre-#935
+            # behavior, kept deliberately: on memory-limited Raspberry Pis a rare lost
+            # fireball clip is cheaper than a per-block snapshot copy, and FF files and
+            # detection are unaffected. Pass the mp.Array BASE, not a numpy view - the base
+            # crosses the process boundary by handle under 'fork' and 'forkserver'/'spawn'
+            # alike, while a view would be pickled by value (~236 MB, stalling compression
+            # and spiking RSS).
             if self.config.enable_fireball_detection:
+                frames_base = self.array1_base if buffer_one else self.array2_base
                 extractor = Extractor(self.config, self.data_dir)
-                extractor.start(frames_snapshot_base, frames.shape, compressed,
-                    filename_millis)
+                extractor.start(frames_base, frames.shape, compressed, filename_millis)
 
                 log.debug('Extractor started for: ' + filename_millis)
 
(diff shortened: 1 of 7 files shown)


dvida:
Thanks — the timing correction is right and I've fixed it in 2cc19408. The grace is about **one** block, not two: capture wraps back to this buffer after filling only the *other* one, and it's less than a full block in practice since the handshake is released (and FF/JPG/intensities saved) before the extractor starts. The comment overstated it; corrected.

On the substance, though, the overrun→corrupt-clip is **intentional and accepted**, not a regression. This PR restores the original pre-#935 design where the extractor reads the live ping-pong buffer directly, and the corruption on a slow extraction is the known trade-off that design has always carried.

The two remedies suggested here were both considered and deliberately rejected for the target hardware (memory-limited Raspberry Pis):

- **Retain buffer ownership until extraction completes** would serialize capture on every extraction — capture would stall waiting for the buffer to free, dropping frames on the floor. That's worse than a rare bad clip.
- **Bounded immutable storage** is exactly the reusable snapshot pool from #961, which costs up to +472 MB per station. That's the memory we're specifically trying not to spend here; #961 was closed in favor of this approach for that reason.

Net: on a normal night, extraction finishes well inside the window. Only a slow extraction (noisy sky, long `find3DLines`) outruns it and corrupts that one FR clip — and those nights don't produce usable fireball clips anyway. FF files and meteor detection are never affected. A lost clip on a bad night is the accepted cost of keeping capture memory fixed on a Pi.

Cybis320 (commented):
Posting this post-merge for the record, since the trade-off here will come back when Python 3.14 lands.

First: the by-handle base passing is a clear improvement over both prior designs — no by-value pickle under `spawn`, no per-block allocation, and the race is now documented in-code instead of lurking silently. Three residual points worth keeping on the radar:

1. **The refill race is new under `forkserver`/`spawn`.** In the pre-#935 code the race was fork-only: under `spawn`, `Process.start()` pickled the view by value, which acted as an accidental snapshot — the child read a private copy. With the base passed by handle, a spawned child now reads the **live** buffer, and on a Pi it spends seconds importing numpy/RMS before its first read, so startup latency eats into the ~one-block grace window. On 3.14 (`forkserver` default per #938), the race will fire more easily than it ever did under `fork`.

2. **Extractor concurrency is still unbounded.** A pile-up no longer pins ~236 MB copies, but under `spawn` each queued extractor is a full interpreter (~100 MB RSS), and under `fork` it still competes for CPU with capture and compression on exactly the nights the system is already behind. "OOM impossible" holds for the frame buffers; a deep noise-night process backlog on a 2 GB Pi is a softer version of the same story.

3. **Overruns remain unobservable.** The mid-night OOM frequency in production implies extraction outruns the ~10 s cadence routinely — which means the race corrupts or loses FR clips routinely, and nothing logs it. A behavior-neutral counter (prune-and-count live extractors each block, warn when the count grows) would give real overrun-rate data before 3.14 forces this trade-off to be re-decided. Happy to send that as a small follow-up PR if there's interest.

PR #965, opened 2026-08-11
Replay pre-init config warnings into the night log (follow-up to #924)

dvida:
Follow-up to #924 (merged), same theme — make night-time problems visible in the uploaded night log. #924 covered night *product* failures; this round closes the gap it left and tidies a few related spots.

## 1. Replay pre-init config warnings into the night log

Config parsing runs **before** `initLogging` in every entry point, so config-validation warnings — clamped FPS, bad binning factor/method, jpg/png quality ranges, FOV check, band ratios, "upload disabled on the default station code" — only ever hit the console and never the uploaded night log, which is exactly where you'd look to answer *"did my station configure correctly?"*.

Fix: a small `EarlyRecordBuffer` on the root logger (`RMS/Logger.py`) collects WARNING+ records emitted before logging is up; `initLogging` drains it and replays them into the night log with their original timestamps (`InRmsFilter` still drops non-RMS noise on replay).

`ConfigReader.parse()` installs the buffer as its first step — the single funnel every config load passes through — so it rides along with the parser and needs no per-entry-point wiring. `installEarlyLogBuffer()` is idempotent and **self-managing**: a no-op once real logging is up (a `QueueHandler` on the root logger, main process or child), so a re-parse after startup can't leave an undrained buffer, and a bare `import RMS.ConfigReader` no longer mutates the caller's root logger.

## 2. Tidy night-product logging (`Reprocess.py`)

Collapse the night-product exception handlers onto `log.warning(..., exc_info=True)` instead of hand-formatting the traceback into two warning lines, and route the last FT-archiving `print()` calls through logging so they land in the night log with a level and timestamp.

## 3. FRbinViewer key legend

Default the on-screen legend to off so it never covers the image unless asked for (toggle with `h`); print the key banner to the console when a file is opened instead.

## 4. README

Drop `RMS.ClearSkyDetector` from the runnable-module list — it has no CLI entry point.

## Testing

- `installEarlyLogBuffer`: verified a bare `import RMS.ConfigReader` installs no buffer; the buffer captures WARNING+ and ignores INFO; `_drainEarlyLogBuffer` returns the records; and it no-ops once a `QueueHandler` is on the root logger.
- All changed modules compile; `ConfigReader`/`Logger` import clean.

### DIFF
--- a/RMS/ConfigReader.py
+++ b/RMS/ConfigReader.py
@@ -21,6 +21,7 @@ import math
 import os
 import sys
 from RMS.CompileArgs import getCompileArgs
+from RMS.Logger import installEarlyLogBuffer
 from RMS.Misc import getRmsRootDir
 from Utils.GenerateTimelapse import isFfmpegWorking
 import matplotlib.colors as mcolors
@@ -736,6 +737,12 @@ def parse(path, strict=True):
         config: [Config]
 
     """
+
+    # Buffer any warnings raised while parsing (clamped FPS, bad binning, upload disabled on the
+    # default station code, ...) so initLogging can replay them into the night log. Config parsing
+    # runs before logging is initialized in every entry point, and every config load funnels
+    # through here. Idempotent and a no-op once logging is up (see installEarlyLogBuffer).
+    installEarlyLogBuffer()
 
     delimiter = ";"
 
--- a/RMS/Logger.py
+++ b/RMS/Logger.py
@@ -148,6 +148,109 @@ class InRmsFilter(logging.Filter):
         return any(_inside(p, root) for root in self.allowed_dirs)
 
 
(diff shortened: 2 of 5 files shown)


dvida:
Good catch — fixed in c52465b8. The double print was real: `_default_handler` prints the config warning to stderr before logging is up, and the replay then went through the queue to *both* the listener's file and console handlers, so the console showed it twice.

Fix: replayed records are now tagged (`replayed_from_early_buffer`), and a `_SuppressReplayedOnConsole` filter on the listener's **console** handler drops them. The **file** handler still receives them, so the night log keeps the warning exactly once, and the console shows it once (the pre-init stderr print). Verified the filter drops tagged records and passes normal ones.

PR #966, opened 2026-08-15
Add trackstack meteor mask

tammojan:
This adds an option to stack only the part of the images which have the meteor detection. With this option, in the resulting trackstack there are less planes and satellites.

Example before:
trackstack_without

After:
trackstack_with_filter

### DIFF
--- a/Utils/TrackStack.py
+++ b/Utils/TrackStack.py
@@ -15,7 +15,7 @@ import RMS.ConfigReader as cr
 from RMS.Astrometry.ApplyAstrometry import xyToRaDecPP, raDecToXYPP
 from RMS.Astrometry.Conversions import date2JD, jd2Date
 from RMS.Formats.FFfile import validFFName, getMiddleTimeFF
-from RMS.Formats.FTPdetectinfo import readFTPdetectinfo
+from RMS.Formats.FTPdetectinfo import readFTPdetectinfo, validDefaultFTPdetectinfo
 from RMS.Formats.FFfile import read as readFF
 from RMS.Formats.Platepar import Platepar
 from RMS.Math import angularSeparation
@@ -28,7 +28,7 @@ from RMS.QueuedPool import QueuedPool
 import time
 import datetime
 
-def find_ftp_file(dir_path, config):
+def findFTPFile(dir_path, config):
     if os.path.isfile(os.path.join(dir_path,'.config')):
         tmpcfg = cr.loadConfigFromDirectory('.config', dir_path)
     else:
@@ -42,15 +42,16 @@ def find_ftp_file(dir_path, config):
 
     return ftp_list[0]
 
-def make_mask(ftp_points, initial_mask):
+def makeMeteorMask(ftp_points_list, initial_mask):
     """Make a mask in which only the meteor is visible"""
-    meteor_mask = np.zeros_like(initial_mask.img)
+    meteor_mask = np.zeros_like(initial_mask)
 
-    pts = np.array([[round(p[2]), round(p[3])] for p in ftp_points], dtype=np.int32)
-    meteor_mask = cv2.polylines(meteor_mask, [pts], False, 255, 1)
+    for ftp_points in ftp_points_list:
+        pts = np.array([[round(p[2]), round(p[3])] for p in ftp_points], dtype=np.int32)
+        meteor_mask = cv2.polylines(meteor_mask, [pts], False, 255, 1)
 
     meteor_mask = cv2.dilate(meteor_mask, np.ones((150, 150), np.uint8))
-    return np.minimum(meteor_mask, initial_mask.img)
+    return np.minimum(meteor_mask, initial_mask)
 
 def trackStack(dir_paths, config, border=5, background_compensation=True, 
         hide_plot=False, showers=None, darkbackground=False, out_dir=None,
@@ -115,7 +116,7 @@ def trackStack(dir_paths, config, border=5, background_compensation=True,
         for dir_path in dir_paths: 
 
             try:
-                ftp_file = find_ftp_file(dir_path, config)
+                ftp_file = findFTPFile(dir_path, config)
             except FileNotFoundError as e:
                 print(e)
                 return False
@@ -144,13 +145,13 @@ def trackStack(dir_paths, config, border=5, background_compensation=True,
     if mask_meteors:
         for dir_path in dir_paths:
             try:
-                ftp_file = find_ftp_file(dir_path, config)
+                ftp_file = findFTPFile(dir_path, config)
             except FileNotFoundError as e:
                 print(e)
                 return False
 
             for ftp_entry in readFTPdetectinfo(os.path.dirname(ftp_file), os.path.basename(ftp_file)):
-                ftp_points[(ftp_entry[0], ftp_entry[2])] = ftp_entry[-1]
+                ftp_points.setdefault(ftp_entry[0], []).append(ftp_entry[-1])
 
     # Take the platepar with the middle time as the reference one
     ff_found_list = []
@@ -450,15 +451,9 @@ def stackFrame(ff_name, recalibrated_platepars, mask, border, pp_ref, img_size, 
     max_deavg = maxpixel - avepixel
 
     if mask_meteors:
-        for i in range(1, 10):
-            # In case there are multiple meteors in a frame, mask around the first one.
-            # When using a shower filter, this may be the wrong one, we'll live with that.
-            try:
-                ff_mask = make_mask(ftp_points[(os.path.basename(ff_name), i * 1.0)], mask)
-                mask_deavg[ff_mask == 0] = 0
-                break
-            except KeyError:
-                pass  # Fall back to using the entire image
+        meteor_tracks = ftp_points.get(ff_basename, [])
+        ff_mask = makeMeteorMask(meteor_tracks, mask.img)
+        max_deavg[ff_mask == 0] = 0
 
     # Normalize the background brightness by applying a large-kernel median filter to avepixel
     if background_compensation:
--- a/Utils/TrackStack.py
+++ b/Utils/TrackStack.py
@@ -45,9 +45,11 @@ def find_ftp_file(dir_path, config):
 def make_mask(ftp_points, initial_mask):
     """Make a mask in which only the meteor is visible"""
     meteor_mask = np.zeros_like(initial_mask.img)
-    meteor_mask = cv2.line(meteor_mask, (round(ftp_points[0][2]), round(ftp_points[0][3])),
-                          (round(ftp_points[-1][2]), round(ftp_points[-1][3])), 255, 1)
-    meteor_mask = cv2.dilate(meteor_mask, np.ones((150, 150)))
+
+    pts = np.array([[round(p[2]), round(p[3])] for p in ftp_points], dtype=np.int32)
+    meteor_mask = cv2.polylines(meteor_mask, [pts], False, 255, 1)
+
+    meteor_mask = cv2.dilate(meteor_mask, np.ones((150, 150), np.uint8))
     return np.minimum(meteor_mask, initial_mask.img)
 
 def trackStack(dir_paths, config, border=5, background_compensation=True, 
@@ -74,6 +76,7 @@ def trackStack(dir_paths, config, border=5, background_compensation=True,
         draw_constellations: [bool] Show constellation lines on stacked image
         one_core_free: [bool] leave one core free whilst processing
         textoption: [int] 0 - no text, 1 - filename, 2 - stationID, date, meteor count overlayed
+        mask_meteors: [bool] use only a portion of the image around a detection
     """
     start_time = time.time()
     # normalise the path in a platform neutral way
@@ -437,25 +440,25 @@ def stackFrame(ff_name, recalibrated_platepars, mask, border, pp_ref, img_size, 
     stack_x = stack_x[filter_arr]
     stack_y = stack_y[filter_arr]
 
-    ff_mask = mask.img
+    # Apply the mask to maxpixel and avepixel
+    maxpixel = copy.deepcopy(ff.maxpixel)
+    maxpixel[mask.img == 0] = 0
+    avepixel = copy.deepcopy(ff.avepixel)
+    avepixel[mask.img == 0] = 0
+
+    # Compute deaveraged maxpixel image
+    max_deavg = maxpixel - avepixel
+
     if mask_meteors:
         for i in range(1, 10):
             # In case there are multiple meteors in a frame, mask around the first one.
             # When using a shower filter, this may be the wrong one, we'll live with that.
             try:
                 ff_mask = make_mask(ftp_points[(os.path.basename(ff_name), i * 1.0)], mask)
+                mask_deavg[ff_mask == 0] = 0
                 break
             except KeyError:
                 pass  # Fall back to using the entire image
-
-    # Apply the mask to maxpixel and avepixel
-    maxpixel = copy.deepcopy(ff.maxpixel)
-    maxpixel[ff_mask == 0] = 0
-    avepixel = copy.deepcopy(ff.avepixel)
-    avepixel[ff_mask == 0] = 0
-
-    # Compute deaveraged maxpixel image
-    max_deavg = maxpixel - avepixel
 
     # Normalize the background brightness by applying a large-kernel median filter to avepixel
     if background_compensation:
@@ -564,7 +567,7 @@ if __name__ == "__main__":
                             help="""Leave at least one core free""")
 
     arg_parser.add_argument('--mask-meteors', action="store_true",
-                            help="""Render only the part of the image around the meteor to suppress planes and satellites (works best for large trackstacks""")
+                            help="""Render only the part of the image around the meteor to suppress planes and satellites (works best for large trackstacks)""")
 
     # Parse the command line arguments
     cml_args = arg_parser.parse_args()
--- a/Utils/TrackStack.py
+++ b/Utils/TrackStack.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 from __future__ import print_function
 
 import os, sys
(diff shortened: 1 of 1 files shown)


Utils/TrackStack.py
```diff
@@ -411,11 +429,24 @@ def stackFrame(ff_name, recalibrated_platepars, mask, border, pp_ref, img_size,
...
+        for i in range(1, 10):
+            # Attempt to make something work for multiple meteors in one frame.
+            # Some of them may not be in ftp_points because of a shower filter.
+            # This is not water tight.
+            try:
+                ff_mask = make_mask(ftp_points[(os.path.basename(ff_name), i * 1.0)], mask)
+                break
+            except KeyError:
+                raise RuntimeError(f"Can't find {(os.path.basename(ff_name), i * 1.0)} in {list(ftp_points.keys())}")
+                pass
```
dvida: **Blocking — leftover debug code.** The `raise RuntimeError(...)` needs to go (and the `pass` after it is unreachable). As written, every FF without a detection raises here. `QueuedPool` catches it, logs the traceback and returns `None`, so you get one traceback per detection-less FF — each interpolating the entire key list — the frame contributes nothing, and `finished_count` is never incremented so the progress print stalls short of 100%.

**Blocking — the loop only ever masks meteor #1.** `break` on the first hit plus contiguous meteor numbering from 1 means the body runs exactly once. An FF with two meteors keeps only the first, contrary to the comment. Also, `ftp_points` is unfiltered by shower, so the shower filter isn't the reason for a miss; detection-less FFs are.

Both go away if `ftp_points` maps `ff_name -> [meteor_meas, ...]`:

```python
ff_mask = mask.img
if mask_meteors:
    meteor_tracks = ftp_points.get(ff_basename, [])
    if meteor_tracks:
        ff_mask = makeMeteorMask(meteor_tracks, mask.img)
```

No magic `10`, no `i * 1.0`, no try/except as control flow — and it handles multiple meteors for real. Worth deciding explicitly what a detection-less FF should contribute when `mask_meteors` is on: currently it falls back to the full station mask, so its planes and satellites go in unmasked, which is the opposite of the intent.

Utils/TrackStack.py
```diff
@@ -88,20 +112,7 @@ def trackStack(dir_paths, config, border=5, background_compensation=True,
...
-            ftp_list = glob(os.path.join(dir_path, 'FTPdetectinfo_{}*.txt'.format(tmpcfg.stationID)))
-            ftp_list = [x for x in ftp_list if 'backup' not in x and 'unfiltered' not in x]
-            ftp_list.sort() 
-
-            if len(ftp_list) < 1:
-                print('unable to find FTPdetect file in {}'.format(dir_path))
-                return False
-            
-            ftp_file = ftp_list[0] 
+            ftp_file = find_ftp_file(dir_path, config)
```
dvida: **Blocking — regression.** The inline block this replaced did `return False` from `trackStack()` when no FTPdetectinfo was found. `find_ftp_file` still returns `False`, but nothing checks it, so `showerAssociation(config, [False], ...)` gets called with a bool where a path is expected. Please restore the early return here (and at the new call site on line 140).

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
...
+        tmpcfg = config
+    ftp_list = glob(os.path.join(dir_path, 'FTPdetectinfo_{}*.txt'.format(tmpcfg.stationID)))
+    ftp_list = [x for x in ftp_list if 'backup' not in x and 'unfiltered' not in x]
+    ftp_list.sort()
+
+    if len(ftp_list) < 1:
+        print('unable to find FTPdetect file in {}'.format(dir_path))
+        return False
+
+    return ftp_list[0]
```
dvida, line 43: Returning `False` as a sentinel is what makes the two call sites silently misbehave. Since `RMS/Formats/FTPdetectinfo.py:251` already raises `FileNotFoundError` for this case, raising here would be consistent with the rest of the codebase; otherwise return `None` and check it. Either way both call sites need to handle it.

Utils/TrackStack.py
```diff
@@ -123,6 +134,13 @@ def trackStack(dir_paths, config, border=5, background_compensation=True,
                     ff_list.append(file_name)    
     ff_list = list(set(ff_list))
 
+    ftp_points = {}
+    if mask_meteors:
+        for dir_path in dir_paths:
+            ftp_file = find_ftp_file(dir_path, config)
+            for ftp_entry in readFTPdetectinfo(os.path.dirname(ftp_file), os.path.basename(ftp_file)):
```
dvida, line 153: If `find_ftp_file` returned `False`, this is `os.path.dirname(False)` → `TypeError: expected str, bytes or os.PathLike object, not bool`. Needs the same check as the shower branch.

Separately (line 140): with both `--shower` and `--mask-meteors`, `find_ftp_file` now runs twice per directory, re-globbing and re-loading `.config` each time. Minor, but resolving the FTPdetectinfo path once per directory and reusing it would be tidier.

Utils/TrackStack.py
```diff
@@ -123,6 +134,13 @@ def trackStack(dir_paths, config, border=5, background_compensation=True,
...
+                ftp_points[(ftp_entry[0], ftp_entry[2])] = ftp_entry[-1]
```
dvida: Keying on `(ff_name, meteor_No)` means only one meteor per key and forces the `i * 1.0` probing loop downstream. `ftp_points.setdefault(ftp_entry[0], []).append(ftp_entry[-1])` gives you every meteor per FF in one pass, and lets `stackFrame` do a plain `.get(ff_basename, [])`.

Utils/TrackStack.py
```diff
@@ -411,11 +429,24 @@ def stackFrame(ff_name, recalibrated_platepars, mask, border, pp_ref, img_size,
...
+                raise RuntimeError(f"Can't find {(os.path.basename(ff_name), i * 1.0)} in {list(ftp_points.keys())}")
+                pass
+
     # Apply the mask to maxpixel and avepixel
     maxpixel = copy.deepcopy(ff.maxpixel)
-    maxpixel[mask.img == 0] = 0
+    maxpixel[ff_mask == 0] = 0
     avepixel = copy.deepcopy(ff.avepixel)
-    avepixel[mask.img == 0] = 0
+    avepixel[ff_mask == 0] = 0
```
dvida: Masking `avepixel` is what removes the sky background rather than just the planes. Planes and satellites reach the stack via `max_deavg = maxpixel - avepixel`, so masking that is sufficient; stars appear in both images and largely cancel there, coming instead from the avepixel blend. Because `ones_img` is derived from the masked `avepixel`, masking it means `avg_stack_count` only increments inside the meteor strips, so the blended background is zero wherever no strip landed and the auto-crop shrinks to the union of the strips.

Suggestion — drop the masking here and on line 447, and mask after the subtraction instead:

```python
max_deavg = maxpixel - avepixel
if mask_meteors:
    max_deavg[meteor_mask == 0] = 0
```

Same plane/satellite suppression, full star field retained. If the strips-only look is what you're after, that's a legitimate choice — just document it in the `--mask-meteors` help.

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
...
+        return False
+
+    return ftp_list[0]
+
+def make_mask(ftp_points, initial_mask):
+    """Make a mask in which only the meteor is visible"""
+    meteor_mask = np.zeros_like(initial_mask.img)
+    meteor_mask = cv2.line(meteor_mask, (round(ftp_points[0][2]), round(ftp_points[0][3])),
+                          (round(ftp_points[-1][2]), round(ftp_points[-1][3])), 255, 1)
+    meteor_mask = cv2.dilate(meteor_mask, np.ones((150, 150)))
```
dvida: The 150x150 kernel is a fixed +/-75 px halo regardless of sensor resolution or `--scalefactor` — about a fifth of the frame height at 1280x720, a thin line at 4K. Worth exposing as a keyword argument (default 150) and/or scaling it off `pp_ref.Y_res`.

Two smaller points: `np.ones((150, 150))` is float64, so prefer `np.ones((150, 150), np.uint8)` or `cv2.getStructuringElement(cv2.MORPH_RECT, (150, 150))`; and this dilation runs per FF in every worker, so the structuring element could be built once.

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
...
+        print('unable to find FTPdetect file in {}'.format(dir_path))
+        return False
+
+    return ftp_list[0]
+
+def make_mask(ftp_points, initial_mask):
+    """Make a mask in which only the meteor is visible"""
+    meteor_mask = np.zeros_like(initial_mask.img)
+    meteor_mask = cv2.line(meteor_mask, (round(ftp_points[0][2]), round(ftp_points[0][3])),
+                          (round(ftp_points[-1][2]), round(ftp_points[-1][3])), 255, 1)
```
dvida: A single segment from the first to the last centroid under-covers fragmenting events and strongly curved (long, near-horizon) meteors. `cv2.polylines` over all the centroids is the same amount of code and is exact:

```python
pts = np.array([[round(p[2]), round(p[3])] for p in ftp_points], dtype=np.int32)
meteor_mask = cv2.polylines(meteor_mask, [pts], False, 255, 1)
```

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
 import time
 import datetime
 
+def find_ftp_file(dir_path, config):
+    if os.path.isfile(os.path.join(dir_path,'.config')):
+        tmpcfg = cr.loadConfigFromDirectory('.config', dir_path)
+    else:
+        tmpcfg = config
+    ftp_list = glob(os.path.join(dir_path, 'FTPdetectinfo_{}*.txt'.format(tmpcfg.stationID)))
+    ftp_list = [x for x in ftp_list if 'backup' not in x and 'unfiltered' not in x]
```
dvida, line 37: This filter duplicates `validDefaultFTPdetectinfo()` in `RMS/Formats/FTPdetectinfo.py:31`, and misses `uncalibrated`, which that predicate also excludes. `findFTPdetectinfoFile(path)` at `RMS/Formats/FTPdetectinfo.py:251` covers most of this too. The existing helper doesn't do the `stationID`/`.config` matching, so a thin wrapper is defensible — but it should call the shared predicate rather than re-implementing the substring checks.

(To be clear, the duplication is pre-existing — this PR just moved it into a function. Since it's being touched anyway, it's a good moment to switch.)

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
 import time
 import datetime
 
+def find_ftp_file(dir_path, config):
```
dvida: Naming: this file and the wider codebase use camelCase for functions (`trackStack`, `stackFrame`, `getArray`, `shouldInclude`). Suggest `findFTPFile` here and `makeMeteorMask` on line 46.

Also, `find_ftp_file` has no docstring while everything else in this module does.

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
...
+
+    if len(ftp_list) < 1:
+        print('unable to find FTPdetect file in {}'.format(dir_path))
+        return False
+
+    return ftp_list[0]
+
+def make_mask(ftp_points, initial_mask):
+    """Make a mask in which only the meteor is visible"""
+    meteor_mask = np.zeros_like(initial_mask.img)
```
dvida: This takes a `MaskStructure` but returns a bare `ndarray`, and `initial_mask` is really the station mask. Taking `mask.img` directly would be clearer, and would make the function unit-testable without constructing a `MaskStructure`.

Utils/TrackStack.py
```diff
@@ -26,11 +28,33 @@
...
+    meteor_mask = cv2.line(meteor_mask, (round(ftp_points[0][2]), round(ftp_points[0][3])),
+                          (round(ftp_points[-1][2]), round(ftp_points[-1][3])), 255, 1)
+    meteor_mask = cv2.dilate(meteor_mask, np.ones((150, 150)))
+    return np.minimum(meteor_mask, initial_mask.img)
 
 def trackStack(dir_paths, config, border=5, background_compensation=True, 
         hide_plot=False, showers=None, darkbackground=False, out_dir=None,
         scalefactor=None, draw_constellations=False, one_core_free=False,
-        textoption=0):
+        textoption=0, mask_meteors=False):
```
dvida, line 59: `mask_meteors` is missing from the `trackStack` docstring's keyword-argument list just below.

Utils/TrackStack.py
```diff
@@ -526,6 +557,9 @@ def getArray(size, shared_arr):
     arg_parser.add_argument('--freecore', action="store_true",
                             help="""Leave at least one core free""")
 
+    arg_parser.add_argument('--mask-meteors', action="store_true",
+                            help="""Render only the part of the image around the meteor to suppress planes and satellites (works best for large trackstacks""")
```
dvida: Unbalanced parenthesis: `"...(works best for large trackstacks"`. Also worth stating here whether the sky background outside the meteor is intentionally dropped (see the `avepixel` comment).

Utils/TrackStack.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
dvida, line 1: The shebang and the `100644` -> `100755` mode change are unrelated to the feature. Only 4 of the `Utils/*.py` scripts carry a shebang, so this is a bit inconsistent as housekeeping. Not harmful — but I'd rather it were a separate commit, or applied across the directory at once.

dvida (changes requested):
Thanks for this — the idea is a good one, and the before/after images make the case nicely. Suppressing planes and satellites in long trackstacks is something people have asked for.

There are three things that need fixing before this can go in, plus one design question worth settling.

**Blocking**

1. A leftover debug `raise RuntimeError(...)` is still in the `except KeyError` branch (with a dead `pass` after it). This fires for every FF that has no detection — and without `--shower`, `ff_list` is *every* valid FF in the directory, so that's most of them. `QueuedPool` catches worker exceptions (`RMS/QueuedPool.py:475`), prints the traceback and returns `None`, so the run doesn't die; instead you get one traceback per detection-less FF (each dumping the full key list), those frames silently contribute nothing, and `finished_count` is never incremented for them, so the progress print never reaches 100%. I suspect this is actually why your example image looks right — the detection-less FFs are being dropped by the exception rather than by intent.

2. The extracted `find_ftp_file()` still returns `False` on failure, but neither call site checks it any more. Previously the inline block did `return False` from `trackStack()`. Now `False` flows into `showerAssociation(config, [False], ...)` in the shower branch, and into `os.path.dirname(False)` in the new mask branch, which raises `TypeError: expected str, bytes or os.PathLike object, not bool`.

3. The `for i in range(1, 10)` loop doesn't do what its comment says. `break` on the first hit, plus meteor numbering that starts at 1 and is contiguous, means the body runs exactly once — only meteor #1 is ever masked, so an FF with two meteors keeps only the first. The comment's reasoning is also off: `ftp_points` is built from the whole FTPdetectinfo file with no shower filtering, so a shower filter is not the reason for a miss — FFs with no detection at all are. Building `ftp_points` as `ff_name -> [meteor_meas, ...]` and OR-ing the per-meteor line masks together removes the magic `10`, the `i * 1.0` float-key coercion, and the try/except control flow all at once.

**Design question**

Masking `avepixel` as well as `maxpixel` throws away the sky background, not just the planes. Planes and satellites are carried into the stack by `max_deavg = maxpixel - avepixel`; masking that alone suppresses them. Masking `avepixel` too means `avg_stack_count` only increments inside the meteor strips, so `avg_stack_sum / avg_stack_count` is zero everywhere no strip landed — the star field survives only as narrow bands, and the auto-crop shrinks with it (which is why the "after" image is 1065x1116 vs 1165x1174). Masking only `max_deavg` gives the same suppression with the full star background intact. If the strips-only look is deliberate, that's fine, but it should be said in the `--mask-meteors` help text.

**Smaller things** are in the inline comments: reusing the existing `findFTPdetectinfoFile`/`validDefaultFTPdetectinfo` helpers, the hard-coded 150 px dilation, camelCase naming, and the missing docstring entry.

For the record, a few things I checked that are correct as written: the `[2]`/`[3]` centroid indices match `meteor_meas`' `[calib_status, frame_n, x, y, ...]` layout and the FTPdetectinfo `Col Row` header ordering, so the `cv2.line((x, y))` call is right; the `(ff_name, meteor_No)` float-keyed dict mirrors the existing `shouldInclude`/`associations` convention; and pickling `ftp_points` through `func_extra_args` is cheap at realistic detection counts.

Utils/TrackStack.py
tammojan: Nice suggestion, done.

Utils/TrackStack.py
tammojan: Done

Utils/TrackStack.py
tammojan: Done

Utils/TrackStack.py
tammojan: Fixed

Utils/TrackStack.py
tammojan: Nice, this fixes the background subtraction for bright meteors (one of which is visible in the example I posted).

Utils/TrackStack.py
tammojan: Ouch, fixed.

tammojan (commented):
Thanks Denis and AI for this thorough review. I addressed most of it.

dvida:
Awesome - once you're fully done, could you regnerate and post the example above? I'd love to see the new plot :)

Utils/TrackStack.py
tammojan: Nice suggestion, applied

tammojan:
I added most changes. Somehow the background subtraction doesn't work well with bright meteors. I don't see why that is, since `avepixel_median` is computed on the non-masked image (`ff.avepixel`).

With the current state of the code, I get this:
afterchanges

PR #967, opened 2026-08-18
AutoPlatepar/AstrometryNet: optional position_hint for star-starved solves

GlassOnTin:
`autoFitPlatepar` (via `astrometryNetSolve`/`astrometryNetSolveLocal`) always runs a blind all-sky solve; `position_hint` is hardcoded to `None`. Narrow-FOV or faint cameras extract too few stars for a blind solve, so it fails with "No solution found".

This adds an optional `position_hint=(ra_deg, dec_deg, radius_deg)` threaded through to the astrometry.net `Solver`. It defaults to `None` everywhere, so existing callers are unchanged.

Tested on an IMX296 mono + 4 mm f/0.95 Pi 5 station (~75° field, ~25-40 extractable stars per frame). The blind solve only locks when the field happens to reach ~100 stars; with a zenith hint (RA = LST, Dec = latitude) the same ~28-star field solves cleanly.

A follow-up could derive the hint automatically from a `platepar_template` with a known pointing - kept out of this PR to stay minimal.

### DIFF
--- a/RMS/Astrometry/AstrometryNet.py
+++ b/RMS/Astrometry/AstrometryNet.py
@@ -32,7 +32,7 @@ if os.environ.get('RMS_DISABLE_LOCAL_ASTROMETRY', '').lower() in ('1', 'true', '
 
 def astrometryNetSolveLocal(ff_file_path=None, img=None, mask=None, x_data=None, y_data=None,
                             fov_w_range=None, fov_w_hint=None, max_stars=100, verbose=False, x_center=None, y_center=None,
-                            lat=None, lon=None, jd=None, input_intensities=None):
+                            lat=None, lon=None, jd=None, input_intensities=None, position_hint=None):
     """ Find an astrometric solution of X, Y image coordinates of stars detected on an image using the
         local installation of astrometry.net.
 
@@ -52,6 +52,9 @@ def astrometryNetSolveLocal(ff_file_path=None, img=None, mask=None, x_data=None,
         lon: [float] Station longitude in degrees. Required for iterative matching.
         jd: [float] Julian date. Required for iterative matching.
         input_intensities: [ndarray] Star intensities for brightness-based matching. Optional.
+        position_hint: [tuple] Optional (ra_deg, dec_deg, radius_deg). When given, astrometry.net is
+            told roughly where the field is, collapsing the blind all-sky search to that patch --
+            which lets narrow-FOV / faint cameras solve with far fewer stars. None = blind (default).
 
     Returns:
         [tuple] A tuple containing the following elements:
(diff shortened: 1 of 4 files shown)


dvida:
Updated this PR directly with the review fixes:

- Rebased the astrometry feature onto current prerelease and preserved the original author attribution.
- Dropped the obsolete white-ratio Detection commit, which is already merged through PR #970 with its follow-up fixes and tests.
- Preserved the existing autoFitPlatepar positional API by appending position_hint after the established parameters.
- Kept position_hint on the tight-to-wide AutoPlatepar retry.
- Forwarded the hint through local solving, remote coordinate/image solving, and local-error remote fallbacks.
- Added remote Astrometry.net translation to center_ra, center_dec, and radius.
- Added focused coverage for current and legacy local solver APIs, blind defaults, all fallback paths, remote upload fields, API order, and the wide retry.

Validation:

- python3 -m pytest -q Tests/TestAstrometryNet.py Tests/TestDetection.py — 10 passed
- compileall for all changed Python modules and the new test — passed
- git diff --check — passed

The PR is now mergeable and its diff is limited to the astrometry implementation and tests.

dvida:
@GlassOnTin could you please test this new version on your end?

GlassOnTin:
Tested on my station (RPi 5, IMX296 global-shutter + 4 mm f/0.95, ~64×50° FOV, local astrometry.net series‑4100).

✅ `Tests/TestAstrometryNet.py` — 7/7 pass.

⚠️ One wide‑FOV regression from the hint propagation, though. On a real 68‑star frame:
- **pre‑fix (b2a8da43):** tight‑with‑hint fails → wide‑FOV fallback runs blind → **solves** (44 matched).
- **this PR (1d54f4ee):** tight‑with‑hint fails → wide‑FOV fallback now runs *with the hint* → **no solution**.

Reproducible. Cause is hint‑radius vs field size: my hint was `radius=15°`, but the FOV half‑diagonal is ~35°, so the local solver restricts the catalog to 15° around the centre and drops the outer stars → quad match fails. The same PR code with `radius=45°` (> field radius) solves fine (identical result to the blind fallback).

So forwarding the hint into the wide‑FOV all‑sky fallback can hurt wide‑field cameras when the caller's radius is smaller than the field. Options: keep the wide‑FOV fallback blind (it's the last‑resort recovery path), or clamp the hint radius up to ≥ the field angular radius when `wide_fov_search=True`. Happy to test whichever you prefer.

GlassOnTin:
Pushed a fix for that regression in b83ed7f7 — keeps the hint flowing everywhere (no revert of the propagation), just makes it safe on wide fields.

The local solver restricts the reference catalog to within `radius_hint` of the hint centre, so when that radius is smaller than the image's own angular field it drops the outer stars and the quad match fails. The commit clamps `radius_hint` up to the field's circum-radius (half-diagonal, +10% margin) before building the `PositionHint`, only when a FOV estimate is available (always, from `autoFitPlatepar`).

On my station the previously-failing 15° hint now solves — clamped to ~49° — with the identical result to the blind fallback, and `Tests/TestAstrometryNet.py` stays 7/7. Narrow-FOV cameras are unaffected (their hint radius is already ≥ their small field), and the blind path is unchanged. Margin's arbitrary (10%) — tweak as you like.

Issue #969, opened 2026-08-19
Detection threshold (k1·stdpixel + j1) silently fails on video paths that smooth per-pixel noise


GlassOnTin:
### Summary

The `stdpixel`-based detection threshold assumes per-pixel *temporal* noise survives into the FF stack. Video paths that smooth it — software H.264 re-encoding, and some IP-camera encoders — break that assumption: `stdpixel` drops well below the residual `maxpixel` noise floor, the `k1·stdpixel + j1` threshold falls under that floor, the white-ratio guard in `checkWhiteRatio()` rejects every time window, and detection silently yields zero meteors all night.

### Evidence (Pi 5 / IMX296, re-encoded RTSP stream, clear night)

- median `stdpixel` ≈ 3, but median `maxpixel − avepixel` ≈ 26 ADU
- at default `k1=3.5, j1=12`: **~68 %** of the frame passes threshold → the 5 % white-ratio guard aborts **every** time window → 0 KHT lines → 0 meteors
- a **saturated, unambiguous meteor** (≈7.2°, 0.80 s, ≈9 °/s, perfectly linear) logged `detected meteors: 0` live; the *same* FF reprocessed with `j1=30` (whole-frame white → ~2.2 %) detects it correctly
- **control:** on cloudy/twilight frames the background statistics rise, the threshold rises with them, and detection "works" — which is exactly why the failure looks intermittent and goes unnoticed

The failure is also completely silent at INFO level (see companion PR to surface it).

### Why it matters

Pi + CSI-camera stations that bridge to RTSP (no hardware H.264 encoder on the Pi 5) are a growing population, and any such re-encoded path is susceptible. An affected station uploads nightly but detects nothing on clear nights, with no error to indicate why.

### Proposed directions (for discussion)

1. **Adaptive threshold** — when the white ratio exceeds the limit, raise `j1` (or scale the threshold) and retry rather than aborting, so detection recovers without per-station tuning.
2. **Warn + document** — surface the abort (companion PR) and document recommended `k1`/`j1` for encoder-fed stations.
3. **Robust noise estimate** — derive the threshold from a noise measure less sensitive to the encoder's temporal smoothing.

Happy to contribute the adaptive-threshold change (1) if there's appetite for it.

GlassOnTin:
[...]

PR #970, opened 2026-08-19
Warn when the white-ratio check silently skips meteor detection

GlassOnTin:
### Problem

`checkWhiteRatio()` aborts all line-finding for an image when too many pixels pass the detection threshold, but logs this only at `debug` level. The line does reach the nightly log file (the file handler defaults to DEBUG), but never the console: there the operator sees nothing but `detected meteors: 0`, indistinguishable from a quiet sky, and nothing prompts them to dig through the debug log.

### When it bites

On stations whose video path smooths per-pixel *temporal* noise — notably software re-encoded RTSP streams (Pi + CSI-camera bridges, some IP-camera encoders) — `stdpixel` collapses while the `maxpixel` noise floor remains. The `k1·stdpixel + j1` threshold then lands *below* the noise floor, a large fraction of every frame passes, and the white-ratio guard rejects **every image of a clear night**. The station is silently blind to meteors. (Cloud/twilight raises the background statistics and masks the problem, so it looks intermittent.)

Full analysis and evidence in #969.

### This PR

Promotes the first rejection (and every 100th thereafter, to avoid log spam) to a `warning` that reports the measured white ratio and hints that `k1`/`j1` may be too low for the camera. **No behavioural change to detection** — it purely makes an existing silent failure visible. The deeper fix (adaptive thresholding) is proposed in #969.

### Testing

Verified in an isolated harness that the warning fires on the 1st and 100th rejection only, rejections still return `False`, and sub-threshold images still pass. Found in production on a Pi 5 / IMX296 station feeding via a re-encoded RTSP stream.

### DIFF
--- a/RMS/Detection.py
+++ b/RMS/Detection.py
@@ -364,8 +364,16 @@ def merge3DLines(line_list, vect_angle_thresh, last_count=0):
 
 
 
+# Count of images rejected by the white ratio check, used only to rate limit the warning below
+# instead of firing it on every rejected image. Detection runs one image per worker process, so
+# each worker keeps its own count - fine for throttling, but not a run-wide total, which is why
+# the count is not printed in the message
+white_ratio_reject_count = 0
+
 def checkWhiteRatio(img_thres, ff, max_white_ratio):
     """ Checks if there are too many threshold passers on an image. """
+
+    global white_ratio_reject_count
 
     # Check if the image is too "white" and any further processing makes no sense
     # Compute the radio between the number of threshold passers and all pixels
@@ -375,8 +383,24 @@ def checkWhiteRatio(img_thres, ff, max_white_ratio):
 
     if white_ratio > max_white_ratio:
 
-        log.debug(("Too many threshold passers! White ratio is {:.2f}, which is higher than the "\
-            "max_white_ratio threshold: {:.2f}").format(white_ratio, max_white_ratio))
+        white_ratio_reject_count += 1
+
+        msg = ("Too many threshold passers! White ratio is {:.2f}, which is higher than the "\
+            "max_white_ratio threshold: {:.2f}, so detection is skipped for this image. "\
+            "This is expected on bright or cloudy images, but "\
+            "if it happens on every image of a clear night the detection k1/j1 thresholds are "\
+            "too low for this camera (e.g. video paths that smooth per-pixel temporal noise, "\
+            "such as software re-encoded streams, can push the stdpixel-based threshold below "\
+            "the maxpixel noise floor).").format(white_ratio, max_white_ratio)
+
+        # Make the first rejection (and every 100th) visible at warning level - a station whose
+        # every image is rejected is effectively blind to meteors, and outside of debug logging
+        # this used to be completely silent
+        if (white_ratio_reject_count == 1) or (white_ratio_reject_count%100 == 0):
+            log.warning(msg)
+
+        else:
+            log.debug(msg)
 
         return False
 
--- a/RMS/Detection.py
+++ b/RMS/Detection.py
@@ -364,16 +364,16 @@ def merge3DLines(line_list, vect_angle_thresh, last_count=0):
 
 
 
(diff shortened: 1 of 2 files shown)


Cybis320:
The failure mode in #969 is real and well diagnosed, and I think promoting this is the right call. I was initially worried the warning would be noisy on ordinary nights, so I measured it: I grepped 40 nightly logs from ~20 stations (US, NZ, DE, UK, CZ, BE, NL, HR, MY, CA), across winter and summer, ~90,000 FF files in total.

About 7 rejections in ~90,000 images. Most nights are zero; the worst is 2. 

That's a good result for this PR — the guard essentially never trips on a healthy station, so the warning stays silent in normal operation and fires precisely on the pathology you found. It also matches your control observation in #969: clouds raise avepixel and stdpixel together, the threshold rises with the signal, and the guard doesn't trip.

It also means the rate limiter is doing real work in the only case where it matters. A healthy station never reaches the every-100th branch; an affected one produces ~4300 rejections a night, which without the throttle would be 4300 console warnings. Worth keeping.

**Two notes.**

**The count in the message isn't the run's count**

white_ratio_reject_count is a module-level global, but detection runs one FF per worker across config.num_cores processes (DetectStarsAndMeteors.py:413, with _workerFunc a long-lived loop at QueuedPool.py:380-413), so each worker gets its own copy.

The throttle survives this fine — with 4300 rejections across 4 workers each sees ~1075 and warns at 1, 100 ... 1000, giving 44 warnings against the 44 a single process would emit. But the printed number does not: "{:d} rejected so far this run" reports the worker's private count, so a message reading "100 rejected so far this run" on a 4-core station means ~400, and the counts from different workers interleave in one log stream as 100, 100, 200, 100 ...

Simplest fix is to drop that clause from the string and keep the counter purely as the throttle, which is the job it does correctly. The white ratio and the diagnostic hint are the parts an operator acts on anyway.

Minor, related: the two call sites feed different units into the same total. Detection.py:423 is a whole-FF rejection, Detection.py:466 a per-time-window rejection on non-FF input.

**Small correction to the description**

The abort isn't invisible at INFO — it's invisible on the console. checkWhiteRatio uses log.debug on the plain rmslogger (Detection.py:65), so the file handler takes the keyword default file_level=logging.DEBUG (Logger.py:452, applied at Logger.py:817). The message is in the nightly log file on every stock station today — that's how I measured the rates above.

Doesn't change the case for the PR, since nobody reads a Pi's debug log unprompted. Just worth stating accurately in the commit message.

GlassOnTin:
Thank you for measuring it rather than guessing. 7 rejections in ~90,000 files across 20 stations settles the noise question far better than my isolated harness could, and it is good to have the healthy-station base rate on the record here.

Both notes are applied, head is now 59b1da8:

The "{:d} rejected so far this run" clause is gone from the message. You are right that the module global is per worker process, so the printed number was a private count masquerading as a run total. The counter now serves only as the rate limiter, which as you showed still does its job under multiprocessing (about the same number of warnings in total, just distributed across workers), and the comment above it now states the per-worker semantics so the next reader does not reintroduce the count. Dropping the printed number also disposes of the mixed units between the whole-FF call site and the per-window one, since the total no longer surfaces anywhere.

The commit message and the PR description now say the abort was invisible on the console rather than invisible outright, with the debug line reaching the nightly log file via the file handler's DEBUG default. That distinction is also how you were able to measure the base rate at all, so it is worth stating correctly.

Cybis320:
Looks good. Thank you.

dvida:
Implemented the review follow-up in commit 1e91048f:

- increased the displayed white-ratio precision to four decimal places, so near-threshold values remain distinguishable;
- made the process-local counter private and corrected its inline documentation to describe long-lived workers accurately;
- matched the surrounding top-level spacing;
- added focused regression coverage for accepted/rejected return values, warning counts 1/100/200, debug-level intervals, counter isolation, and near-threshold message precision.

Validation:
- 11 focused detection and multiprocessing tests passed;
- RMS/Detection.py and Tests/TestDetection.py compile successfully;
- git diff --check passed.

Issue #974, opened 2026-08-23
Cython build fails on Cython 3.3.0


g7gpr:
When RMS_Installer.sh was run on a newly built Debian machine following error was seen.

```
Obtaining file:///home/au000w/source/RMS
  Checking if build backend supports build_editable ... done
  Preparing editable metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing editable metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [48 lines of output]
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
              for x in range(x_size):
      
                  max_val = compressed[0, y, x]
      
                  # Compute the threshold limit
                  avg_std = int(float(compressed[2, y, x]) + k1*float(compressed[3, y, x])) + j1
                                                                                            ^
      ------------------------------------------------------------
      RMS/Routines/Grouping3Dcy.pyx:481:86: Cannot assign type 'double' to 'unsigned int'
```

This is the first machine I have deployed on Cython version 3.3.0

```
(vRMS) au000w@pioneer-alpha:~/source/RMS$ cython --version
Cython version 3.3.0
```

Although I'm unable to find where this change in behaviour is documented, I believe it is a restriction on type narrowing that has been introduced in 3.3.x. 

I'm testing a fix, and will submit a PR in the next 24 hours if it is successful and backward compatible.

```            
# Compute the threshold limit
# This line no longer supported in Cython 3.3.0
# avg_std = int(float(compressed[2, y, x]) + k1*float(compressed[3, y, x])) + j1

avg_std = <unsigned int> (<double> compressed[2, y, x] + k1 * <double> compressed[3, y, x] + <double> j1)
```

g7gpr:
Tested on several machines, closing, with thanks.

PR #975, opened 2026-08-23
Cython 3.3.0 compatibility

g7gpr:
This PR is in response to issue #974, and avoids Cython build failing.

### DIFF
--- a/RMS/Routines/Grouping3Dcy.pyx
+++ b/RMS/Routines/Grouping3Dcy.pyx
@@ -489,7 +489,7 @@ def thresholdAndSubsample(np.ndarray[UINT8_TYPE_t, ndim=3] frames, \
 
             # Keep the original truncation order while making both narrowing conversions explicit.
             avg_std = <unsigned int> (
-                <int> (<double> compressed[2, y, x] + k1 * <double> compressed[3, y, x])
+                <int> (<double> compressed[2, y, x] + k1*<double> compressed[3, y, x])
                 + <double> j1
             )
 
--- a/RMS/Routines/Grouping3Dcy.pyx
+++ b/RMS/Routines/Grouping3Dcy.pyx
@@ -488,7 +488,10 @@ def thresholdAndSubsample(np.ndarray[UINT8_TYPE_t, ndim=3] frames, \
             max_val = compressed[0, y, x]
 
             # Compute the threshold limit
-            avg_std = int(float(compressed[2, y, x]) + k1*float(compressed[3, y, x])) + j1
+            # This line no longer supported in Cython 3.3.0
+            # avg_std = int(float(compressed[2, y, x]) + k1*float(compressed[3, y, x])) + j1
+
+            avg_std = <unsigned int> (<double> compressed[2, y, x] + k1 * <double> compressed[3, y, x] + <double> j1)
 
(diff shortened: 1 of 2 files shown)


Cybis320 (approved):

dvida:
While reviewing this change, I found that the original patch moved j1 across a truncation boundary. With average=10, stddev=1, k1=0.6, and j1=0.6, the old implementation produced a threshold of 10 while the initial PR produced 11, causing a valid point at level 10 to be dropped.

I pushed follow-up commit 0b6a094, which keeps the original truncation order while making the narrowing conversions explicit for Cython 3.3.0. I also added regression coverage for fractional and integer j1 values and the 255 clamp.

Verification completed:
- Full extension build with Cython 3.3.0
- Focused regression tests: 3 passed
- git diff --check: clean

dvida:
Could you please test again with my latest changes?

Cybis320:
> Could you please test again with my latest changes?

The rewrite looks right to me, and the regression test came with it is a great idea. I mutation-tested the new test to see what it actually catches:

| Deliberately broken version | Caught? |
|---|---|
| Single-cast form (adds `j1` before truncating) | ✅ case 1 fails |
| `if avg_std > 255` clamp removed | ✅ case 3 fails |
| `avg_std = 0` (threshold computation gutted) | ❌ **all 3 pass** |

So the test catches a threshold that comes out too *high*, but not one that comes out too *low* — all three cases assert `== 1`, so nothing pins the rejection side.

One extra case closes it. I ran it against the PR head (returns 0, correct) and against the gutted version (returns 1, caught):

```diff
 @pytest.mark.parametrize(
-    "max_value, average, stddev, k1, j1",
+    "max_value, average, stddev, k1, j1, expected",
     [
         # Truncate average + k1*stddev before adding the fractional offset.
-        (10, 10, 1, 0.6, 0.6),
+        (10, 10, 1, 0.6, 0.6, 1),
         # Preserve the normal integer-offset behavior.
-        (15, 10, 1, 0.6, 5.0),
+        (15, 10, 1, 0.6, 5.0, 1),
         # Clamp thresholds above the uint8 range to 255.
-        (255, 250, 10, 1.0, 0.0),
+        (255, 250, 10, 1.0, 0.0, 1),
+        # Reject a pixel under the threshold: avg_std = 10, max_val = 5.
+        (5, 10, 1, 0.6, 0.0, 0),
     ],
 )
-def testThresholdAndSubsampleThresholdConversion(max_value, average, stddev, k1, j1):
-    assert _thresholdCount(max_value, average, stddev, k1, j1) == 1
+def testThresholdAndSubsampleThresholdConversion(max_value, average, stddev, k1, j1, expected):
+    assert _thresholdCount(max_value, average, stddev, k1, j1) == expected
```

Minor, while you're in there: case 1 is the only one that pins the truncation order, and it needs the fractional `j1=0.6` to do it. Both `ConfigReader` paths parse `j1` with `getint` (`ConfigReader.py:1436`, `:1532`), so production never
produces a fractional `j1` — worth a comment on that line so nobody later "tidies" it to an integer and silently removes the only test of the thing this PR is about.

The fix itself is good. Confirmed it cythonizes clean on both 3.3.0 and 3.2.5, and that Grouping3Dcy was the only one of the eight `.pyx` files 3.3.0 rejected.

PR #976, opened 2026-08-24
Add star candidate and PSF-fit visualization

dvida:
## Summary
- add a --show-candidates option to RMS.ExtractStars
- display raw local-maximum candidates before PSF fitting, including over-limit sets
- after fitting, display raw candidates as red circles and accepted PSF fits as green markers
- auto-level the avepixel background from the 1st to 99.99th percentile with a flat-image fallback
- show candidate/fitted counts and a legend in the comparison plot
- process FF files sequentially while interactive candidate windows are enabled
- keep default parallel extraction unchanged
- infer integer image bit depth safely and default floating display images to 8-bit unless specified

## Tests
- pytest -q Tests/TestExtractStars.py (8 passed)
- real-data run on FF_US002B_20260427_084517_340_0752128.fits (256 candidates, 253 fitted)
- python -m compileall -q RMS/ExtractStars.py Tests/TestExtractStars.py
- python -m RMS.ExtractStars --help
- git diff --check

### DIFF
--- a/RMS/ExtractStars.py
+++ b/RMS/ExtractStars.py
@@ -56,7 +56,7 @@ log = getLogger("rmslogger")
 def extractStars(img, img_median=None, mask=None, gamma=1.0, max_star_candidates=1000, border=10,
                  neighborhood_size=10, intensity_threshold=18,
                  segment_radius=4, roundness_threshold=0.5, max_feature_ratio=0.8, bit_depth=8,
-                 extra_info=None):
+                 extra_info=None, show_candidates=False):
     """ Extracts stars on a given image by searching for local maxima and applying PSF fit for star 
         confirmation.
 
@@ -79,6 +79,7 @@ def extractStars(img, img_median=None, mask=None, gamma=1.0, max_star_candidates
             (hot pixels are narrow, while stars are round).
         max_feature_ratio: [float] Maximum ratio between 2 sigma of the star and the image segment area.
         bit_depth: [int] Bit depth of the image. 8 bits by default.
+        show_candidates: [bool] Show the raw star candidates before PSF fitting. False by default.
     
     Return:
         x2, y2, background, intensity, fwhm: [list of ndarrays]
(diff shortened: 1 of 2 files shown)


Cybis320:
Just a heads up that this PR collides with the yet-to-be-merged #918 (candidate subsampling), and depending on the motivation behind it, might loose some of its utility once/if #918 is merged.

## The conflict

I test-merged this into `improve-star-detection`. One conflict in `RMS/ExtractStars.py`, the overflow block: this PR reorders it to plot before `return False`, while #918 deletes the `return False` outright and replaces it with the stratified top-hat subsampler.

Two of the five new tests assert the removed behavior and will fail: `testDefaultOverLimitRejectionDoesNotPlot` and `testOverLimitCandidatesAreShownThenRejected`.

## In #918 `max_stars` stops being a threshold

Both call sites pass `max_star_candidates=config.max_stars`. #918 doesn't change the number, it changes what it is:

- Today: a panic threshold. Cross it, the whole FF yields zero stars. The only sane tuning is "high enough to never trip it" — hence 800 in the template.
- After #918: a sampling budget. Cross it, you keep the best N spatially stratified and the PSF fit vets them. It binds routinely and benignly, so the right value is *lower* — 400 on my stations.

So "over-limit" stops being an error condition and becomes the normal operating regime on any noisy camera. The PR body's "including over-limit sets" is the half that goes away.


The part I'd keep regardless is the `plotStars` rework — ~30 lines, correct bit-depth handling, `fig`/`ax` instead of pyplot global state, image-or-FF input. Strictly better than the commented-out call it replaces, and reusable by the overlay above. One bug in it: `bit_depth = 8*img.dtype.itemsize` infers 32 for a float32 image → `max_level = 2**32 - 1` → black plot. `extractStars` always passes it explicitly, so only future callers are exposed.

Minor: this adds `Tests/TestExtractStars.py`, I have `Tests/test_ExtractStars.py` coming on the adaptive-gate branch. Two files for one module differing only in case. Happy to fold mine into yours or rename either way.

dvida:
Thanks for the detailed heads-up and for catching the float32 inference issue. I fixed that in 3a72121b: integer images now infer their width with np.iinfo, while floating-point images default to the historical 8-bit display scale unless the caller supplies the source bit depth. I also added a float32 regression test.

For the possible future integration with #918, I suggest keeping the candidate overlay before subsampling so it continues to show the complete raw local-maximum set. We could then optionally distinguish the spatially retained sample with a second colour and show raw/selected counts in the title. The over-limit tests would become sampling-budget tests: the plot should receive all raw candidates, while PSF fitting should receive at most max_star_candidates. The ExtractStars tests can also be consolidated into one canonically named file when that branch is rebased.

For this PR I am leaving the current prerelease overflow semantics intact, since #918 is not merged.

PR #977, opened 2026-08-25
Fix star FWHM convention: collapse to standard 2.355*sigma for circular stars

Cybis320:
## Summary

The star extractor computes the reported FWHM as the **quadrature sum** of the two fitted axis sigmas:

```python
sigma_fitted = np.sqrt(sigma_x_fitted**2 + sigma_y_fitted**2)
fwhm = 2.355*sigma_fitted
```

For a circular star (σx = σy = σ) this gives 2.355·σ·√2 — **41% larger than the standard FWHM** (2.355·σ), the width between the half-maximum points of the profile. The value therefore does not collapse to the textbook definition for round stars. Git archaeology shows the origin: `b1c89067` ("finished flux engine", Jan 2021) relabeled a pre-existing quadrature-combined "Gaussian stddev" diagnostic as FWHM by multiplying it by 2.355, a conversion only valid for a 1-D sigma.

### Consequences of the inflated convention

- Every CALSTARS FWHM value written since Jan 2021 is inflated by √2.
- SkyFit2's manual moment-based measurement (`sqrt((mx+my)/(2*flux))`) already uses the standard convention, so hand-picked and auto-detected stars carried **inconsistent FWHMs in the same paired-star list** (blend filtering, match radii).
- The flux angular-velocity magnitude loss (Vida et al. 2022, eq. 38, which explicitly defines FWHM as the standard PSF full width at half maximum) received the inflated value, making the meteor limiting magnitude **0.376 mag too faint** and biasing computed fluxes **low by r^0.376** (×1.30 for s=1.75 up to ×1.41 for s=1.99).
- Sampling analyses based on CALSTARS read ~2.5 px FWHM where the true value is ~1.8 px, masking that typical 720p setups are marginally undersampled.

## Changes

1. **`RMS/ExtractStars.py`**: FWHM is now `2.355*sqrt((sigma_x**2 + sigma_y**2)/2)` (RMS mean of the axis sigmas) — collapses to 2.355·σ for circular stars and agrees with SkyFit2's moment-based measurement.
2. **Behavior-preserving rescales** (effective pixel radii unchanged): `DEFAULT_BLEND_FWHM_MULT` 2.0 → 2.8; SkyFit2 dynamic match radius 1.5×FWHM → 2.1×FWHM; segment-radius auto-tuner margin 1.5 → 2.1; placeholder FWHM fallbacks 2.5 → 1.8.
3. **CALSTARS format versioning** (`RMS/Formats/CALSTARS.py`): files now carry a `Version = 2` header line. `readCALSTARS()` detects the version and **normalizes legacy (version 1) files by ÷√2 on read**, so all consumers see the standard convention regardless of file age. The −1.0 "no FWHM" sentinel is preserved, and the added header line is ignored by the old parser (fails the numeric parse and is skipped).
4. **Flux input normalization** (`Utils/Flux.py`): the cached `flux_sensor_characterization.json` is versioned the same way (legacy caches normalized ÷√2 on load), and `default_fwhm` is rescaled 3 → 2.1 to the standard convention.

## Validation

- Synthetic circular Gaussians (σ = 1.2): fitted FWHM 2.823–2.832 vs theoretical 2.826 (old code: 4.00).
- CALSTARS round-trip: version 2 files read back unchanged; a legacy file's 2.30 reads as 1.626 (= 2.30/√2); sentinel −1.0 untouched.
- End-to-end on archived GMN nights (US005E 2025-07-29 CAP/SDA, 2025-08-12 PER): the meteor LM shifts by exactly −0.376 mag in every time bin, and flux@+6.5ᴹ scales by exactly r^0.376 per shower (CAP ×1.412, SDA ×1.368, PER ×1.298), matching the analytic prediction. A pristine-baseline rerun of the archived server flux tables reproduces them to ≤0.04% before the fix, confirming the harness.

## Science impact (intentional)

Computed fluxes/ZHRs increase by r^0.376 (≈ +30–41%) because the angular-velocity loss term now receives the quantity its derivation calls for. This interacts with the angular-velocity correction fix submitted separately — for typical far-radiant pointing the two errors partially cancel, so they should be evaluated (and ideally merged) together.

https://claude.ai/code/session_01B6s6Z27vWeL1Ume1t9AAfz

### DIFF
--- a/RMS/Astrometry/AutoPlatepar.py
+++ b/RMS/Astrometry/AutoPlatepar.py
@@ -362,7 +362,7 @@ def autoFitPlatepar(dir_path, config, catalog_stars, platepar_template=None,
 
     Keyword arguments (filtering parameters):
         photometric_sigma: [float] Sigma threshold for photometric outlier removal (default: 2.5)
-        fwhm_mult: [float] Multiplier of FWHM for blend detection radius (default: 2.0)
+        fwhm_mult: [float] Multiplier of FWHM for blend detection radius (default: 2.8)
 
         wide_fov_search: [bool] If True, use a wide FOV search range (2° to 200°) instead of
                          the config-based range. Used as fallback when the tight search fails.
@@ -624,7 +624,7 @@ def autoFitPlatepar(dir_path, config, catalog_stars, platepar_template=None,
             sky_obj = CatalogStar(cat_ra, cat_dec, cat_mag)
 
             # Find closest detected star to get FWHM, SNR, saturation
-            fwhm, snr, saturated = 2.5, 1.0, False
+            fwhm, snr, saturated = 1.8, 1.0, False
             if len(x_data) > 0:
                 distances = np.sqrt((x_data - img_x)**2 + (y_data - img_y)**2)
                 closest_idx = np.argmin(distances)
--- a/RMS/Astrometry/StarFilters.py
+++ b/RMS/Astrometry/StarFilters.py
@@ -19,7 +19,10 @@ from RMS.Astrometry.ApplyAstrometry import extinctionCorrectionTrueToApparent, r
 
 # Default filtering parameters
 DEFAULT_PHOTOMETRIC_SIGMA = 2.5
-DEFAULT_BLEND_FWHM_MULT = 2.0  # Multiplier of FWHM for blending detection radius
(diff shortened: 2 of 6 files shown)


Cybis320:
The angular-velocity correction fix referenced above is #978. For typical far-radiant pointing the two errors partially cancel, so the flux impact should be evaluated with both PRs together.

PR #979, opened 2026-08-28
Sprite detector feeding the spritemap server

9helix:
Live detection of sprites and elves (TLEs) on every FF block during capture. Confirmed detections are sent to the spritemap server (`9helix/spritemap-data-pipeline`), which triangulates them across stations.

Off by default (`detect_sprites: false`). With it off, nothing is imported or started.

## How it works

1. The Compressor hands each FF name to the sprite process through a bounded queue. It uses `put_nowait`, so capture never waits on sprite detection. If the queue is full the FF is skipped and a warning is logged at most once a minute.
2. The YOLOv5 TFLite model runs on the maxpixel image, with the station mask if `sprite_use_mask` is set.
3. An artifact filter rejects boxes whose light is not concentrated in a few frames, and gives the frame of peak light.
4. A burst filter rejects a detection if more than 3 FF blocks have detections within 60 s (lightning, headlights, clouds). It is driven only by FF times, so the result does not depend on machine speed. As a result, confirmation lags the event by about a minute.
5. Each confirmed detection gets RA/Dec and azimuth/altitude from the station platepar. This uses the time of the peak frame and the FF resolution. Refraction is always taken out: explicitly, or inside the fitted distortion for platepars fitted with refraction off, so every platepar is used.
6. Products are written into the night directory, so they are archived like the other products:
   - `<night>_sprites.csv`: every candidate, with a status column.
   - `<night>_sprites.json`: confirmed detections with platepar provenance.
   - Up to `sprite_max_images` pairs of `<FF>_marked.jpg` and `<FF>_unmarked.jpg`.
7. If `sprite_upload_url` is set, the detections are posted to the server. FF files and images are uploaded only when the server asks for them, which it does once a second station has seen the same event.

## Server protocol

- **Signing.** Requests are signed with the station's existing SSH key (the one used for the nightly upload). The signature covers the method, path, station, time and a SHA-256 of the body, so a captured request cannot be replayed elsewhere.
  - `Tests/data/spritemap_canonical_vectors.json` is committed in both repositories and tested on both sides, so the two copies of the signing code cannot drift apart.
- **Platepar.** Each payload carries the platepar, trimmed to the mapping fields (1-2 KB, no star list), so the server can recompute the directions later.
  - The server stores each distinct platepar once, under the SHA-256 of its canonical JSON. The station's local record has the same key.
- **Retries.** A payload is written to disk once and the same bytes are posted on every retry, so the server recognises a repeat as a duplicate.
- **Queue.** Pending uploads are kept in `data_dir/SPRITE_UPLOADS.json` across restarts, with exponential backoff.
  - A 401 (key not registered yet) pauses uploads for an hour instead of dropping them.
- **Safety.** HTTPS only, and redirects are refused. The shutdown drain holds its own reboot lock (`reboot_lock_file + ".sprite"`), and `StartCapture` now waits for every `lock*` file.

## The model

The model (35 MB) is **not in the repository**; the history of this branch was rewritten to remove it. On first use, the station downloads it over HTTPS from https://globalmeteornetwork.org/projects/sprite_detector/sprite_detector.tflite (option `sprite_model_base_url`). No key is needed.
- The download is streamed, capped at 200 MB and checked against the SHA-256 in `RMS/Sprite/DownloadModel.py` before it replaces anything.
- Without a model, sprite detection is disabled for the night and capture is unaffected.

It can also be fetched by hand with `python -m RMS.Sprite.DownloadModel`.

## Main fixes over the previous version of this PR

- **The process never exited.** The loop only ended after 150 FF files with detections. On a normal night `stop()` timed out and the process was killed, losing the last detections.
- **Live astrometry never ran.** The platepar was looked for in `dirname(config.config_file_path)`, one level above the config directory.
- **Event time was wrong.**
  - The FF milliseconds were dropped (up to 1 s early).
  - Astrometry used the FF start instead of the detection time (up to 10 s).
  - Refraction was applied, although the server needs true directions.
- **Uploads failed on the server:**
  - base64 images inside the JSON exceeded its 1 MiB limit;
  - image names still contained `.fits`;
  - the signature covered only the body;
  - `sent_at` changed on every retry.
- **Crashes:**
  - the image writer returned `None` on error, which the caller unpacked into two values;
  - an HTML error page was parsed as JSON inside the error handler;
  - an exception in one FF stopped the process.
- **Artifact filter.** It reconstructed all 256 frames per detection. It is now one weighted histogram over the box, with the same result.
- **Options that did nothing:** the station mask was never applied, and `--confidence` was ignored.
- **No `.config` entries.** The sprite options were missing from `.config`, so the config audit flagged seven options on every station. All of them are now documented there.

## Layout

All sprite code is in the `RMS.Sprite` package:

| Module | Contents |
|---|---|
| `RMS/Sprite/Detection.py` | model, mask, artifact filter, frame times |
| `RMS/Sprite/Astrometry.py` | platepar checks, sky coordinates |
| `RMS/Sprite/Filter.py` | burst filter (replaces `Utils/FalsePositiveFilter.py`, which needed `dataclasses`) |
| `RMS/Sprite/Products.py` | CSV, JSON, images, upload payload |
| `RMS/Sprite/Upload.py` | signed client, persistent queue, upload thread |
| `RMS/Sprite/Detector.py` | capture-time process, offline run, command line |
| `RMS/Sprite/DownloadModel.py` | model download |

The `DeleteOldObservations.py` changes and the `SpriteData` directory tree are gone: the products live in the night directory and follow the normal retention.

A night can be re-run offline:

    python -m RMS.Sprite.Detector /path/to/night_dir [--upload] [--overwrite]

## Testing

- `Tests/test_SpriteCore.py`, `Tests/test_SpriteUpload.py` and `Tests/test_SpriteDetector.py`: 141 tests, none of which need TFLite. The whole `Tests/` suite passes (218 tests).
  - The upload tests run against a loopback fake server that checks every signature independently.
- Real model on a real night (CA0005, 2018-10-08, 148 FF files): 15 s on a desktop CPU, 360 MB RSS. The run produced 3 detections. The one inspected is a lightning-lit cloud at the horizon, so the model has false positives near the horizon.
- **End to end against the real spritemap server** on loopback: 10 scenarios, 89 checks, all pass. Only the neural network is replaced, by a stand-in that finds the synthetic flash in the model input image. The cameras use real Croatian platepars (HR001H, HR001G, HR001B, HR0025, HR001A, HR000S), including one fitted with refraction off.
  - **Sprite at 60-80 km, 3 stations, live capture path:** height 70,015 m vs 70,000 m, 0.66 km off; every field on the server equals the station's record; all FF files and images requested and delivered byte-identical.
  - **Elve at 80-95 km, 2 stations, offline path:** 87,585 m vs 87,500 m, 0.26 km off.
  - **Refraction-off platepar paired with a refraction-on one:** 69,333 m vs 70,000 m, 0.60 km off.
  - **Other scenarios:** single-station event, duplicate resend, burst rejection, missing platepar, server down then back, unregistered key then added, and mask. Each behaves as intended.
  - **Platepars:** one stored per station regardless of the number of frames. The stored copy reproduces the stored directions to 0.0005 deg.

## Before enabling on stations

- ~~Publish the model~~: done. It is at globalmeteornetwork.org, and its checksum was verified from a real download.
- ~~Register station keys~~: not needed. The server reads the existing `/etc/rmssftp/users/<station>/authorized_keys` files, the same keys the stations use for their nightly upload.
- Benchmark inference on a Pi 4 and a Pi 5. It must stay well under the 10.24 s FF period.
- Merge and deploy the spritemap server (9helix/spritemap-data-pipeline) on gmn.uwo.ca behind HTTPS.
- On the stations: `detect_sprites: true` and `sprite_upload_url` set to the server (it is empty by default, which keeps uploads off).

https://claude.ai/code/session_01TPFxFiaabcrQJ2VrCEutfU


dvida (commented):
## Code Review: TLE Detector → Sprite Detector

I've reviewed the PR and pushed a commit with all the fixes directly to your branch. Here's a summary of what was found and changed.

### Critical Fixes Applied

**C1. Multiprocessing state isolation — filter + counter non-functional**
`find_tle` was registered as a QueuedPool worker. QueuedPool uses `multiprocessing.Pool`, so workers are separate processes. All mutations (`self.detection_counter += 1`, `self.fp_filter.add_detection(...)`) happened in forked copies — the parent never saw them. The FP filter accumulated nothing. The daily detection cap never incremented.

**Fix:** Restructured into standalone `detectSprites()` worker function + `processResults()` post-processing. Worker returns pure data. Filtering, CSV, image marking, FF copying all run in the parent process after `pool.getResults()`. This matches the existing `detectStarsAndMeteors` pattern in `StartCapture.py`.

**C2. FalsePositiveFilter confirmation logic inverted**
With `max_detections=3`, the check `len(self._buffer) + 1 <= self.max_detections` at pop-time only rejected the oldest detection in a burst of 4 — the other 3 passed through.

**Fix:** Taint-on-overflow approach. When buffer exceeds threshold, ALL buffered detections get tainted. Only untainted detections are confirmed when they age out.

**C3. TFLite interpreter not picklable / unsafe after fork**
TLEDetector (with TFLite Interpreter) was created in `Compressor.__init__` (parent process). Under fork: C extension state undefined. Under spawn/forkserver (Python 3.14+): TypeError crash.

**Fix:** Interpreter created lazily inside the worker function with a module-level cache per process, same pattern as `MLFilter.classifyPNGs`.

**C4. CALSTARS loaded before any stars are detected**
`readCALSTARS()` ran in `Compressor.__init__` before capture started — file doesn't exist yet. `check_star_num()` was always vacuously true.

**Fix:** Removed premature CALSTARS loading. Star filtering can be applied during post-hoc reprocessing when CALSTARS is available.

**C5. File deletion silently broken — wrong helper functions**
`getRawItems`/`deleteRawItems` filter by processed-frame suffixes (`.json`, `.mp4`, etc.). Sprite directories contain `.png`, `.csv`, `.fits` — none match. Nothing was ever deleted.

**Fix:** Switched to `getNightDirs`/`deleteNightFolders` which work with night-named directories.

### High-Severity Fixes

- **H1/H2.** `requests` import removed — `API_URL` was empty string, upload always failed. Upload functionality can be added when the API endpoint is ready.
- **H4.** `check_star_num` no longer returns `None` implicitly (moved to post-processing).
- **H5.** Global rename TLE → Sprite throughout all files to avoid collision with Two-Line Element satellite terminology.

### Styling Changes

- All functions renamed to camelCase (`get_prediction` → `getPrediction`, etc.)
- Module-level `log = getLogger("rmslogger")` instead of `self.log` passed via constructor
- f-strings → `.format()` for consistency with codebase
- Docstrings added to every function (codebase format with `[type]` brackets)
- GPL v3 license header added to both new files
- `from __future__` imports added
- 3-tier TFLite import cascade copied from `MLFilter.py` (ai_edge_litert → tflite_runtime → tf)
- Removed type annotations (`self.log: logging.Logger`)
- Local vars `MARKED_DIR` etc. → lowercase
- `import csv` moved to top with stdlib
- `log.info("Masking image")` → `log.debug` (runs every frame)
- Removed no-op slice `output[: output.shape[0]]`
- Config keys support both new (`sprite_*`) and old (`tle_*`) names for backward compat

### Architecture (before → after)

**Before:**
```
Compressor.__init__ (parent) → creates TLEDetector → creates QueuedPool + TFLite + FPFilter
Compressor.start() → forks child
Compressor.run() (child) → adds jobs → workers mutate FPFilter (invisible to parent)
Compressor.stop() (parent) → tle_detector.close() (sees nothing)
```

**After:**
```
StartCapture → creates QueuedPool(detectSprites) → passes to Compressor
Compressor.run() (child) → adds jobs (data_dir, ff_name, model_path, config)
Workers → create TFLite lazily, return (ff_name, detections, timestamp)
StartCapture → pool.closePool() → getResults() → processResults() (FP filter + CSV + images)
```

### Suggestion: Use max-ave Frame as Model Input

The current code feeds `ff.maxpixel` to the YOLO model. `maxpixel` contains everything bright: stars, light pollution, hot pixels, AND sprites. The model has to distinguish sprites from all persistent features.

`maxpixel - avepixel` removes all persistent features and is already the standard transient isolation technique throughout the codebase (`MLFilter.py:278`, `Detection.py:1043`, `StackFFs.py:115`).

```python
# Current:
image = Image.fromarray(ff.maxpixel).convert("RGB")

# Recommended:
max_ave = ff.maxpixel.astype(np.int16) - ff.avepixel.astype(np.int16)
max_ave = np.clip(max_ave, 0, 255).astype(np.uint8)
image = Image.fromarray(max_ave).convert("RGB")
```

**Note:** This changes the input distribution, so the `.tflite` model would need retraining on max-ave frames. The current model can stay on maxpixel in the interim.

You can also use `stdpixel` as a pre-filter to skip quiet frames (saves ~200-500ms per frame on RPi):
```python
max_ave = ff.maxpixel.astype(np.float32) - ff.avepixel.astype(np.float32)
threshold_map = max_ave > (3.0 * ff.stdpixel.astype(np.float32) + 5.0)
if threshold_map.sum() < 50:
    return (ff_name, [], timestamp)  # No significant transient, skip inference
```

A detailed review document with all findings is available here — let me know if you'd like the link.

dvida:
### Follow-up: what's left for you to do

The commit fixes all the architecture/style issues, but a few things need your input:

**1. Upload functionality was removed**
The `requests` import and `upload_detections` were removed because `API_URL` was empty and `requests` wasn't in the project dependencies. When the API endpoint is ready, the clean way to re-add it:
- Add `sprite_upload_url = ""` to ConfigReader defaults
- Add `requests` to `requirements.txt`
- In `processResults()`, after the FP filter confirms a detection, post to the API:
```python
if config.sprite_upload_url:
    import requests
    try:
        response = requests.post(config.sprite_upload_url, json=det.data)
        log.info("Upload status: {:d}".format(response.status_code))
    except Exception as e:
        log.error("Upload failed: {:s}".format(repr(e)))
```

**2. Model retraining on max-ave frames (recommended but not blocking)**
The current model works on maxpixel — that's fine for now. When you retrain on `maxpixel - avepixel` frames, just change the one line in `detectSprites()`:
```python
# Change this:
image = Image.fromarray(ff.maxpixel).convert("RGB")
# To this:
max_ave = ff.maxpixel.astype(np.int16) - ff.avepixel.astype(np.int16)
max_ave = np.clip(max_ave, 0, 255).astype(np.uint8)
image = Image.fromarray(max_ave).convert("RGB")
```

**3. `max_daily_detections` cap was removed**
The original had a 150-detection daily cap, but it was non-functional (counter in forked process). If you want it back, add it to `processResults()` — it's a simple `if len(confirmed) >= max_daily:` check in the parent process where it actually works.

**4. Star count filtering removed from live path**
`check_star_num` with CALSTARS was broken during live capture (file doesn't exist yet). If you want it for reprocessing, add it as a post-filter in `processResults()` where CALSTARS is available.

**5. Testing**
- Run `python -m RMS.SpriteDetector <test_folder>` in standalone mode
- Check FP filter: 2 detections in 60s → both pass; 4 → all rejected
- Verify `SpriteFiles` directories get cleaned up by `deleteOldDirs`

dvida:
## Real-Time Sprite Detection Pipeline

This batch of commits converts the sprite detection from a **batch QueuedPool** architecture to a **dedicated real-time `multiprocessing.Process`** that runs inference, filtering, calibration, and upload during capture — not after.

### What changed

**`RMS/SpriteDetector.py`** — Major rewrite (804 lines):

- **`SpriteDetector(multiprocessing.Process)`** — Long-running process (same pattern as `UploadManager`/`EventMonitor`). Pulls FF filenames from `multiprocessing.Queue`, runs TFLite inference, feeds the `FalsePositiveFilter` in real-time, and handles confirmed detections immediately. Uses `AtomicFlag` for lock-free IPC. Clean shutdown via `stop()` with SIGINT → terminate → SIGKILL escalation.

- **`SpriteUploader`** — HTTPS REST API client using `urllib.request` (stdlib, no new deps). Disk-backed JSONL queue (`SPRITE_UPLOADS_PENDING.jsonl`) for crash recovery. Exponential backoff: 15s × 2^attempt, capped at 5 min, max 5 retries. Two endpoints: `POST /api/v1/detections` (JSON + optional base64 PNG) and `POST /api/v1/files` (multipart FF upload when `sprite_upload_ff = True`). **Security: only `station_id` is sent — no station coordinates (lat/lon/elevation). HTTPS only.**

- **`detectSpritesInFF()`** — Standalone function for single-FF inference + calibration. Usable from both the Process and the `__main__` CLI.

- **`_calibrateDetections()`** — Converts 5 points per detection (centroid + 4 bounding box corners) from image x,y → RA/Dec J2000 → apparent alt/az. Uses `xyToRaDecPP` + `trueRaDec2ApparentAltAz` which handles J2000-to-epoch-of-date precession and refraction correction in one call. Sprites are exo-atmospheric (~40-90 km), so `refraction=True` is correct.

- **`_appendCSV()`** — Incremental CSV writer (appends per-detection, creates header once). Extended columns include all sky coordinates for centroid and 4 box corners.

- All inference utilities unchanged (`_getInterpreter`, `getPrediction`, `processPredictions`, NMS, etc.)

**`RMS/Compression.py`**:
- `sprite_detector=` parameter → `sprite_queue=` (`multiprocessing.Queue`)
- Removed `self.sprite_model_path` computation (model path now lives in `SpriteDetector`)
- `stop()` returns just `self.detector` (not a tuple)
- Queue put is a simple `(data_dir, filename)` tuple — no model path or config needed

**`RMS/ConfigReader.py`**:
- `sprite_dir`: `"SpriteFiles"` → `"SpriteData"` (matches `VideoData` naming)
- `sprite_days_to_keep`: `8` → `30` days retention
- New defaults: `sprite_upload_url = ""`, `sprite_upload_timeout = 10`, `sprite_upload_ff = False`
- New parser options in `parseCapture` for the upload config

**`RMS/StartCapture.py`**:
- Import: `detectSprites, processResults` → `SpriteDetector` class
- Creation: `QueuedPool(detectSprites, ...)` → `SpriteDetector(night_data_dir, config).start()`
- Wiring: `sprite_detector=sprite_detector` → `sprite_queue=sprite_detector.input_queue`
- Stop: `compressor.stop()` returns just detector (no tuple unpack)
- End-of-night: Removed `closePool/getResults/processResults/shutdownManager` batch block → single `sprite_detector.stop()` (results already saved in real-time)

### Why not QueuedPool

1. **Batch shutdown** — `closePool()` waits for `output_queue.qsize() == total_jobs`. Can't drain results mid-flight.
2. **Unnecessary overhead** — Manager server process + Pool for 1 core = 2 extra processes.
3. **State isolation** — FP filter and upload client need persistent state. Can't live in a Pool worker.

### API payload schema

```json
{
  "station_id": "US0001",
  "timestamp": "2026-08-28T03:14:22.000000Z",
  "ff_name": "FF_US0001_20260828_031422_123_0000512.fits",
  "night_dir": "US0001_20260828_031000_000000",
  "detections": [{
    "detection_type": "sprite",
    "confidence": 0.72,
    "centroid_x": 320.0, "centroid_y": 180.0,
    "box_x1": 290, "box_y1": 100, "box_x2": 350, "box_y2": 260,
    "ra_j2000": 183.50123, "dec_j2000": 45.19876,
    "azimuth": 145.321, "altitude": 62.718,
    "box_ra_j2000_1": ..., "box_altitude_4": ...
  }],
  "marked_image": "<base64 PNG>"
}
```

### Testing

60/60 tests pass covering: module imports, NMS/IoU/processPredictions with synthetic data, SpriteUploader queue/persist/retry/backoff, SpriteDetector process lifecycle (start → run → stop), FalsePositiveFilter integration (isolated confirm + burst rejection), CSV append (header + no-dup), ConfigReader defaults, Compression parameter signature, StartCapture import/wiring verification.

dvida:
## What changed in this pull request, in plain words

This comment explains the rework pushed to this branch. You don't need to know the code to follow it.

### The big picture

This feature lets an RMS camera station spot **sprites and elves**, brief red flashes high above thunderstorms, while it records the night. It then reports them to a central server, the "spritemap" server. When two or more stations see the same flash, the server works out where it happened and how high it was.

The first version of this pull request did most of the work, but it was written before the server was finished. The server's rules changed in the meantime, so the station and the server could no longer understand each other. The first version also had a number of bugs. Some of them meant the feature silently did nothing on a real station.

### What happens now, step by step

1. **The camera records as usual.** Every 10.24 seconds RMS saves a block of video as an "FF file". Each FF file is now also passed to the sprite detector.
   - The hand-off never makes the camera wait. If the detector falls behind, it skips a file and says so in the log. Recording is never affected.
2. **A trained model looks at the image.** It marks anything that looks like a sprite or an elve with a box and a confidence score.
3. **A first check throws out things that aren't real flashes.** A real sprite lasts a split second. The check looks at *when* the light inside the box appeared. If it was spread over many frames (for example a slowly moving plane or a star), the box is rejected. The same check tells us the exact moment of the flash.
4. **A second check throws out bursts.** If detections pile up (more than 3 files within a minute), the cause is almost always lightning, car headlights or clouds, not sprites. All of them are rejected.
   - This check needs to wait about a minute to see if a burst is coming, so each report goes out about a minute after the flash.
5. **The station works out where in the sky the flash was.** It uses its calibration file (the "platepar", which maps image pixels to sky directions). The result is compass direction and height above the horizon, at the exact time of the flash.
6. **Everything is saved in the night's folder:**
   - a table of every candidate, including the rejected ones, with the reason;
   - a file with the confirmed detections;
   - a picture of each detection, one with the box drawn on and one without.

   These files are archived along with the rest of the night's data.
7. **The detection is sent to the server.** The large video file and the pictures are *not* sent straight away. They are sent only when the server asks for them, which it does once a second station has seen the same flash. Most detections are seen by only one station, so this saves a lot of upload traffic.

### Bugs that were fixed

**Serious: the feature did not work on real stations**
- **The detector never shut down properly.** It only stopped after it had found 150 detections in one night, which almost never happens. Normally RMS waited for it, gave up, and killed it, losing the last detections of the night.
- **The sky positions were never calculated.** The code looked for the station's calibration file one folder above where it actually is, so it never found it.
- **The server rejected everything the station sent:**
  - the pictures were packed inside the message, which made it too big;
  - the picture file names were in the wrong format;
  - the security signature was in the old format.

**Timing and position errors**
- The time of the flash was up to 1 second early, because the milliseconds were cut off the file name.
- The sky position used the start of the 10-second block instead of the moment of the flash. The sky turns noticeably in 10 seconds.
- The position included the bending of light by the atmosphere. The server needs the direction without that bending, because it combines sight lines from several stations.

**Things that crashed the detector**
- A failed picture save.
- An error page from the server (for example during maintenance).
- A problem with a single FF file.

Each of these now logs a message and carries on.

**Things that silently did nothing**
- The station mask option (which blocks out trees and buildings) was never applied.
- The `--confidence` setting was ignored.
- None of the new settings were listed in the `.config` file, so the config check warned about them every night on every station.

**Speed**
- The first check rebuilt all 256 video frames in memory for every detection. On a Raspberry Pi that is slow. It now gets the same answer from one quick pass over the box.

### Security and reliability of the upload

- **Signature.** Every message is signed with the key the station already uses for its nightly upload, so there is nothing new to set up on the station. The signature covers what is sent, where it is sent, and when. A copied message cannot be sent again later or to a different address.
- **Retries.** If the internet is down, or the station's key isn't registered on the server yet, messages wait in a queue on disk and are retried later, even after a reboot. A retried message is identical to the first attempt, so the server never counts it twice.
- **Signing test.** The station and the server each have their own copy of the signing code, since a station can't run server code. Both projects now test against the same set of example messages, so if one side changes, a test fails before anything breaks in the field.

### The model file

The trained model is 35 MB, which is too big to keep in the RMS repository. Every RMS user would have to download it, even with sprite detection turned off. It has been **removed from the repository and from the history of this branch**.

Instead, a station downloads it once from the GMN server the first time sprite detection is turned on, the same way it already downloads a new calibration file. The download is checked with a checksum (a fingerprint of the file), so a damaged or tampered file is refused. If there is no model, sprite detection is simply off for the night and the camera works normally.

### How the code is organised

The first version was one large file of about 1,400 lines. It is now split into smaller files, each with one job: finding sprites, sky positions, the burst check, saving files, uploading, downloading the model, and running it all during the night.

Most of these can be tested without the model or any special software, which is how most of the bugs above were found.

### How it was tested

- **134 new automated tests.** The full RMS test suite passes (211 tests).
- **A real night.** A real night of data from station CA0005 (148 files) was run through the real model. It took 15 seconds on a desktop computer.
  - It found 3 detections. The one checked by eye was a false alarm: a cloud lit by lightning, low on the horizon. The code handled it correctly; the model itself makes this kind of mistake near the horizon.
- **The whole chain against the real server,** with two simulated stations looking at the same flash:
  - the server matched the two reports and placed the flash at 70.003 km, against a true height of 70.000 km;
  - it asked both stations for their files;
  - both stations sent them.

### Still to do before switching it on

1. Put the model file on the GMN server, in `sprite_model/sprite_detector.tflite` under the upload directory. Its checksum is `a73da954ab97a7b68683edc249eafe457f1085bfd8d619f05c9e0d94cdad831a`.
2. Time the model on a Raspberry Pi 4 and 5. Each file has to be processed well within its 10.24 seconds.
3. Register the stations' public keys on the spritemap server.
4. It would help to improve the model's handling of lightning-lit clouds near the horizon.

The feature is **off by default** (`detect_sprites: false` in `.config`). Stations that don't turn it on are not affected at all.

https://claude.ai/code/session_01TPFxFiaabcrQJ2VrCEutfU

dvida:
## Update: the model is now downloaded from the GMN website

The model file is published at https://globalmeteornetwork.org/projects/sprite_detector/sprite_detector.tflite, so stations now download it from there. The new commit is 9eb23d2e.

**What this means for a station:**
- The first time sprite detection is switched on, the station downloads the model over an ordinary secure web (HTTPS) connection. No login or key is involved.
- The file is checked against its known fingerprint (SHA-256) before it is used. A damaged or incomplete download is thrown away and tried again the next night. In the meantime the camera keeps recording normally.
- A download that is far too large (over 200 MB) or cut off part-way is also rejected, so a bad connection can't fill up the SD card or leave a broken file behind.
- The setting `sprite_model_base_url` in `.config` says where to download from. Nobody needs to change it; it exists in case the file is ever moved.

**Checked:** I downloaded the published file with the new code, and its fingerprint matches the one in the code. All 212 RMS tests pass.

**Station keys:** nothing needs to be registered. The server reads the keys the stations already use for their nightly upload, in `/etc/rmssftp/users/<station>/authorized_keys`.

**Left before switching it on:** timing the model on a Raspberry Pi 4 and 5.

https://claude.ai/code/session_01TPFxFiaabcrQJ2VrCEutfU

dvida:
## Update: all calibrations are now used, and the calibration is sent to the server

The new commit is 45d97b49. It makes two changes.

### 1. Calibrations fitted "with refraction off" are no longer refused

The air bends starlight slightly, which makes objects look a little higher than they are. This is called refraction. The server needs the true directions, without that bending.

RMS always removes the bending, in one of two ways:
- if the calibration (the "platepar") was fitted with refraction switched on, RMS removes it explicitly;
- if it was fitted with refraction switched off, the bending is absorbed into the fitted lens distortion, so it is removed anyway.

The earlier version refused calibrations of the second kind, so those stations would never have sent anything. About 12 of the 28 Croatian calibrations I looked at are of that kind. This check is now gone, and every calibration is used.

### 2. The calibration now goes to the server with each detection, stored only once

**Why:** with the calibration, the server can recompute the sky directions later without asking the station. For example, it could use the more accurate calibration that is refitted every night.

**Kept small:** a calibration file is about 7 KB, mostly a list of the stars used to fit it. The station sends a trimmed copy of 1–2 KB, with only the numbers needed to turn image pixels into sky directions.

**No duplicates:** the same calibration arrives with every detection of the night, but the server keeps only one copy. It recognises copies by a fingerprint of the content (SHA-256). Each detection just points to the stored copy. A new calibration, for example after the camera was bumped, gets a new fingerprint and is stored as a new copy. The station also writes the fingerprint into its own local record, so each detection can be matched to the exact calibration on the server.

**Privacy:** the calibration contains the exact station position. The public map only shows positions rounded to about 10 km, so calibrations are never published.

### How it was tested

I ran this station code against the real server code on my computer. The cameras used real Croatian calibrations, and the flashes were synthetic, placed at known heights. Only the neural network was replaced, by a stand-in that finds the synthetic flash in the image. 10 scenarios, 89 checks, all passing:
- **Sprite seen by 3 stations:** the server put it at 70.015 km, against a true height of 70.000 km, and 0.66 km from its true position. Every value on the server matched what the stations recorded. The video files and pictures were requested and arrived unchanged.
- **Elve seen by 2 stations:** 87.585 km against 87.500 km.
- **A "refraction off" calibration paired with a "refraction on" one:** 69.3 km against 70.0 km, 0.6 km off. So both kinds work together.
- **Calibrations on the server:** one copy per station, however many detections used it. The stored copy alone reproduces every stored sky direction to within 0.0005°.
- **Other checks:** single-station detections, repeated sends, bursts of lightning, a station without a calibration, the server being down, a station whose key wasn't registered yet, and the station mask. All behaved as intended.

All 218 RMS tests pass.

### Before switching it on

1. Dino times the model on a Raspberry Pi 4 and 5.
2. The spritemap server is deployed on gmn.uwo.ca.
3. On the stations, sprite detection is turned on and the server address is set in `.config` (`detect_sprites` and `sprite_upload_url`).

https://claude.ai/code/session_01TPFxFiaabcrQJ2VrCEutfU

dvida:
## Update: all sprite code is now in one folder, `RMS/Sprite/`

The new commit is 591c83bb. The sprite files used to sit loose in `RMS/`, next to the rest of RMS. They are now in their own package, `RMS.Sprite`, which keeps them together and makes clear what belongs to sprite detection. Inside the folder the "Sprite" prefix was no longer needed, so the names got shorter:

| Before | Now | What it does |
|---|---|---|
| `RMS/SpriteDetector.py` | `RMS/Sprite/Detector.py` | runs detection during capture, or on a finished night |
| `RMS/SpriteDetection.py` | `RMS/Sprite/Detection.py` | the model, the mask, the check that a flash is short |
| `RMS/SpriteAstrometry.py` | `RMS/Sprite/Astrometry.py` | turns image positions into directions in the sky |
| `RMS/SpriteFilter.py` | `RMS/Sprite/Filter.py` | throws out bursts (lightning, headlights) |
| `RMS/SpriteProducts.py` | `RMS/Sprite/Products.py` | writes the table, the records, the pictures and the message to the server |
| `RMS/SpriteUpload.py` | `RMS/Sprite/Upload.py` | sends to the server and answers its file requests |
| `RMS/DownloadSpriteModel.py` | `RMS/Sprite/DownloadModel.py` | downloads the model |

**Nothing else changed.** Only the file locations and the lines that import them are different. The files were moved with `git mv`, so each one keeps its history.

**The commands are now:**
- Run detection on a finished night: `python -m RMS.Sprite.Detector /path/to/night_dir`
- Download the model by hand: `python -m RMS.Sprite.DownloadModel`

**Checked after the move:**
- All 218 RMS tests pass.
- The full test against the real server still passes all 89 checks.

https://claude.ai/code/session_01TPFxFiaabcrQJ2VrCEutfU

PR #983, opened 2026-09-02
CameraControl: add --ip for standalone control without an RMS config

Cybis320:
## What

Adds `--ip` to `Utils/CameraControl.py` so a camera can be controlled directly by IP address, with no RMS config file:

```
python -m Utils.CameraControl --ip 192.168.42.10 GetDeviceInformation
```

## Why

Today `CameraControl` requires `--config` and drives whichever camera the config's device URL points at. That's right for a configured station, but awkward for a camera that isn't in a config yet — bench bring-up, provisioning a new unit, or flashing firmware before the station is set up.

## How

- New `--ip` argument (default `None`).
- When `--ip` is given, `__main__` calls the existing `cameraControl(ip, user, pwd, cmd, opts)` path directly instead of loading a config.
- Reuses the existing `--user`/`--password` overrides; they default to `admin` / empty, matching what `cameraControlV2` extracts from a device URL.
- `--ip` and `--config` are mutually exclusive.

Scope is deliberately just the entrypoint (argparse + `__main__`) — no behaviour change for the existing config-based path.

## Testing

- `--ip <addr> GetDeviceInformation` against a live GK7205V200 returns its Hardware / SoftwareVersion / Serial.
- `--ip <addr> --config <path>` is rejected with a clear message.
- `--help` lists the new argument.
- The default config path is unchanged (config-based invocation untouched).

### DIFF
--- a/Utils/CameraControl.py
+++ b/Utils/CameraControl.py
@@ -1078,8 +1078,12 @@ if __name__ == '__main__':
         help="Camera password (overrides the password in the config device URL)."
     )
 
+    # NOTE: no '-i' short form on purpose. Firmware-side ISP control forwards
+    # daemon args verbatim (isp_ctl `manual`/`auto` use -a/-d/-i/-e, where
+    # -i = ISP-digital gain). A '-i' short form here would let argparse swallow
+    # the daemon's `-i <val>` as the camera IP, silently dropping it.
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


markmac99 (approved):
Looks good to me. Will also be useful for non-RMS situations where one is using a compatible camera for other purposes.

PR #985, opened 2026-09-03
Stop a git credential prompt from blocking nightly processing

dvida:
During the GitHub outage on 2026-09-03, around 15 stations got stuck in the middle of nightly processing. The log stopped here:

```
2026/09/03 04:01:40-INFO-Reprocess-line:510 - Generate config audit report
Cloning into 'RMS'...
Username for 'https://github.com':
```

Pressing ENTER twice got them going again, and processing continued to archive creation. This happened on stations running continuous capture too.

## What was actually happening

The config audit report is not the problem, it is just the last thing that logs before the hang. `Utils/AuditConfig.py` only reads local files.

The real culprit is the next step, the repository lag check. `Reprocess.py` calls `finalizeObservationSummary`, which calls `daysBehind()`, which clones the commit history of every remote to see how far behind the station is. That clone was:

```python
p = subprocess.Popen(["git", "clone", url, "--filter=blob:none", "--no-checkout"],
                     cwd=target_directory, stdout=subprocess.PIPE)
p.wait()
```

Three things went wrong at once:

1. stdin was inherited. RMS runs in a terminal, so git had a real tty to ask on.
2. `GIT_TERMINAL_PROMPT=0` was never set, so git asked instead of failing. GitHub was answering anonymous HTTPS requests with a 401 during the outage, and a 401 is exactly what makes git ask for credentials, even on a public repository.
3. There was no timeout at all, just a bare `p.wait()`.

So git asked for a username, then a password. That is the two blind ENTERs.

The `try/except` around the call could not help, because nothing was ever raised, the process simply stopped. It was also a bare `except:`, which is why nobody saw this coming: when the lag check failed for any reason, it silently recorded "Not determined" and never wrote a single line to the log.

`Scripts/RMS_Update.sh` had the same problem and it is arguably worse. It runs unattended at boot from `RMS_FirstRun.sh`, before capture starts, and its `git ls-remote` check sends stdout and stderr to `/dev/null` but not stdin. During the same outage that would hang with no visible prompt at all.

## The fix

The idea is that no git call from RMS should ever be able to ask a question or wait forever, and that it should give up early rather than late.

**New helpers in `RMS/Misc.py`:**

- `runGitCommand()` runs git with every interactive prompt disabled, stdin on `/dev/null`, a hard timeout (120 s for network operations, 30 s for local ones), and output going to temp files instead of pipes. On a timeout it kills the whole process group, not just git, because git hands the transfer to a `git-remote-https` child that would otherwise keep the connection open. It also clears the credential helper list and tells git to give up on a transfer that stalls below 1 kB/s for 30 s.
- `hostReachable()` does a plain TCP connect with a 5 s timeout, so a dead link or a DNS failure is caught before anything slow is attempted.
- `gitUrlHost()`, `gitUrlHostAndPort()`, `anonymousHttpsGitUrl()`, `sanitiseGitUrl()` for picking apart remote URLs and keeping credentials out of the log.
- `disableGitPrompts()` sets the same environment variables for the whole process, which is called at the top of `StartCapture.py` and `Reprocess.py`. This is the safety net: it also covers GitPython and any git call added later that forgets to use `runGitCommand`.

**In `RMS/Formats/ObservationSummary.py`:**

- All 11 git calls now go through `runGitCommand`, so a stuck local git (a stale `index.lock`, say) cannot block either.
- Only remotes on known hosts are used. Previously `daysBehind()` cloned *every* remote from `git remote -v`, so a station with a private fork configured would hit a credential prompt even when GitHub was perfectly healthy.
- An ssh remote on an allowed host is rewritten to its anonymous https equivalent rather than skipped, so checkouts cloned over ssh still get a lag number without needing a key or a passphrase.
- The module now has a logger. When the lag check fails it says why, instead of quietly writing "Not determined".

**In `Scripts/RMS_Update.sh`:** the same environment variables plus a `git_net()` wrapper that adds `timeout` and closes stdin, applied to all six network git calls. The existing failure paths already degrade nicely ("Cannot reach RMS repository. Please check your internet connection"), they just were not being reached.

## Bugs fixed along the way

All of these were on the same code path and were being hidden by the bare `except:`:

- `os.listdir(target_directory)[0]` raised `IndexError` when the clone failed. The clone now goes into a directory we name ourselves, so there is nothing to guess.
- `commit_repo_directory` was only assigned inside the first-remote branch, so an empty remote list raised `UnboundLocalError`.
- `daysBehind()` returned a 2-tuple on success but a bare string `"Unable to determine"` on failure, so the caller's tuple unpack raised `TypeError`.
- The temporary directory was only cleaned up on the two success paths, never in a `finally`.
- `getCommit()` and friends used `os.getcwd()`, so the whole thing silently depended on the process being started from the RMS directory. Now uses `getRmsRootDir()`.
- Return codes of all four git calls were never checked.

## Tested

- 401 repository: returns rc=128 in 0.2 s with no prompt. Before the fix this is where it hung.
- Blackholed address (`https://10.255.255.1/x.git`, timeout 15 s): killed at exactly 15 s, no orphan `git-remote-https` left behind.
- Unreachable host: warns and skips without attempting a clone.
- Remote filtering: drops ssh remotes on unknown hosts, URLs with embedded credentials, and local paths; rewrites `git@github.com:...` to `https://github.com/...`.
- Happy path unchanged: `days_behind=20.194`, branch `remotes/walesto/master`, 6 s, temp directory cleaned up.
- `disableGitPrompts()` verified with a plain `subprocess` git clone that does not go through `runGitCommand`, to confirm the safety net works.
- `git_net()` tested on its own: rc=128 immediately on a 401, rc=124 at the timeout.
- `python -m py_compile` on all four Python files, imports of `RMS.Reprocess` and `RMS.StartCapture`, `bash -n` on `RMS_Update.sh`.

## Still to test before merging

I could not run these on my machine, so they need doing on a real station:

1. **A full night through `Reprocess.py`**, with and without GitHub reachable (easiest way to fake it is a `127.0.0.1 github.com` line in `/etc/hosts`). There is no captured data on my dev box. What to look for: `Generate config audit report` is followed by `Archiving detections to ...`, the archive and `observation_summary.txt`/`.json` are produced, and in the broken-network run the summary shows `repository_lag_remote_days: Not determined` with a warning in the log and no stall.
2. **`Scripts/RMS_Update.sh` end to end.** I deliberately did not run it, since it does `git stash`, `fetch` and `reset --hard` on the working repo. Needs a run with GitHub reachable to confirm a normal update still works, and a run with it blackholed to confirm it exits with the "Cannot reach RMS repository" message instead of hanging. Also worth a run on a station with https remotes, since my machine only has ssh ones.
3. **A Raspberry Pi.** `timeout` is from coreutils and is present on Raspberry Pi OS, but the fallback branch in `git_net()` has not been exercised. Also worth confirming the git version on the oldest station image is at least 2.3, which is when `GIT_TERMINAL_PROMPT` appeared. Anything older falls back to `GIT_ASKPASS`, which still works, but it would be good to know.
4. **A station with an ssh remote**, to confirm the https rewrite produces a sensible lag number rather than "Not determined". My dev machine has four ssh remotes and it worked there, but stations are set up differently.
5. **Whether 120 s is enough** on the slowest rural links. A blobless clone of the RMS history took 5 s here. If a slow station starts reporting "Not determined" the timeout is the first thing to raise.
6. **`daysBehind()` picks the first remote alphabetically** as the base clone, not `origin`. That is pre-existing behaviour and I did not change it, but on my machine it means the lag is measured against `remotes/walesto/master`, which is probably not what was intended. Worth deciding separately whether it should prefer `origin`.

## Not in this PR

The sweep turned up other unbounded calls that are real but separate problems: `urlopen` with no timeout in `EventMonitor.py`, `StarCatalog.py` and `AstrometryNetNova.py`; the clock-status subprocesses in `ObservationSummary.py`; `du` in `DeleteOldObservations.py`; and the ffmpeg calls in `GenerateMP4s.py` and `FRbinViewer.py` that are missing `-nostdin`. Happy to open an issue for those.

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -47,11 +47,20 @@ import tempfile
 import ephem
 
 from RMS.ConfigReader import parse
-from RMS.Misc import niceFormat, isRaspberryPi, sanitise, getRMSStyleFileName, getRmsRootDir, UTCFromTimestamp
+from RMS.Misc import niceFormat, isRaspberryPi, sanitise, getRMSStyleFileName, getRmsRootDir, \
+    UTCFromTimestamp
+from RMS.Misc import runGitCommand, GitCommandTimeout, hostReachable, gitUrlHost, \
+    gitUrlHostAndPort, sanitiseGitUrl, anonymousHttpsGitUrl, GIT_ALLOWED_HOSTS, \
+    GIT_NETWORK_TIMEOUT, GIT_LOCAL_TIMEOUT
+from RMS.Logger import getLogger
 from RMS.Formats.FFfits import filenameToDatetimeStr
 from RMS.Formats.Platepar import Platepar
 from RMS.CaptureDuration import captureDuration
 from RMS.CaptureModeSwitcher import SWITCH_HORIZON_DEG
+
+
+# Get the logger from the main module
+log = getLogger("rmslogger")
 
 
 if sys.version_info.major > 2:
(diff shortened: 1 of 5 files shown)


Cybis320:
Looking good. The only minor concern is the 300 s hard cap is too tight for a full reclone on slow links. The upstream repository is about 300 MB, so the attempt-5 reclone needs roughly 8 Mbit/s sustained to finish inside the budget. On a slower rural link the reclone can never succeed. The failure path restores the backup directory, so nothing is lost, but the recovery mechanism is dead for exactly the stations most likely to need it. The http.lowSpeedLimit setting already aborts a stalled transfer, so the hard timeout should be a large backstop for clone and fetch. I'd use two budgets: a short one for ls-remote, which moves kilobytes, and a long one for clone and fetch.

Cybis320:
Also see earlier fix already in prerelease introduced by PR #955. (I'haven't seen the Gh hang issue in any branch based on Prerelease).

Cybis320:
Although, I think 985 is the more complete fix and we should slim down 955 to deconflict with 985..

PR #987, opened 2026-09-06
Use the epoch-of-date convention for the platepar reference pointing everywhere

Cybis320:
## Summary

The projection kernels read a platepar's `RA_d`/`dec_d` as true equatorial coordinates in the **epoch of date** of the platepar JD: `pointingCorrection` shifts them by sidereal time and precesses to J2000 itself. But `Platepar.updateRefRADec`/`updateRefAltAz` (since 68bed054, 2021) convert between alt/az and `RA_d`/`dec_d` with the J2000 converters, and so do the ECSV/ASTRA header pointing, EventMonitor's FOV centre, SkyFit2's FOV-centre loading, its astrometry.net tangent-point refinement and ECSV export, and `Platepar.__repr__`.

The mismatch is the precession since 2000, ~21′ today and growing 50″/yr:

- a pointing set from alt/az (SkyFit "use config pointing", FOV-centre loading) lands that far from where it was requested, until the next star fit corrects `RA_d`;
- the `alt_centre`/`az_centre` written by SkyFit are that far from where the camera points (consumers: the CAL writer, EventMonitor's FOV checks, the FOV sky map, observation summaries, alignment initial guesses).

**Fitted platepars and all projections are unaffected**: the fit determines `RA_d` in the kernel's convention. Station-recalibrated platepars were already close because `ApplyRecalibrate` used the epoch-of-date conversion, but it stored the true altitude where SkyFit stores the apparent one (a 1–2′ mismatch between the two sources of `alt_centre`).

Measured on a platepar built through `updateRefRADec` from the shipped 720p template (lat 45, lon 15): the calibrated centre was **0.315°** from its own `alt_centre`/`az_centre`; after this change **0.001°**, with refraction on and off.

## Changes

- `Platepar.computeRefAltAz()`: the single conversion (epoch of date, plus refraction to apparent altitude when the platepar models it), used by `updateRefAltAz`, `__repr__`, `EventMonitor.platepar2AltAz` and SkyFit2. `updateRefRADec` does the inverse (apparent → true altitude, then epoch-of-date RA/Dec via `cyaltAz2RADec`).
- `ApplyRecalibrate` calls `updateRefAltAz()` so recalibrated platepars carry the same apparent alt/az flavour as SkyFit.
- ECSV/ASTRA header pointing (`ApplyAstrometryECSV`, `Astra`, SkyFit2 export) via `raDec2AltAz` (true altitude, no precession), as they intended with `refraction=False`.
[...]

PR #989, opened 2026-09-06
Fix the nutation in the true-of-date <-> J2000 transform used by the astrometric kernels

Cybis320:
## Summary

`equatorialCoordAndRotPrecession`, which `pointingCorrection` uses for every XY ↔ RA/Dec projection, applied nutation incorrectly: the nutation in longitude was a rotation about the **equatorial** pole instead of the ecliptic pole, and both nutation rotations were applied **after** precessing to J2000 rather than removed in the frame of date. The 4-term nutation series itself is fine, and the mean precession (`equatorialCoordPrecession`) is accurate to 0.08″.

Against the IAU 2006/2000A model (erfa), the transform true-of-date → J2000 was off by:

| Date | error of the transform (max over the sky) | with the fix |
|---|---|---|
| 2021-03 | 30.8″ | 0.05″ |
| 2024-09 | 3.4″ | 0.17″ |
| 2026-01 | 12.2″ | 0.02″ |
| 2027-07 | 24.4″ | 0.20″ |
| 2029-06 | 32.3″ | 0.17″ |

It follows the 18.6-year nutation cycle: near zero in 2024, growing towards ~34″ in 2029.

## Change

- Nutation is the standard rotation N = R1(−(ε+Δε))·R3(−Δψ)·R1(ε) from the existing series, removed at the start epoch and applied at the final epoch; J2000 is treated as the mean catalog frame. The kernel transform and the two converters `trueOfDateFromJ2000` / `j2000FromTrueOfDate` share one implementation, `frameRotation()` / `frameRotateVector()`, in plain C: the parameters are evaluated once per call and applied to both the pointing and the north vector. `equatorialCoordAndRotPrecession` went from 8.4 to 0.5 µs per call, and `cyXYToRADec` on the default meteor path (`precompute_pointing_corr=False`) from 11.5 to 3.7 µs per point.
- The position-angle change is computed from a north vector transported through the same rotation: a gnomonic projection defined in one frame is exactly the same projection in the other, rotated by the angle between the two frames' north directions at the centre.
- `cyTrueRaDec2ApparentAltAz` / `cyApparentAltAz2TrueRADec` use the same true-of-date transform. Before, they applied no nutation at all.
- **The sidereal time is now the apparent one.** True-of-date right ascensions have to be taken against the apparent sidereal time; with the mean sidereal time the hour angle is off by the equation of the equinoxes, Δψ·cos ε, up to 15.8″ over the nutation cycle. There were three implementations (`cyjd2LST`, `JD2HourAngle`, `JD2LST`, the last one feeding `geo2Cartesian`); all three now come from one kernel routine, `cyjd2GST`, within 0.36″ of erfa's `gst06a` from 2018 to 2036. Absolute error of the alt/az converters against erfa (GCRS → `pnm06a` → `gst06a` → topocentric):

  | date | before this PR | true frame + mean sidereal time | this PR |
  |---|---|---|---|
  | 2021-03 | 6.91″ | 14.44″ | 0.04″ |
  | 2024-09 | 9.62″ | 1.50″ | 0.09″ |
  | 2026-01 | 8.54″ | 5.72″ | 0.02″ |
  | 2027-07 | 7.51″ | 11.51″ | 0.10″ |
  | 2029-06 | 6.60″ | 15.05″ | 0.05″ |

- `Platepar.loadFromDict()` recomputes `Ho` from `JD`. Only the legacy text reader did; the JSON reader kept the stored value, so a platepar fitted with the old sidereal time would have been off by the equation of the equinoxes until its next fit. With `Ho` recomputed, the stale `RA_d` and the new `Ho` shift together and the pointing stays where it was fitted.
- `pointingCorrection` refracts the reference pointing in its own frame. It went through `eqRefractionTrueToApparent`, which expects J2000 input and precesses it to date, so the refraction of a true-of-date reference was evaluated ~22′ (in 2026) away from its real altitude: 8.2″ at 10° elevation, 2.4″ at 20°, 0.6″ at 45°, zero at J2000. The fit absorbed it and measurements were consistent, so it never showed.
- `pointingCorrection` is `cpdef` so it can be tested from Python. `pyerfa` is declared in `requirements.txt`.

## Verification

`Tests/test_precession_nutation.py` (erfa-dependent tests skip if pyerfa is absent). A distortion-free synthetic camera fixed in alt/az, referenced at the start of the night and followed for 8 hours on a 9×7 pixel grid over an 82×46° frame (two pointings, five dates 2021–2029): the kernel's XY → J2000 stays within 0.5″ of erfa at every grid point, with the rigid roll of the error field within 0.3″ and the non-rigid part within 0.1″. The centre alone cannot see a rotation or scale error; the corners can. The old transform drifts by up to 11″ in position and 11″ in roll over a night on this test. Also: transform within 0.5″ of erfa at five dates, J2000→date→J2000 round trips to 0.01″, the kernel transform and the converters agree to 0.001″, the position-angle change within 0.5″ of the exact geometry, a fixed camera's kernel centre within 2″ of the true direction with no drift over a night, the sidereal time against erfa over 2018–2036 and the identity of the three routines, the alt/az chain against erfa on the five dates, the reference refraction against refracting the true-of-date direction at its own altitude, and the hour angle recompute on a JSON platepar.

Real data: USV001 2026-09-05 (217 FFs) recalibrated with the old and new kernels. Star-fit residuals identical (mean 0.1587 vs 0.1594 px). Only the fitted parameters move: the pointing by a few arcsec (the old transform's error happens to be small at that direction and date) and `pos_angle_ref` by +14″. Within-night pointing scatter is unchanged (0.163′ → 0.148′ in altitude).

## What this changes for users

Nothing in measured astrometry: the error was a small rigid rotation of the sky, and the star fit absorbed it into `RA_d`, `dec_d` and `pos_angle_ref`. Those fitted values in **existing platepars** compensate for the old transform, so applied with the new kernel *without refitting* they are off by the same few arcsec to a few tens of arcsec until the next fit; the nightly recalibration does that automatically, and a SkyFit refit does it for a manually made platepar. The az/alt columns of FTPdetectinfo and ECSV files, which come from the alt/az converters, improve by the amounts in the second table.

`RA_d`/`dec_d` are, and always were, true-of-date coordinates: the kernel precesses them to J2000 before comparing with the catalog. Everything outside the kernel that reads them (`Platepar.updateRefAltAz`/`updateRefRADec`, `CheckFit`, `ApplyRecalibrate`, `EventMonitor.platepar2AltAz`, the ECSV headers, the ground-reference chain in SkyFit2) treats them as J2000 instead, which puts the stored alt/az centre 8–19′ from the real pointing. That is a separate, larger change and is in the follow-up PR stacked on this one. Nothing in this PR makes it better or worse.

Found while checking the coordinate chain during the review of #986; independent of that PR and of #987. Stacked under #990.


dvida:
## Review

I re-derived the new kernel math in pure Python and checked it against `pyerfa` rather than taking the numbers from the description. Everything substantive in this PR holds up, including the table in the description, which I reproduce exactly (30.81 / 3.37 / 12.19 / 24.36 / 32.29″). **This is correct and should go in.** One thing should be folded in before it does.

### Verified

| check | result |
|---|---|
| `nutationRotate` = N = R1(−(ε+Δε))·R3(−Δψ)·R1(ε) | correct; `inverse=True` is its exact transpose |
| `meanObliquity` vs Meeus eq. 22.2 | correct |
| new transform vs `erfa.pnm06a`, extended to 2010–2040 | **≤ 0.24″** (you claim 0.5″ — conservative) |
| `equatorialCoordPrecession` vs `precessionMatrix` | agree to 0.000000″, so the two new helpers and the matrix path are the same transform by construction |
| nutation over one 82°×46° FOV, after removing a best-fit rigid rotation | **0.001″** |

That last row is worth stating in the description: nutation is a *pure rigid rotation* at FOV scale, which is the reason the star-fit residuals do not move (0.1587 → 0.1594 px) and why nothing in measured astrometry can change. It is a much stronger argument than the empirical one.

Also checked: `equatorialCoordAndRotPrecession` has exactly one caller (`pointingCorrection`), so redefining it from "mean→mean precession" to "true-of-date→true-of-date" breaks nothing else; the `start_epoch != J2000_DAYS` float comparison is safe because that caller passes the module constant; and `Platepar.fitPointingNN`'s `delta_Ho` trick stays exactly consistent, since it leaves `ra_ref + LST(jd) − h0` invariant.

---

### 1. The equation of the equinoxes is ±15.8″, not "~1 arcsec" — please fold the fix in

The new `cyjd2LST` docstring says the equation of the equinoxes is "~1 arcsec", and the description uses that to justify leaving it out. It is ~1 *second of time*, i.e. **up to ±15.8″ of arc** (Δψ·cos ε). From `erfa.ee06a` on your own test dates: −14.37″, −1.59″, +5.78″, +11.39″, +15.28″.

This matters because mean-of-date RA with mean sidereal time was a *self-consistent pair*; true-of-date RA with mean sidereal time is not. Measured against erfa (GCRS → `pnm06a` → Earth rotation by `gst06a` → topocentric, no refraction), the absolute error of `cyTrueRaDec2ApparentAltAz` / `cyApparentAltAz2TrueRADec`:

| date | before this PR (mean frame + GMST) | with this PR (true frame + GMST) | with the EE term added |
|---|---|---|---|
| 2021-03 | 6.91″ | **14.44″** | 0.04″ |
| 2024-09 | 9.62″ | 1.50″ | 0.09″ |
| 2026-01 | 8.54″ | 5.72″ | 0.02″ |
| 2027-07 | 7.51″ | **11.51″** | 0.10″ |
| 2029-06 | 6.60″ | **15.05″** | 0.05″ |

So as it stands the PR makes the alt/az converters *worse* than before in roughly half the 18.6-year cycle — and those converters are what write the az/alt columns of every FTPdetectinfo and ECSV. The kernel itself is fine: `pointingCorrection` only ever uses `cyjd2LST(jd) − h0`, and the EE moves by <0.06″ over 10 hours, so it cancels within a night.

The fix is two lines and brings the whole chain to 0.03–0.11″:

- `cyjd2LST`: add `Δψ·cos(ε + Δε)` from the `nutationComponents` / `meanObliquity` you already have.
- `RMS/Astrometry/Conversions.py:284` `JD2HourAngle`: the identical term. **These two must move together**, because `pointingCorrection` differences them.

No platepar compatibility break: `Platepar.read()` recomputes `self.Ho = JD2HourAngle(self.JD)` (`RMS/Formats/Platepar.py:2218`), so every loaded platepar picks up the new definition automatically.

While you are there — RMS uses UTC as UT1 everywhere, and DUT1 reaches 0.9 s = 13.5″. That is also a rigid rotation about the pole and also absorbed by the fit, but it bounds what "absolute" can mean here, so I would soften *"The absolute meaning of `RA_d`/`dec_d` becomes correct"* accordingly.

### 2. `cyjd2LST`'s summary line now contradicts its own `Return:` section

The summary still reads *"Convert Julian date to apparent Local Sidereal Time. The times is apparent, not mean!"*, and the inline comment below still says `# Compute the apparent Local Sidereal Time (LST)`. With the fix above the summary becomes true and the new `Return:` note should go instead.

### 3. The transform is implemented twice (low)

`trueOfDateFromJ2000` / `j2000FromTrueOfDate` go through the scalar Meeus `equatorialCoordPrecession`; `equatorialCoordAndRotPrecession` goes through `precessionMatrix`. They are numerically identical, and you have a test asserting it — but having the matrix path call the helpers would make that test unnecessary.

### 4. The nutation series is evaluated four times per call (low)

`equatorialCoordAndRotPrecession` calls `nutationRotate` twice per epoch (pointing and north vector), each recomputing `nutationComponents` + `meanObliquity`, plus 4–6 NumPy allocations. This runs **per data point** in `cyXYToRADec` when `precompute_pointing_corr=False`, which is the default meteor path. Rotating both vectors in one call, or hoisting `(dpsi, deps, eps)`, would halve it.

### 5. Test and merge notes

- Rebase: this is based on `prerelease@3c1cded2`, which is now at `c65bde9a`. `git merge-tree` shows no conflicts.
- `pyerfa` is not declared in `requirements.txt`; it only arrives transitively via `astropy>=4.2`. On a Pi pinned to an older astropy, 4 of the 5 tests here silently skip. Worth declaring it, or using `pytest.importorskip("erfa")` like the newer files in `Tests/`.
- `TT_MINUS_UT = 69.2/86400` is hardcoded and the tests treat UTC as UT1. Harmless, because the truth model and the kernel make the same assumption — but a one-line comment saying so would save the next reader the derivation.
- #990 stacks on this: `Tests/test_aberration.py` imports nine helpers from `Tests/test_precession_nutation.py`, so if #990 ever lands first it fails at *import*, not as a skip.

### 6. Interaction with #991

#991 adds `Tests/TestRefraction.py`, whose truth model is built with `equatorialCoordPrecession` + `JD2HourAngle` (mean precession, mean sidereal time) and which asserts `< 1.0″` against it. **This PR is safe there** — nutation is rigid over one field (0.001″ measured), so the test's own fit absorbs it. #990 is not; see the comment there.

---

Numbers above are from a standalone transcription of the kernel checked against pyerfa, not from running the compiled kernel — the working tree here is on the #991 review branch.

Cybis320:
@dvida — thanks, item 1 is right and is now in. I reproduced your alt/az table to the hundredth of an arcsecond on the compiled kernel before touching anything, then pushed one commit (513e1c69, branch rebased onto current `prerelease`) and rewrote the description. While checking it I found two things the review and the original PR both missed; one of them is larger than anything in either.

## Folded into this PR

- **Apparent sidereal time everywhere.** There were three implementations, not two: `cyjd2LST`, `JD2HourAngle`, and `JD2LST` in `Conversions`, which feeds `geo2Cartesian` and so every ground reference. All three now come from one kernel routine, `cyjd2GST` (mean + `equationOfEquinoxes()`), within 0.36″ of `gst06a` over 2018–2036. Alt/az converters vs erfa: 0.02–0.10″ on the five dates.
- **`Ho` on load.** Your "no compatibility break" relied on `Platepar.py:2218`, which is the legacy text reader. JSON platepars go through `loadFromDict`, which kept the stored `Ho` (I wrote one with `Ho = 123.456` and read it back unchanged). `loadFromDict` now recomputes it from `JD`, so the stale `RA_d` and the new `Ho` shift together and existing platepars stay where they were fitted.
- **Reference refraction in its own frame.** `pointingCorrection` handed a true-of-date direction to `eqRefractionTrueToApparent`, which precesses J2000 input to date, so the refraction was evaluated ~22′ off in altitude: 8.2″ at 10° elevation, 2.4″ at 20°, 0.6″ at 45°, exactly zero at J2000 (control). Absorbed by the fit, so invisible, but it bounded the "absolute `RA_d`" claim more than DUT1 does. It now refracts via `cyraDec2AltAz` directly.
- **Items 2–5.** Docstring fixed. One `frameRotation()`/`frameRotateVector()` pair in plain C replaces both transform paths and the matrix helper; parameters evaluated once and applied to pointing and north vector: `equatorialCoordAndRotPrecession` 8.4 → 0.5 µs, `cyXYToRADec` on the default meteor path 11.5 → 3.7 µs/point. TT−UT1 comment added, `pyerfa` declared in `requirements.txt`. Test file name and helper names unchanged so #990's imports still work.
- New tests: sidereal time vs erfa and identity of the three routines, alt/az chain vs erfa, reference refraction vs own-frame refraction, `Ho` recompute on a JSON platepar. 11 pass; full `Tests/` 85 passed, 3 skipped.

## The larger thing: `RA_d`/`dec_d` are of-date, and nothing outside the kernel knew

This PR's description said "the absolute meaning of `RA_d`/`dec_d` (and anything derived from it, such as the stored alt/az centre) becomes correct", and you suggested softening it by DUT1. Neither was right. `pointingCorrection` precesses the reference from the observation date to J2000, so the fit defines `RA_d`/`dec_d` as true-of-date; but `Platepar.updateRefAltAz`/`updateRefRADec`, `platepar2AltAz`, the ECSV headers, `StarFilters`, `DrawConstellations` and the `__repr__` label all treat them as J2000 and precess a second time. Against the kernel's own reference direction on real platepars:

| platepar | `RA_d` read as J2000 (what `updateRefAltAz` did) | read as of-date |
|---|---|---|
| USV001, 2026-07 | +275″ az, +385″ alt | +3″, −9″ |
| UK009C, 2023-04 | +1079″ az, +380″ alt | +11″, −4″ |

That is the precession displacement since 2000, growing ~50″/yr. It never showed because nothing validates the alt/az centre independently. `CheckFit` and `ApplyRecalibrate` meanwhile convert without precession *and* without refraction, so two paths wrote `az_centre`/`alt_centre` that disagree with each other and with the kernel.

The ground references have the mirror problem: `geo2Cartesian` gives a true-of-date vector (now with GAST), but `GeoPoints.update` and `geoHt2RaDec` take it to J2000 with mean precession only, leaving the nutation in: 6.6–9.6″ from erfa today, and it would be 9.7–16.6″ with the sidereal-time fix alone. With `j2000FromTrueOfDate` it is 0.02–0.05″.

That is a change to a user-visible quantity in every platepar, so it is **#999**, stacked on this branch. This PR's description now says plainly that the stored alt/az centre is not fixed here. #991 touches `GeoPoints.update`, `geoHt2XY` and `geoHt2XYInsideFOV` too, so it will need a small rebase once #989 and #999 land; I would rather do that after your review of #991 than churn it now.

Cybis320:
Correction to my comment above: the `RA_d`/`dec_d` frame mismatch is not new, and it was not missed. **#987** (open since 6 September, "Use the epoch-of-date convention for the platepar reference pointing everywhere") already diagnoses it and fixes the platepar, SkyFit2, EventMonitor and ECSV-header side with `Platepar.computeRefAltAz()`, and the matching Janus change is merged. I overlooked it when writing #999, which overlaps #987 on six files and adds the ground-reference chain (nutation via `j2000FromTrueOfDate`), `CheckFit`, the FOV culling, and erfa-based tests. The two will be consolidated into one PR; until then treat #999 as provisional. Apologies for the noise.

PR #990, opened 2026-09-06
Model the annual aberration of light in the astrometric kernels

Cybis320:
Stacked on #989 (base branch `fix-nutation-frame`); only the aberration change shows in the diff. Rebased from the study branch `aberration-correction` and simplified: the correction is always on, there is no `.config` option.

## Summary

The star catalog gives barycentric directions; the camera sees them displaced by the **annual aberration of light**, up to 20.5″ towards the apex of the Earth's motion. Across a night's hour-angle range the alt/az projection of that displacement changes by up to ~0.6′. Without a model, the star fit absorbs it into `RA_d`, `dec_d` and `pos_angle_ref`: the fitted pointing of a fixed camera wanders through the night (the residual "chase" left in the calibration variation plots after the precession fix), and the platepar's reference pointing is not the true pointing.

- `cyraDecToXY` displaces catalog directions by the aberration before projecting; `cyXYToRADec` removes it. The platepar then describes the true pointing and stars still come out at their catalog positions.
- Earth's velocity from the Keplerian orbit (Meeus ch. 25, mean longitudes referred to the J2000 equinox): 0.02″ from erfa's `epv00` + `ab`.

## Stars and meteors are different frames

Light from a meteor is **not** aberrated: source and observer move together (the 0.3 ms light-time displacement and the aberration cancel exactly to first order). So:

| conversion | stars (`measurement=False`, default) | image measurements (`measurement=True`) |
|---|---|---|
| `xyToRaDecPP` | aberration removed → catalog direction | left in → geometric direction in the Earth's frame, which is what a trajectory solver uses |
| `raDecToXYPP` | catalog direction aberrated, then projected | projected as is (new keyword) |

The fit, star matching, photometry and the sky residuals all go through the star path. Every caller that converts meteor centroids or picks already passes `measurement=True` (`applyPlateparToCentroids`, `applyAstrometryFTPdetectinfo`, the ECSV writer, EventMonitor, SkyFit2 and Astra picks), so FTPdetectinfo gets Earth-frame directions with no further change. Callers that project Earth-frame objects through the star path (satellite positions, FOV footprints, EventMonitor's trajectory projection) are off by the 20″, i.e. 0.1 to 0.3 px at GMN scales, and were left alone.

Without this distinction (the study branch removed the aberration for everything), meteor directions would have stayed where they are today: biased by the aberration, because the uncorrected fit absorbs it and the removal then takes it out again.

## Verification

`Tests/test_aberration.py`: shift magnitude and inverse, velocity against erfa (0.1″), forward/reverse round trips for both paths, that a measurement at a star's pixel comes out at the star's apparent (aberrated) direction to 0.01″, and the star fit as in real life on the distortion-free synthetic camera of #989's test: with the platepar holding the true pointing, stars project to their catalog directions (erfa velocity aberration inverted) within 0.5″ over the whole 82×46° frame through the night, with no roll (0.3″) or scale (0.1″) error. Today's kernel on the same test drifts 33″ at the centre over a night and leaves 7 to 15″ at the corners that no pointing or rotation refit can absorb.

44 nights from 17 GMN stations, recalibrated with the correction off and on (study branch): the fits move by the predicted amount (slope 1.01 in altitude, 0.99 in azimuth, expected 1.00, for every lens, pointing and season including two southern stations), and the settled-night altitude drift of the quiet cameras drops by about 40%. The residual after the correction has no aberration signature (slope 0.01 on the prediction) and does not repeat by sidereal time; it is thermal and mechanical. The nutation error fixed in #989 contributes only a constant offset within a night (its diurnal wander is 0.02 to 0.09′ p2p for the study dates), so those conclusions carry over.

USV001 2026-09-05 (217 FFs, 693 meteor centroids), this branch against the study's runs:

| | off | study on | this branch |
|---|---|---|---|
| star-match residual | 0.1587 px | 0.1594 px | 0.1586 px |
| settled pointing scatter alt / az | 0.156′ / 0.076′ | 0.057′ / 0.068′ | 0.057′ / 0.061′ |
| meteor centroid shift vs off, mean / max | | 6.4″ / 16.5″ | 16.9″ / 29.3″ |

The aberration at those centroids' directions and times is 16.8″ mean (20.2″ max); the measured shift divided by it has a median of 0.98. The maximum includes the constant 7 to 9″ from the nutation fix. The study run moved the meteors only by the part of the aberration that a rigid pointing shift cannot absorb across this 82° field.

## What changes for users

- Meteor RA/Dec and alt/az in FTPdetectinfo move by the aberration at the meteor's direction, up to 20.5″ (direction-dependent, so not a common-mode shift between stations): ~10 m at 100 km range and ≲0.01° in radiants, systematic with respect to the apex. WMPL treats the observations as geometric Earth-frame directions, so this removes a bias rather than adding a correction to be undone downstream.
- Existing platepars absorbed the aberration of their fit date; applied without a refit they are off by the difference to today's aberration (up to ~0.6′ seasonally). The nightly recalibration refits them. In the calibration variation plots the corrected fits sit at a constant offset from an uncorrected reference platepar until the reference is refitted.
- The kernel keyword `aberration` (default `True`) remains on `cyraDecToXY` / `cyXYToRADec` for tests and for reproducing the old behaviour.

https://claude.ai/code/session_01CzZHCZoftFjV712njkWB5v


dvida:
## Review

Checked the same way as #989 — the aberration model re-derived and compared against `pyerfa`, and the field-scale effect measured rather than taken from the description. **The physics is right, the velocity model is better than you claim, and the stars-vs-measurements frame split is the correct design.** The blocker is the set of call sites the description says were "left alone".

### Verified

| check | result |
|---|---|
| `v = κ(−(sin λ + e sin ϖ), cos λ + e cos ϖ)` | re-derived from the two-body hodograph — correct, including `perihelion_lon = L0 − M + π` being the *Earth's* perihelion longitude (≈102.9°), which is the easy sign to get wrong |
| precession of the longitudes to the J2000 equinox (1.3969713°/cy) with the J2000 obliquity | self-consistent; the neglected ecliptic tilt contributes ~0.001″ |
| velocity vs `erfa.epv00` + `erfa.ab`, 2021–2027 | **≤ 0.018″** (you claim 0.02–0.05″) |
| ordering vs refraction | correct: aberrate → refract → project, exactly inverted on the way back |
| first-order inverse in `removeAberration` | ~0.002″, as documented |

And the quantitative case for the PR, which I think is stronger than what the description says. Residual of the aberration field over one FOV after removing a best-fit rigid rotation, and after also absorbing an isotropic scale term (i.e. what a full `RA_d`/`dec_d`/`pos_angle_ref`/`F_scale` refit cannot take out):

| FOV width | rotation absorbed | rotation + scale absorbed |
|---|---|---|
| 82° | 7–16″ | **10–12″** (≈0.12 px at 90″/px) |
| 40° | 2–8″ | 1.5–3″ |
| 20° | 0.7–4″ | 0.4–0.8″ |

So the correction genuinely buys something on a wide GMN lens and almost nothing at 20°. Worth saying, because it tells a station operator whether to expect their plots to change.

---

### 1. Blocker: Earth-frame directions still go through the star path

These are one-keyword changes, and the first one is a round-trip error in a shipped path, not a cosmetic offset:

| file | what happens | error |
|---|---|---|
| `ApplyAstrometry.py:1207` `applyPlateparToRaDecCentroids` | reads RA/Dec that `applyPlateparToCentroids` wrote with `measurement=True`, then reprojects them with the default | **~41″, 2× the aberration** — this is the `radec_input=True` path of `applyAstrometryFTPdetectinfo` |
| `ApplyAstrometry.py:1294` `xyHt2Geo` | passes `measurement=False` with the comment *"Disables refraction correction"*; that flag now also strips the aberration from a ground line of sight | 20.5″, and the comment becomes actively misleading. Feeds `FOVArea` and `Flux` collecting area |
| `ApplyAstrometry.py:1383` `geoHt2XY`, `:1528` `geoHt2XYInsideFOV` | geodetic targets through the star path — **and at `J2000_JD.days`**, so the aberration vector applied is the one for 2000-01-01 | 20.5″ in a fixed wrong direction |
| `Platepar.py:109` `getCatalogStarsImagePositions`, called at `SkyFit2.py:4219` with `self.geo_points` | ground reference points, used as **plate-fit references**, get the star aberration | 20.5″ |
| `SatellitePositions.py:485` | satellite topocentric ICRF directions through the star path | 20.5″ |
| `KalmanFilter.py:549` `saveECSV` | meteor picks with `measurement=False`, while `ApplyAstrometryECSV.py:131` writes the same columns with `measurement=True` | two ECSV writers, two frames, 20.5″ apart |
| `SkyFit2.py:15212` → `:15291` `updateGreatCircle` | picks read with `measurement=True`, great circle reprojected with the default | ~41″ ≈ 0.4 px on the overlay (cosmetic) |
| `AddCelestialGrid.py:389`, `:420` | alt/az grid lands 20.5″ off the equatorial grid (cosmetic) | 0.2 px |

The SkyFit2 geo-points one is the one I would not ship: those are fit *references*, and it is the same ground-vs-sky contract that the refraction path already honours via `pp_noref` — the aberration side of that contract is now missing.

### 2. `measurement` now carries two unrelated meanings

`xyToRaDecPP(measurement=True)` means "post-hoc refraction correction **and** do not aberrate", while `raDecToXYPP(measurement=True)` means only "do not aberrate". That overload is precisely why the sites above are easy to get wrong, and why `xyHt2Geo`'s existing comment silently became false. Either give the kernel-facing concept its own name (`earth_frame=`, matching the ground-vs-sky wording already used for refraction) or state the one contract in a single place that both wrappers point at.

### 3. Perf asymmetry in the reverse kernel (low)

`cyraDecToXY` hoists `earthVelocityJ2000` out of the loop; `cyXYToRADec` calls `removeAberration` per point, recomputing ~10 trig ops each time, on the path where `pointingCorrection` is already per-point. The velocity moves 0.12″ over 8 hours, so caching it whenever `jd` changes costs nothing in accuracy.

### 4. Omitted terms worth naming in the description (not blockers)

- **Diurnal aberration**: up to 0.32″ at the equator, 0.23″ at 45°, and it has a *diurnal* signature — directly relevant given that the stated goal is the night-long pointing wander. Worth one sentence saying it is 40× smaller than the annual term and deliberately out of scope.
- **Annual parallax** of the brightest catalog stars: Sirius 0.38″, α Cen 0.75″, Procyon 0.29″ — larger than the 0.1″ tolerance the tests assert against erfa.
- Heliocentric rather than barycentric velocity: 0.006–0.011″, already inside the measured 0.018″.

### 5. Interaction with #991 — this is the one that needs coordinating

#991 adds `Tests/TestRefraction.py`, which builds its truth model with `equatorialCoordPrecession` + `JD2HourAngle` (mean precession, mean sidereal) and asserts `< 1.0″` against it at `:281`, `:435-436` and `:521`.

- #989 is safe there — nutation is rigid over one field (0.001″ measured), so the fit absorbs it.
- **This PR is not**: 10–16″ over the 82° field survives any rigid refit, so those assertions are expected to fail. `TestRefraction.py:518` also passes `measurement=False` on a **ground** path, which becomes wrong by 20.5″ here.

Whichever of #990 and #991 lands second has to update that truth model.

### 6. Operational: this is a data discontinuity worth marking

Every meteor RA/Dec in the network moves by up to 20.5″, direction-dependent and systematic with respect to the apex, so it does not cancel between stations. `Platepar` already carries `self.version = 2` (`Platepar.py:193`) — bumping it, or adding an explicit marker, would let downstream trajectory work tell aberrated from non-aberrated observations instead of inferring it from a date.

### 7. Merge ordering

`Tests/test_aberration.py` imports nine helpers from `Tests/test_precession_nutation.py`, so this must land after #989 or it fails at import rather than skipping.

---

Separately, two pre-existing things I noticed while tracing the call sites, unrelated to this PR: `Utils/Flux.py:1119` `detectMoon` feeds `ephem`'s apparent *of-date* `m.ra`/`m.dec` into `raDecToXYPP`, which expects J2000 (~0.36° in 2026, only used for a coarse FOV margin check); and `CheckFit.py:722` / `ApplyRecalibrate.py:626` write `az_centre`/`alt_centre` through the non-refracting `raDec2AltAz` while `Platepar.updateRefAltAz()` uses the refracting kernel, so that field already differs by the refraction depending on which path last touched it.

Numbers above are from a standalone transcription of the kernel checked against pyerfa, not from running the compiled kernel — the working tree here is on the #991 review branch.

Cybis320:
Rebased onto #991 (which sits on #999 and #989), base changed accordingly: two commits now.

**Commit 1** is the original aberration model, unchanged apart from the merge: the kernels take both `refraction_scale` (from #991) and `aberration`.

**Commit 2** is what item 5 of the review asked of whichever PR landed second, plus the minimum of item 1 needed for the two stacks to agree:

- `xyToRaDecPP` / `raDecToXYPP` get an explicit `aberration` keyword (`None` = the previous "not measurement"; `False` for a direction fixed to the Earth). As you noted in item 2, `measurement` cannot carry this, since on an unrefracted plate `measurement=True` also triggers the post-hoc refraction correction. `xyHt2Geo`, `geoHt2XY` and `geoHt2XYInsideFOV` pass `aberration=False`.
- `targetRaDecToPlateRaDec` takes the aberration out of the plate coordinates of a ground reference too, `input = A⁻¹(T⁻¹(T(true, f·s), s))`, which is the fit-reference case you said you would not ship without.
- The truth models in `Tests/TestRefraction.py` and `Tests/TestReferenceFrame.py` return catalog directions for stars and geometric ones for ground targets and meteor-style measurements. Before that, the star fits carried the 10–16″ non-rigid residual and the ground path was 20″ off, as predicted.

Checked on the compiled kernel: `applyAberration` is within 0.007″ of `erfa.ab`, the remove/apply closure is 0.0001″. Full `Tests/`: 151 passed, 3 skipped.

**Still open from the review**, deliberately not touched in a rebase: the other Earth-frame call sites in item 1 (`applyPlateparToRaDecCentroids`, `getCatalogStarsImagePositions` for geo points in SkyFit2, `SatellitePositions`, `KalmanFilter.saveECSV`, `updateGreatCircle`, `AddCelestialGrid`), the per-point `removeAberration` in the reverse kernel (item 3), the description notes on diurnal aberration and parallax (item 4), and the platepar version marker (item 6). Merge order is #989 → #999 → #991 → #990; `Tests/test_aberration.py` still imports its helpers from `Tests/test_precession_nutation.py`, which is unchanged in name.

PR #991, opened 2026-09-06
Scale the refraction with the station height and add the finite-target-height fraction

Cybis320:
## Summary

The kernels apply Bennett's refraction formula at sea-level standard conditions for every station. Refraction is proportional to the air's pressure over temperature, so a station at 2000 m sees 18% less of it and one at 2800 m 24% less. For the GMN cameras above Tucson (USV001–003 at 2074–2796 m) the model error is 14–37″ at the bottom of the frame. The star fit absorbs most of it into pointing, scale and the radial terms, but a vertical gradient of 4–11″ survives into the meteor astrometry (worst at the bottom of a frame pointing at 35°), with star residuals of 1–4″ that do not reveal it.

| station | elevation | refraction ratio | pointing alt | raw model error, frame bottom | star rms after fit | surviving meteor error |
|---|---|---|---|---|---|---|
| USV001 | 2074 m | 0.82 | 44° | 21″ | 1.6″ | 5″ |
| USV002 | 2796 m | 0.76 | 54° | 20″ | 1.1″ | 4″ |
| USV003 | 2610 m | 0.77 | 35° | 37″ | 3.9″ | 11″ |

While reviewing that, three further problems in the same area turned up and are fixed here too: the refraction was still being applied to ground-based references and measurements, the ground projection used a flat-Earth range, and the new inverse refraction was not converged near the horizon.

### Station height

- **`refractionScale(elev)`**: the International Standard Atmosphere pressure-over-temperature ratio relative to sea level, exactly 1 at sea level so sea-level stations are unchanged. Every refraction call in the kernels (`cyraDecToXY`, `cyXYToRADec`, `pointingCorrection`), the converters (`cyTrueRaDec2ApparentAltAz`, `cyApparentAltAz2TrueRADec` and their `_vect` forms), the Python wrappers and `Platepar` takes it from the platepar elevation through a new `refraction_scale` keyword (default 1.0). The weather is not modelled: ±10 K or ±10 hPa move the refraction by about 3.5% or 1%.
- **`refractionTargetFraction(elev, target_height)`**: the part of a star's refraction that applies to a target at a finite height. From the ISA refractivity profile, in closed form: 0.92 for a meteor at 100 km, 0.72 at 30 km, 0.38 for a contrail at 10 km, seen from sea level. Valid for a target above the observer; below it the mean refractivity along the path exceeds the one at the observer and the closed form goes negative, so that case returns 0. **Used by the ground path with a fraction of 0; the meteor and contrail paths still apply the star value**, and wiring them up is the next step.

### Exact inverse refraction

The true-to-apparent refraction is now the exact inverse of the apparent-to-true one instead of Saemundsson's separate formula, which differed from that inverse by up to 4″. The star fit works in image coordinates (true → apparent) and the measurements go the other way, so that difference was a bias of the same size between the two. It also meant `updateRefAltAz()` followed by `updateRefRADec()` drifted by up to 4″ per round trip.

Bennett's formula is factored into `bennettRefraction()`, which returns the refraction and its analytic derivative, and the inverse is solved with Newton's method. A plain fixed-point iteration contracts at `|dR/dh|`, which is ~0.03 at 5° elevation but 0.21 at the horizon, so a fixed number of steps cannot serve both: five steps leave 0.31″ at 0° and 1.08″ at −0.5°, exactly where the celestial-grid horizon line and low-elevation ground targets sit. Newton closes to double precision (8e-11″) from −0.5° to 90° at any scale, in 2.45 refraction evaluations on average.

Cost on `cyraDecToXY`, the inner loop of `fitAstrometry` and so of the nightly recalibration, three clean rebuilds on one machine: 1.494 µs/star before, **1.617 µs/star after (+8.2%)**. The `precompute_pointing_corr=False` path of `cyXYToRADec` is unaffected — refraction is 1.2% of it.

The forward projection moves by 0.6–3.7″ (under 0.05 px at GMN scales); the next recalibration absorbs it.

### Ground-based objects are never refracted

A ground reference is close enough that no measurable refraction accumulates between it and the camera, so its light arrives along its true direction. Two places were still treating it like a star.

**The fit.** `GeoPoints.update()` builds a geo point's RA/Dec from a pure geometric ECI difference vector — no refraction, correctly — but `fitPickedStars` → `fitAstrometry` used the platepar with the user's `refraction` flag, which defaults to on, so the plate refracted the landmark as if it were a star and the fit absorbed the difference into pointing and distortion.

New `ApplyAstrometry.targetRaDecToPlateRaDec()`. The plate maps an input through its own true-to-apparent refraction `T(·, scale)`, the target is seen at `T(true_dir, fraction·scale)`, so the input that makes the plate reproduce the arrival direction of the target is

```
input = T⁻¹(T(true_dir, fraction·scale), scale)
```

With `fraction = 0`, the value for a ground reference, that is the plate's own refraction taken back out. `GeoPoints` now carries these plate coordinates alongside the true ones and `GeoPoint.coords()` returns them, so drawing, FOV filtering and the fit all use them. This only cancels exactly because the two refraction directions are now true inverses. Synthetic camera at 2610 m pointing at 20°, worst error over the frame:

| fit references | ground measurement | sky measurement |
|---|---|---|
| stars only (reference) | 0.03″ | 0.03″ |
| geo points, true direction | **163.81″** | **165.07″** |
| geo points, plate coordinates | 0.03″ | 0.03″ |
| stars + geo points, true direction | **19.47″** | **19.62″** |
| stars + geo points, plate coordinates | 0.03″ | 0.03″ |

The resulting plate is an ordinary refraction-on plate: meteors and stars come out as true J2000 coordinates through `xyToRaDecPP()`, ground picks through `Platepar.switchToGroundPicks()`, both from the same plate.

**The ground projection.** `xyHt2Geo()`, `geoHt2XY()` and `geoHt2XYInsideFOV()` only cleared `platepar.refraction`. That stops the points being refracted, but it leaves the reference pointing at the true direction of the camera axis while the camera looks along the apparent one, so the whole field is offset by the refraction at the FOV centre — 194″ on a camera pointing at 20°. `switchToGroundPicks()` has done this correctly for the manual-reduction ground picks since 2021, so they use it now. Displacement of the projected ground position, target 3 km above the station:

| FOV centre altitude | 60° | 45° | 30° | 20° | 10° |
|---|---|---|---|---|---|
| shift | 0.8 m | 2.4 m | 9.4 m | 38.9 m | 642.4 m |

### The ground projection solves for the range instead of assuming a flat Earth

`AEGeoidH2LatLonAlt()`, which `xyHt2Geo()` projects a pixel onto the ground with, took the range to the target as `r = (h - alt)/sin(elev)`. That ignores the Earth curving away under the line of sight, so the point ends up too high, by more the lower the camera looks:

| elevation | 60° | 45° | 30° | 20° | 10° | 5° |
|---|---|---|---|---|---|---|
| target at 3 km | +0.2 m | +0.7 m | +2.1 m | +5.3 m | +22.7 m | +92.3 m |
| target at 100 km | +258 m | +773 m | +2320 m | +5836 m | +24833 m | +100315 m |

At 5° elevation a target asked for at 100 km was placed at 200 km. It also meant `xyHt2Geo()` and `geoHt2XY()` were not inverses of each other, since `geoHt2XY()` places the target at exactly `h`: round tripping a pixel through the pair came back about 3900 px away on a camera pointing at 25° altitude.

`Conversions.AEH2Range()` already solves this, so it is used. Its default analytical path is left untouched — `EventMonitor`, `ShowerAssociation` and `FOVArea` depend on it — but the `accurate=True` path is rewritten, because it could not be used as it stood: it minimised the square of the height residual with `scipy.optimize.minimize` rather than solving for the root, its starting guess compared an elevation in degrees against `np.radians(10)` so the branch never fired, and the analytical solution it started from returns the far intersection for a target below the observer (1099 km for a point 1.2 km below the camera at −5° elevation instead of 14 km). A ridge-top camera looking down at a landmark is the common geo-point geometry, so that case has to work.

It now solves the sphere intersection in closed form, picking the near crossing when the target is below the observer and the far one when it is above, then refines on the WGS84 ellipsoid with Newton's method using the component of the line of sight along the local vertical as the derivative. That lands within a millimetre of the requested height at every elevation from −10° to 60°, and returns NaN when the line of sight never reaches the height at all, which the flat-Earth formula answered with a point behind the camera. The round trip is now 0.22 px, the rest being the epoch convention of `geoHt2XY()`, which evaluates the plate at J2000 so that the RA/Dec of date it computes may be taken as J2000, plus the forward/reverse polynomial mismatch of the plate itself.

### Other

Several call sites that relied on the `refraction=True` default now pass `refraction=platepar.refraction` (`AddCelestialGrid` ×5, `AutoPlatepar`, `EventMonitor.platepar2AltAz`, `SkyFit2` ×7). This is a consistency fix — `addEquatorialGrid` was refracting the grid while `raDecToXYPP` honoured the flag — but it changes the drawn overlay for any platepar with `refraction = False`, which is what every legacy v1 platepar loads as.

The `int isinf(double)` extern is dropped: C99 makes `isinf` a type-generic macro rather than a function, so declaring it is not portable, and the test was redundant since infinity already satisfies `target_height > 1e6`.

## Verification

`Tests/TestRefraction.py` (26 cases):
- the scale against an independent ISA computation, and exactly 1.0 at sea level;
- the target fraction against a ray trace through a spherically symmetric standard atmosphere (Snell's law in the form n·r·sin z = const): within 2% of the star refraction for targets below 30 km at any elevation from 5°, and within 2% for 100 km above 10°, where the curvature of the Earth starts to matter;
- exact closure of the two refraction directions at any scale, down to −0.5° elevation, which is what catches an unconverged inverse;
- a synthetic USV003 star fit (camera at 2610 m pointing at 35°, stars refracted with the true scaled refraction, platepar fitted by RMS): with the station height the fitted platepar's meteor astrometry is unbiased over the whole frame; with the sea-level model it is biased by more than 4″;
- a synthetic ground-reference fit, geo points alone and mixed with stars, checking that both ground and sky measurements of the fitted plate are right;
- the ground projection against the line of sight of a synthetic camera.

`Tests/TestGeoConversions.py` (17 cases): the solved range against the height it actually reaches, from −10° to 60° elevation and for targets above and below the observer; the near crossing for a downward line of sight; the unreachable case; the unchanged analytical path; and the `xyHt2Geo`/`geoHt2XY` round trip.

## What changes for users

- Sea-level stations: nothing from the scale. All stations: the forward projection moves by up to 3.7″ from the exact inverse, absorbed at the next fit.
- Stations at altitude: after the next recalibration the fitted pointing, scale and distortion re-absorb the smaller refraction, and the meteor astrometry loses the 4–11″ vertical bias. Until then, an existing platepar applied with the new kernel is off by the difference between what its fit absorbed and the new model, of the same size.
- Ground picks at high-altitude stations shift until the next recalibration for the same reason: `switchToGroundPicks` preserves `alt_centre`, which `updateRefAltAz` now computes with the scaled refraction, so an existing platepar moves by 24.6″ for USV003 at 2610 m pointing at 35° (1.8″ at sea level, from the exact inverse). It vanishes after a refit, because `alt_centre` is then the camera's real geometric pointing and is model-independent.
- Fits that include geo points change by up to a full refraction, and ground projections by up to the numbers in the two tables above. Both are corrections.
- The contrail (Janus) and meteor paths can now call `refractionTargetFraction` with the target height they already carry and pass it to `targetRaDecToPlateRaDec`; that is the next step and where the remaining effect is.

## Known limitation, not addressed here

`xyHt2Geo()` and `geoHt2XY()` still disagree by about 0.2 px because they evaluate the plate at different epochs by design. That is far below the accuracy of the ground reference data itself and is left alone.

Independent of #989 and #990, but all three touch the kernel signatures, so expect trivial conflicts in whichever merges last.

https://claude.ai/code/session_01CzZHCZoftFjV712njkWB5v


dvida:
Reviewed and pushed four follow-up commits onto this branch (rebased onto current `prerelease`, which it was behind).

The physics is right: the ISA constants check out (`0.0065/288.15`, `5.25588`, `R*T/(M*g) = 287.05*216.65/9.80665 = 6341.6 m`), `isaRefractivityIntegral(inf) = 8433.9 m`, and the quoted fractions (0.9157 / 0.722 / 0.3766) all reproduce. `refractionScale(0.0)` is exactly 1.0, so sea-level stations are untouched by the scale. Four things needed fixing, and one gap needed closing.

## 1. The fixed-point inversion was not converged near the horizon

`refractionTrueToApparent` ran exactly 5 iterations. The contraction rate of that iteration is `|dR/dh|`, which is ~0.03 at 5° elevation but **0.21 at the horizon**, so five steps is far too few down there. Measured closure error of the two directions:

| true elevation | −0.5° | 0° | 0.5° | 1° | ≥3° |
|---|---|---|---|---|---|
| closure error | **1.08″** | 0.31″ | 0.084″ | 0.023″ | <0.001″ |

The docstring claimed "well below 0.001 arcsec", and `testScaledFormulasStayInverse` only sampled 5°–85°, so nothing caught it. That is exactly where the celestial-grid horizon line and low-elevation ground targets sit.

Replaced with Newton's method: `bennettRefraction()` now returns the refraction **and** its analytic derivative, which costs a few operations once the tangent is in hand. Closure is now at double precision (7.7e-11″) from −0.5° to 90° at any scale.

## 2. Performance

Three clean rebuilds, same machine, `cyraDecToXY` — the inner loop of `fitAstrometry`, and so of the nightly recalibration:

| kernel | closure @ 0° | µs/star |
|---|---|---|
| `prerelease` (Saemundsson) | 3.42″ | 1.494 |
| this PR as submitted | 0.311″ | 1.719 (+15.1%) |
| with Newton | 0.000″ | **1.617 (+8.2%)** |

Newton needs 2.45 refraction evaluations on average instead of 5. The `precompute_pointing_corr=False` path of `cyXYToRADec` is unaffected — refraction is 1.2% of it, it is dominated by `equatorialCoordAndRotPrecession`.

Also hoisted `refractionScale(platepar.elev)` out of the per-point loop in `xyToRaDecPP` and out of the repeated call sites in `AddCelestialGrid` / `SkyFit2`.

## 3. `isinf`

Dropped the `int isinf(double)` extern. C99 makes `isinf` a type-generic macro rather than a function, so declaring it is not portable, and the test was redundant — infinity already satisfies `target_height > 1e6`.

Also moved the tropopause clamp above the height difference in `refractionTargetFraction`, and documented that the closed form is only valid for a target **above** the observer: below it the mean refractivity along the path exceeds the one at the observer and the fraction comes out negative, which the `height_diff <= 1` guard was silently masking as 0. That matters, because a ridge-top camera looking down at a landmark is the common geo-point geometry.

## 4. Ground-based objects — the gap this opened up

The good news first: `xyHt2Geo`, `geoHt2XY`, `geoHt2XYInsideFOV`, `switchToGroundPicks` and the SkyFit geo drawing all force `refraction = False`, so `refraction_scale` is inert on every one of them. I diffed a `prerelease` worktree against this branch through identical platepars: `xyHt2Geo` and `geoHt2XY` come out **bit-identical** at sea level and at 2610 m. Nothing in this PR broke the ground paths.

But two places were still treating a ground target like a star, and since this PR is where the refraction model is being made honest, they are fixed here:

**The fit.** `GeoPoints.update()` builds a geo point's RA/Dec from a pure geometric ECI difference vector — no refraction, correctly — but `fitPickedStars` → `fitAstrometry` uses `self.platepar` with the user's `refraction` flag, which defaults to on. The plate then refracts the landmark as if it were a star, and the fit absorbs the difference. Added `ApplyAstrometry.targetRaDecToPlateRaDec()`: the plate maps an input through `T(·, scale)`, the target is seen at `T(true_dir, fraction*scale)`, so the input that makes the plate reproduce the arrival direction is `T⁻¹(T(true_dir, fraction*scale), scale)`. With `fraction = 0` — a ground reference, too close for any refraction to accumulate — that is the plate's own refraction taken back out. Synthetic camera at 2610 m pointing at 20°, worst error over the frame:

| fit references | ground measurement | sky measurement |
|---|---|---|
| stars only (reference) | 0.03″ | 0.03″ |
| geo points, true direction | **163.81″** | **165.07″** |
| geo points, plate coordinates | 0.03″ | 0.03″ |
| stars + geo, true direction | **19.47″** | **19.62″** |
| stars + geo, plate coordinates | 0.03″ | 0.03″ |

The resulting plate is an ordinary refraction-on plate: meteors come out as true J2000 through `xyToRaDecPP`, ground picks through `switchToGroundPicks`, both from the same plate. This only closes exactly because the two refraction directions are now true inverses — with the old formula pair the cancellation would have left ~4″. The `refraction_fraction` keyword is there so the meteor and contrail paths can pass `refractionTargetFraction()` later; nothing does yet.

**The ground projection.** `xyHt2Geo`, `geoHt2XY` and `geoHt2XYInsideFOV` only cleared `platepar.refraction`. That stops the points being refracted, but it leaves the reference pointing at the *true* direction of the camera axis while the camera looks along the *apparent* one, so the whole field is offset by the refraction at the FOV centre — **194″ on a camera pointing at 20°**. `switchToGroundPicks()` has done this correctly for the manual-reduction ground picks since 2021, so they now use it. Displacement of the projected ground position, target 3 km above the station:

| FOV centre altitude | 60° | 45° | 30° | 20° | 10° |
|---|---|---|---|---|---|
| shift | 0.8 m | 2.4 m | 9.4 m | 38.9 m | **642.4 m** |

Separate and pre-existing, **not** touched here: `xyHt2Geo` converts through the geoid while `geoHt2XY` converts through the WGS84 ellipsoid, so the two are not exact inverses to better than a few pixels. Verified independent of refraction — 2.774 px with it on, 2.779 px with it off on the same synthetic camera.

## 5. Style, and the test file

- Named the keyword `refraction_scale` throughout. It was `scale` on the refraction routines and `refraction_scale` on the `cy*` kernels, so keyword calls into the two halves of the API were not interchangeable.
- Documented the new keyword on `pyRefraction*` / `eqRefraction*`, gave the ISA helpers the repository's `Arguments:` / `Return:` structure, and fixed the pre-existing `pyRefractionTrueToApparent` docstring, which had the two directions the wrong way round. Moved the justification paragraphs out of the docstrings into the commit messages, and brought the comment density up to the module's.
- Restored the continuation layout in `ApplyAstrometry` (the added argument sat at an 8-space indent under calls whose other continuations are at 12, and dropped the trailing `\`), and wrapped the long lines: `AddCelestialGrid.py` has a hard 110-character ceiling and had a 179-character line, `AutoPlatepar.py` 161, `SkyFit2.py` 185. No added line is over 108 now.
- `Tests/test_refraction.py` → `Tests/TestRefraction.py`, converted from the legacy `unittest` style to the pytest style every recent test in `Tests/` uses. Four cases became 26, including the closure test extended down to −0.5° (the regression guard for item 1) and the new geo-point fit and ground-projection tests. Dropped the `removeAberration` try/except — that symbol does not exist in the kernels, so the guard always took the `None` branch. Whole file runs in 2.5 s.

## Still worth adding to the PR description

- Ground picks at high-altitude stations shift until the next recalibration: `switchToGroundPicks` preserves `alt_centre`, which `updateRefAltAz` now computes with the scaled refraction, so an existing platepar applied with the new kernel moves by 24.6″ for USV003 at 2610 m pointing at 35° (1.8″ at sea level, from the exact inverse). It vanishes after a refit, because `alt_centre` is then the camera's real geometric pointing and is model-independent — same mechanism as the star case already described.
- Several call sites that previously relied on the `refraction=True` default now pass `refraction=platepar.refraction` (`AddCelestialGrid` ×5, `AutoPlatepar`, `EventMonitor.platepar2AltAz`, `SkyFit2` ×7). That is a genuine consistency fix — `addEquatorialGrid` was refracting the grid while `raDecToXYPP` honoured the flag — but it changes the drawn overlay for any platepar with `refraction = False`, which is what every legacy v1 platepar loads as. Worth naming.
- The exact-inverse change also fixes a real round-trip bug: `updateRefAltAz()` followed by `updateRefRADec()` used to drift by up to 4″ per round trip, because Bennett and Saemundsson are not inverses of each other.

dvida:
@Cybis320 — follow-up to my review above. I have pushed one more commit and rewritten the PR description to cover everything that is now in the branch. **Could you review my changes?** Five commits of mine sit on top of your three, and the last one reaches outside the refraction model into the geodesy, so I would like your eyes on it before this goes in.

## What changed since the last comment

The remaining item I had flagged as "pre-existing, out of scope" turned out to be worth fixing, and it is larger than I first measured.

`AEGeoidH2LatLonAlt()` — what `xyHt2Geo()` projects a pixel onto the ground with — took the range to the target as `r = (h - alt)/sin(elev)`. That is a flat-Earth range: it ignores the Earth curving away under the line of sight, so the point lands too high, by more the lower the camera looks.

| elevation | 60° | 45° | 30° | 20° | 10° | 5° |
|---|---|---|---|---|---|---|
| target at 3 km | +0.2 m | +0.7 m | +2.1 m | +5.3 m | +22.7 m | +92.3 m |
| target at 100 km | +258 m | +773 m | +2320 m | +5836 m | +24833 m | **+100315 m** |

At 5° elevation a target asked for at 100 km was being placed at 200 km. I had quoted the `xyHt2Geo`/`geoHt2XY` round-trip disagreement as 2.8 px from a camera pointing at 20°; on one pointing at 25°, whose frame reaches lower, it is **3931 px** — the point leaves the image entirely. It is 0.22 px now.

`Conversions.AEH2Range()` already solves this properly, so it is now used. Its default analytical path is untouched, since `EventMonitor`, `ShowerAssociation` and `FOVArea` depend on it, but the `accurate=True` path had to be rewritten, because it could not be used as it stood:

- it minimised the *square* of the height residual with `scipy.optimize.minimize` instead of solving for the root;
- its starting guess compared an elevation in degrees against `np.radians(10)`, so that branch never fired and the guess was always 1000 km;
- the analytical solution it starts from returns the **far** intersection for a target below the observer — 1099 km for a point 1.2 km below the camera at −5° elevation, instead of 14 km. A ridge-top camera looking down at a landmark is the common geo-point geometry, so that case has to work.

It now solves the sphere intersection in closed form, picking the near crossing when the target is below the observer and the far one when it is above, then refines on the WGS84 ellipsoid with Newton's method, using the component of the line of sight along the local vertical as the derivative. Within a millimetre of the requested height at every elevation from −10° to 60°, and NaN when the line of sight never reaches that height at all — which the flat-Earth formula used to answer with a point behind the camera.

New `Tests/TestGeoConversions.py`, 17 cases: the solved range against the height it actually reaches, the downward-looking near crossing, the unreachable case, the unchanged analytical path, and the round trip.

## The five commits

1. **Converge the refraction inverse, drop the `isinf` extern, align the kernel API.** Your fixed 5 iterations left 1.08″ of closure error at −0.5° and 0.31″ at 0°, because the fixed point contracts at `|dR/dh|` — ~0.03 at 5° but 0.21 at the horizon. Replaced with Newton on `bennettRefraction()`, which now returns the analytic derivative too: 8e-11″ everywhere, and 2.45 refraction evaluations instead of 5, so `cyraDecToXY` costs 1.617 µs/star rather than 1.719 (1.494 before this PR).
2. **Evaluate the refraction scale once per call site instead of per point.** Plus continuation-indent and line-length repairs.
3. **Rewrite the refraction test in the current `Tests/` style.** `unittest` → pytest, 4 cases → 26, closure extended to −0.5°.
4. **Never apply the star refraction to ground-based references or measurements.** New `targetRaDecToPlateRaDec()`; geo points are fitted through plate coordinates; the three ground helpers use `switchToGroundPicks()`.
5. **Solve for the range to a target height instead of assuming a flat Earth.** The commit above.

## What I would most like you to check

- **The `refraction_fraction` contract** in `targetRaDecToPlateRaDec()`. `input = T⁻¹(T(true_dir, fraction·scale), scale)`, with `fraction = 0` for ground. It is the hook for wiring `refractionTargetFraction()` into the meteor and contrail paths, so if the algebra does not match what you had in mind for Janus, better to find out now.
- **The near/far root selection** in `AEH2Range(accurate=True)`, particularly for a camera looking down.
- **Whether changing `AEH2Range`'s `accurate=True` behaviour is acceptable to you.** Nothing in the repo calls it with that flag today, which is why I felt free to rewrite it, but you may have external code that does.
- The NaN return for an unreachable height — a behaviour change, previously it returned a point behind the camera.

Everything is on `refraction-model` (rebased onto current `prerelease`, which the branch was behind). 92 tests pass, clean rebuild, `git diff --check` clean.

Cybis320:
@dvida — reviewed your five commits, checked each of the four items independently, and pushed one more commit (300c5b0b) that fixes the one real problem I found and removes a performance regression.

## The four items

- **`refraction_fraction` contract: correct.** On your synthetic camera I fed the true direction of a target through `targetRaDecToPlateRaDec()` and back through the refraction-on plate for fractions 0, 0.31 (10 km), 0.92 (100 km) and 1, on cameras pointing at 20°, 5° and 2°, down to 0.3° apparent altitude: the plate lands on the target pixel to 0.00 px in every case. The frame is consistent too, since `eqRefraction*` precess J2000 to date exactly as `cyraDecToXY` does. This is the algebra Janus needs; the measurement side just has to hand the kernel `fraction * scale` as its refraction scale, which the keyword already allows.
- **Newton inverse:** closure swept from −2° to +2° true elevation at five scales, worst 2e-12″. The derivative kink at the −0.5° clamp cannot trap it, the residual is monotone with slope ≥ 1.
- **`accurate=True` behaviour change: fine.** Nothing outside RMS calls `AEH2Range` at all. Janus only uses `xyHt2Geo`, which is where the change actually lands (see performance below).
- **NaN for an unreachable height: fine, and narrower than it sounds.** A ray from inside the sphere of the target's height always exits it, so for any target *above* the station a finite range always exists. 285,264 rays from four stations at −10° to 89° elevation, targets from 1 m above the station to 100 km: zero NaN. NaN can only occur for a target below the station, which no meteor or contrail path can hit.

## Near/far root selection: right choice, wrong seed

The sphere step used geocentric radii, so its "elevation" was the geocentric one, up to 0.19° off the geodetic elevation being passed in. Within a few tenths of a degree of the horizon, for targets within a couple of hundred metres of the station height, the seed landed on the wrong side of the tangent point, and since `r <= 0` was only checked *before* the Newton loop, bad seeds leaked out as answers. Swept against a brute-force root finder over |elev| ≤ 0.5° and |h − alt| ≤ 200 m, 441 cases per station:

| station | negative range returned | non-converged range returned | false NaN |
|---|---|---|---|
| 31.69° N, 2610 m | 8 | 0 | 30 |
| 45° S, 0 m | 8 | 30 | 0 |
| 60° N, 1200 m | 10 | 0 | 31 |

That is exactly the distant-ridge-on-the-horizon geometry of a ground reference. Outside that regime (|h − alt| ≥ 250 m or elevation beyond ±0.5°) I found no failures in 720 cases, so meteors, contrails and ordinary ground targets were fine.

**Fix:** seed from the sphere osculating the ellipsoid at the observer in the azimuth of the line of sight (Euler's radius from the meridional and prime vertical radii), so `b = rs·sin(elev)` with the geodetic elevation; then accept a range only if it is in front of the camera *and* lands within 1 mm of the requested height after the refinement, NaN otherwise. Same sweep: zero failures at all three stations. A 1248-case comparison with brute-force roots over −10° to 89° and heights from 2 km below the station to 100 km matches to 5 cm.

## Performance

`xyHt2Geo` had gone from 28 µs/pixel on `prerelease` to 88 µs/pixel, because the accurate path was a scalar Python Newton loop per pixel. Janus batches whole flights through `xyHt2Geo`, where that loop was already the dominant cost of its save phase. The solve is now `rangeToHeightWGS84()`, vectorized over the line of sight and the height, with `latLonAlt2ECEF()` / `ecef2LatLonAlt()` made array-safe (scalars still return floats), `AEGeoidH2LatLonAlt()` taking arrays, and `xyHt2Geo()` projecting the whole pixel array in one call via `cyTrueRaDec2ApparentAltAz_vect()`. 5000 pixels of a 720p plate:

| | µs/pixel |
|---|---|
| `prerelease` | 28 |
| e59c0fe1 | 88 |
| 300c5b0b | **4.4** |

## Tests

`Tests/TestGeoConversions.py` gains the grazing cases on both sides of the tangent point, a grid straddling the horizon that must never return a negative or off-height range, the array forms against the scalar calls, and array ECEF round trips. 129 passed, 3 skipped across `Tests/`, `git diff --check` clean.

Everything else in the five commits I agree with. From my side this is ready.

Cybis320:
Rebased onto #999 (which is stacked on #989), base of this PR changed accordingly. Nothing in the refraction model changed; the rebase touched three places where the two stacks meet:

- `pointingCorrection` keeps #989's own-frame centre refraction (`cyraDec2AltAz` → `refractionTrueToApparent(alt, refraction_scale)` → `cyaltAz2RADec`) with this PR's station-height scale.
- `Platepar.updateRefAltAz` / `updateRefRADec`, `platepar2AltAz` and the SkyFit2 alt/az line use #999's true-of-date helpers, now with `refraction_scale=refractionScale(elev)`; the helpers gained that keyword.
- `geoHt2XY` / `geoHt2XYInsideFOV` carry both changes: `switchToGroundPicks()` from here and the plate evaluated at `platepar.JD` from #999, since `geoHt2RaDec` now returns J2000. `GeoPoints.update` goes through `j2000FromTrueOfDate` and then `targetRaDecToPlateRaDec`.

One extra commit fixes `Tests/TestReferenceFrame.py` to pass the station-height scale to `pointingCorrection`, which it did not need before this PR. Rechecked on the rebased tree: the `refraction_fraction` contract reproduces the target pixel to 0.00 px for fractions 0, 0.31, 0.92 and 1 on cameras at 20° and 5°; the derived alt/az centre matches the kernel's reference direction to 0.00″ on USV001 and UK009C; the range-solver sweep has zero failures; `xyHt2Geo` is 4.1 µs/pixel. Full `Tests/`: 146 passed, 3 skipped.

Cybis320:
Re-rebased onto the updated #999 (which now folds in #987: `Platepar.computeRefAltAz()` and the two SkyFit2 pointing sites). Same ten commits; the only merges were the SkyFit2 and EventMonitor lines where this PR adds the station-height scale, kept on top of `computeRefAltAz()`. Rebuilt, 147 passed, 3 skipped.

PR #992, opened 2026-09-14
Disable panning while coloring photometry in manual reduction

dvida:
## Problem

In SkyFit2 manual reduction mode, holding **Shift** puts the cursor into photometry coloring mode (`CursorItem.mode == 2`) and dragging paints photometry pixels (left = add, right = remove). The pyqtgraph `ViewBox` treated the same drag as a pan, so the image slid under the cursor while the user was trying to color the meteor pixels.

## Fix

Added `PlateTool.updatePanningEnabled()` — a single predicate that drives the existing `ViewBox.panning_enabled` flag. Panning is blocked when:

- the Mask tab is open (drags are polygon/brush editing),
- the mask brush is active,
- `mode == 'manualreduction'` and `cursor.mode == 2` (Shift photometry coloring).

It is recomputed:

- at the top of `onMousePressed` — `ViewBox.mousePressEvent` emits `sigMousePressed` *before* it checks `panning_enabled` and the connection is direct, so the decision applies to that very press. No stale or stuck state is possible.
- after `cursor.setMode(2)` / `setMode(0)` in the key handlers, so Shift pressed or released mid-drag is handled.
- in `_exitBrushMode`, instead of relying only on the tab-change handler.

With `panning_enabled = False` the ViewBox accepts the press itself, which makes it the scene's `mouseGrabberItem`. pyqtgraph's `GraphicsScene.mouseMoveEvent` then skips `sendDragEvent`, so neither the left-drag pan nor the right-drag zoom is generated. `sigMouseMoved` still fires unconditionally, so `onMouseMoved` keeps painting.

No changes were needed in `RMS/Routines/CustomPyqtgraphClasses.py` — the `panning_enabled` mechanism was already there, previously used only by the mask tab.

## Testing

Tested manually in SkyFit2 manual reduction mode:

- Shift + left-drag colors photometry pixels, the view no longer pans.
- Shift + right-drag removes pixels, the view no longer zooms.
- Releasing Shift restores normal panning and right-drag zoom.
- Mask tab polygon and brush editing unaffected; star picking (`Ctrl+R`) unaffected.


dvida:
Good catch — confirmed and fixed in ae48e18a.

`panning_enabled` was only consulted in `mousePressEvent`/`mouseReleaseEvent`. That is enough when the press itself happens with panning already disabled (the ViewBox accepts the press, becomes the scene's `mouseGrabberItem`, and pyqtgraph's `GraphicsScene.mouseMoveEvent` then skips `sendDragEvent` entirely). But if Shift is pressed *after* the button is already down, pyqtgraph has already selected the ViewBox as the drag item and keeps calling `mouseDragEvent` on it, so the image panned while pixels were being painted at shifting coordinates.

Rather than deferring the coloring to the next drag, the flag is now a hard lock — the ViewBox gates the drag itself:

```python
def mouseDragEvent(self, ev, axis=None):
    if not self.panning_enabled:
        ev.ignore()
        return
    super().mouseDragEvent(ev, axis=axis)
```

So an in-flight pan stops the moment Shift goes down and coloring takes over from the current cursor position. This also makes the mask-tab lock robust instead of relying on the mouse-grabber side effect.

The reverse case is safe too: releasing Shift mid-drag cannot resume a pan, because with panning disabled at press time pyqtgraph never created a drag for that press at all.

PR #994, opened 2026-09-17
Clip cosine in angularSeparation to avoid NaN for coincident directions

AstroEloy:
`angularSeparation` in `RMS/Math.py` took `arccos` of the spherical cosine without clipping. For coincident (or nearly coincident) directions, rounding can push the cosine just above 1, so `arccos` returns NaN (~3% of 2M random coincident pairs).

The Cython `angularSeparation` in `RMS/Astrometry/CyFunctions.pyx` has the same unclipped formula.

## Why this matters

The trigger needs the two directions to agree to within ~2.1e-8 rad (0.0043"), which is the float64 `arccos` resolution floor. Real sky residuals are arcseconds, so this is not something a station hits during a normal night — it shows up with synthetic or self-consistent platepars, and as a latent hazard wherever the separation of two nominally identical points is computed.

Where it does bite, most of the damage is silent rather than loud. On `prerelease`:

- `Platepar.py:896` `_calcSkyResidualsDistortionVect` and `:1019` `_calcSkyResidualsAstroAndDistortionRadialVect` return `angularSeparation` straight into `scipy.optimize.least_squares`, so a single NaN aborts the fit with `ValueError: Residuals are not finite in the initial point`. That one at least fails loudly.
- `Platepar.py:1078` — `nn_distances = np.min(sep_matrix, axis=1)`; one NaN makes `total_cost` NaN at `:1079`.
- `Platepar.py:1505-1509` — `median_sep = np.median(nn_seps)` becomes NaN, so every `nn_seps > per_star_threshold` comparison is `False` and **outlier rejection is silently disabled for that whole iteration**.
- `Platepar.py:1632-1634` — `np.argmin(sep_matrix, axis=1)` returns the NaN column, so the "nearest" catalog star is whichever NaN appears first in the row.
- The scalar `np.sum(...**2)` variants feeding Nelder-Mead would just have gone NaN quietly.
- In the Cython copy, the callers filter stars by radius instead of feeding an optimizer, so a NaN silently drops a star at the FOV centre (`CyFunctions.pyx:176`) rather than aborting a fit.

## Fix

Clip the cosine to `[-1, 1]` before `arccos` in both copies. `angularSeparationVect` already does this. Results don't change when the cosine is already in range. The Cython fix is @Cybis320's 93137d85, cherry-picked with his authorship.

The Cython copy uses explicit comparisons rather than `fmax(-1.0, fmin(1.0, x))`, because C `fmin`/`fmax` return the non-NaN operand and would have turned a NaN input into a bogus 0 or 180 degrees.

## Checked

- NumPy version: 0 NaN over 2M random coincident pairs (previously ~3%). Bitwise identical to the old code over 2M general pairs. Scalar input, degree wrapper and antipodal points (pi) return the same values as before. NaN coordinates still propagate as NaN.
- Cython version, rebuilt: 0 NaN over 200k coincident pairs (previously 4.9% with the current build). Identical results to the old build on 20k general pairs, antipodal points still give exactly 180 degrees, and it agrees with the NumPy version to 1e-13 deg.
- Regression tests added in 86d8bcff (`Tests/TestMath.py`, plus a working case in `RMS/Math.py testAngSeparationDeg`). Reverting the fix and rebuilding makes three of them fail; restoring it makes all six pass.

## Scope

This PR clips `angularSeparation` only. The same unguarded `arccos` pattern exists in about eight other places in the codebase; those are handled separately in #997.


Cybis320:
Reviewed and cherry-picked onto `alpha2` (4eb98743) for station testing. The reasoning holds up — I re-measured each claim:

- 3.04% NaN over 2M random coincident pairs before, 0 after. Largest separation returned for a coincident pair after clipping is 2.1e-08 rad, i.e. the float64 `arccos` resolution floor, so the clamp costs nothing in accuracy.
- Bitwise identical to the old code on 2M general random pairs; antipodal still returns exactly π; scalar input still returns `np.float64` with the same value. `angularSeparationDeg` inherits the fix.
- No measurable cost: 0.354 s vs 0.361 s for 5 passes over 2M points. Worth confirming because `Platepar.py:1077`, `:1503` and `:1631` call it on full detection×catalog separation matrices.
- The failure path is real: `_calcSkyResidualsDistortionVect` (`Platepar.py:896`) and `_calcSkyResidualsAstroAndDistortionRadialVect` (`:1019`) return `angularSeparation` straight into `least_squares`, so one NaN aborts the fit. The scalar `np.sum(...**2)` variants would just have gone NaN quietly.

One thing this PR doesn't cover: `RMS/Astrometry/CyFunctions.pyx:99` carries the identical unclipped formula in C (`acos(sin·sin + cos·cos·cos)`), and the same arithmetic in float64 puts the cosine above 1 for 3.2% of coincident pairs. Its callers (the catalog radius filter at `:176`, the radius computation at `:1145`) don't feed an optimizer — a NaN there just fails the `<= radius` test and silently drops a star sitting at the FOV centre — so it's latent rather than breaking, but it's the same bug.

I've clipped it on `alpha2` as well (93137d85), rebuilt and checked the compiled function: 0 NaN over 200k coincident pairs (max 1.2e-06 deg), antipodal exactly 180, and agreement with the NumPy version to 1.7e-13 deg over 2000 general pairs. Happy to fold that into this PR instead if you'd rather keep the two copies fixed together.

AstroEloy:
Thanks for the thorough check. I've cherry-picked your 93137d85 into this PR (4b5c14f0) so both copies are fixed together. I also reproduced the Cython case locally before picking it: 4.9% NaN over 200k coincident pairs with the current build, 0 after rebuilding with your change, and identical results on general pairs.

dvida:
Reviewed in depth. The fix is correct, minimal and safe, and I recommend merging it. I re-derived every claim rather than taking the measurements at face value, and pushed the missing regression tests to this branch (86d8bcff). I also rewrote the PR description, because it was understating the case; details below.

## Verification

| Check | Result |
|---|---|
| NaN rate for coincident pairs, before | 2.82% of 2M |
| NaN rate after the clip | 0 of 2M |
| Largest separation returned for a coincident pair after the clip | 2.107e-08 rad, i.e. the float64 `arccos` resolution floor — the clamp costs nothing in accuracy |
| General random pairs | **bitwise identical** to the old code over 2M pairs |
| Antipodal | exactly pi |
| Scalar input | still `np.float64` |
| dtype | float32 in, float32 out; `np.clip` with Python-float bounds does not upcast |
| NaN *input* | still NaN in both copies — the clip does not mask bad data |
| Compiled extension as currently shipped | `angularSeparation(0.0, -82.0, 0.0, -82.0)` returns `nan`, so the bug is live in the binary, not just the source |

Two things worth calling out as deliberately right, not accidental:

- The Cython version uses `if cos_sep > 1.0 / elif cos_sep < -1.0` rather than the tidier-looking `fmax(-1.0, fmin(1.0, x))`. That matters: C `fmin`/`fmax` return the non-NaN operand, so the tidier form would have silently converted a NaN input into a bogus 0 or 180 degrees. The chosen form propagates NaN and matches the NumPy side.
- `angularSeparationVect` at `RMS/Math.py:96` already clipped, so this makes the three Python entry points consistent.

## Tests — pushed as 86d8bcff

`AGENTS.md` asks for regression coverage under `Tests/`, and the measurements in this thread, thorough as they are, are not committed and there is no CI to re-run them.

There was also a trap here. `RMS/Math.py testAngSeparationDeg()` **already** contained a coincident case, `[[45, 45], [45, 45], 0]` — but `(45, 45)` lands on a cosine of exactly `1.0`, so it passed with the bug present. Anyone reading that test would reasonably conclude the case was covered.

The commit adds:

- `Tests/TestMath.py`: a deterministic coincident pair `(0, -82)` that actually trips the domain error, a 1-degree coincident grid (2160 of those pairs used to give NaN), an antipodal check, a NaN-input propagation check, and a Cython-vs-NumPy agreement check — the last being the only thing that would catch a stale or unrebuilt `.so`.
- `RMS/Math.py`: `[[0, -82], [0, -82], 0]` added to `testAngSeparationDeg`, with a comment explaining why the `(45, 45)` case does not cover it.
- Expanded comments in both files, so the clamp does not get deleted later as "redundant".

I verified the tests actually gate the fix: reverting both source files to the merge-base and rebuilding makes three of them fail, and restoring the fix makes all six pass.

## Description rewrite

The original text led with `ValueError: Residuals are not finite in the initial point` from a platepar refit. I could not reproduce that, and I do not think it is reachable with real data — the trigger needs the two directions to agree to ~0.0043", while real sky residuals are arcseconds.

But the PR was also selling itself short. The nearest-neighbour matching paths on `prerelease` fail **silently**, which is worse than an exception. In particular `Platepar.py:1505-1509`: `median_sep` goes NaN, every `nn_seps > per_star_threshold` comparison is then `False`, and outlier rejection switches itself off for the entire iteration with no error and no warning. That is the strongest argument for merging this, and it was not mentioned. I have put it in the description.

## Build and scope

- Recythonized and rebuilt with Cython 3.2.4, clean. `Scripts/RMS_Update.sh` deletes every `.so` and wipes `build/` before reinstalling, so the Cython half does reach stations on the next update.
- Unrelated, for the record: `python -m RMS.Math` reports a failure in `testPointInsideConvexPolygonSphere` on this branch **and** on the merge base, so it is pre-existing and not caused by this PR. Also, `tests()` prints `Test failed` without naming which test failed, which is worth tidying sometime.
- The same unguarded `arccos` pattern exists in roughly eight other places in the codebase, including one more in `CyFunctions.pyx`. Rather than widening this PR, I have handled them in #997.

Thanks @AstroEloy and @Cybis320 — good catch, and the cross-checking in this thread was exactly right.

PR #995, opened 2026-09-17
SkyFit2: overlay WMPL trajectory solutions

AstroEloy:
## Summary

Adds a **display-only overlay of WMPL trajectory solutions** to SkyFit2, so a solution can be checked directly against the image or video being reduced.

## WMPL solution overlay

- **Input:** the `Points` table of a WMPL report file (`*_report.txt`). The report is plain text and doesn't need `wmpl` installed, unlike the trajectory pickle. Columns are found by their header names.
- **Projection:** each point is the position on the trajectory of one observation. It is projected from the current station with the current platepar through the same path as the geo points (`WMPLSolution` subclasses `GeoPoints`). Refraction is applied, as for catalog stars, because these are true meteor positions. The heights in the report are MSL, like the platepar elevation.
- **Redraw:** points are redrawn in `updateStars`, so they follow any platepar change and, in ManualReduction, the current frame.
- **Isolation:** the markers sit at z = 0.5 (above the image, below every other overlay) and don't accept mouse buttons. The legend is a non-interactive text item. The solutions are excluded from the saved state.
- **Settings tab:**
  - "Show WMPL Solutions" and "Show Legend" checkboxes
  - a checkable list of the loaded solutions (it takes no space when empty)
  - "Load WMPL Report" and "Remove Selected Solution" buttons
- **Markers and colours:** up to 5 solutions at a time, each with its own marker (○ □ ▽ ◇ ☆). Each station keeps the same colour across solutions.
- **Legend:** a single legend in the top-right corner of the view, so the legends of overlapping solutions can't overlap.
- **Help:** the Settings help page documents the new section.

The new section makes the Settings tab taller. On short windows the tab relies on the scroll areas from #918.

## Testing

- **Projection:** on a multi-station event, the projected points of the station being reduced lie within 0.18 px (median) of its manual picks; without refraction the median is 0.71 px. The other stations' points fall along the same track.
- **Controls:** tested in SkyFit and ManualReduction modes:
  - the show/hide toggles, the legend and the per-solution checkboxes
  - the 5-solution limit and its warning
  - removing a solution and reloading
  - an invalid file
  - window resizing and frame changes
  - saving the state


Cybis320:
Heads-up on the scroll half of this: the same change is already in #918 (`improve-star-detection` → `prerelease`), commit 34c2511f5 ("Snap max stars sliders to 100 and make tab panels scroll when short"). It got picked up there because the Star Detection tab reorganization is what made the panels tall in the first place.

Two differences worth noting, since both PRs target `prerelease` and will conflict in `RightOptionsTab.__init__`:

- #918 wraps all six form panels (Fit Parameters, Station, Star Detection, Mask, Settings, Debruijn), not just Fit Parameters and Settings.
- It also sets `panel.setSizePolicy(QSizePolicy.Ignored, ...)` on each wrapped panel. The tab column has a fixed width, and a plain `QScrollArea` enforces the panel's *minimum* width and clips the right edge (the corner help button first) instead of compressing as before. With `ScrollBarAlwaysOff` horizontally the clipped part is unreachable, so `scrollable()` here may show that on a narrow tab column.

The WMPL overlay is unrelated and isn't covered anywhere else — suggest dropping the scroll commit content from this PR and letting #918 carry it, or the reverse if this one lands first, but in that case #918's version is the one to keep.

AstroEloy:
Thanks, agreed. I've dropped the scroll change from this PR (2f3e962b) so #918 carries it; this PR is now only the WMPL overlay. One note: the `setSizePolicy(QSizePolicy.Ignored, ...)` you mention isn't in the pushed #918 branch yet (head 024f2ae1), only the scroll wrappers are, so it may still need pushing.

PR #997, opened 2026-09-19
Clamp the remaining arccos arguments into the valid domain

dvida:
Follow-up to #994, which clamped the cosine in `angularSeparation`. The same defect exists in several other routines: they take the arc cosine of a quantity that is mathematically confined to `[-1, 1]` — a dot product of unit vectors, a spherical law of cosines, a direction cosine — without guarding the domain. Rounding pushes such a quantity just outside the interval for degenerate geometries, and `arccos` then returns NaN instead of 0 or pi.

This PR is Python and Cython only. No C or C++ code is touched.

## The one that is reachable in practice

`ECEF2AltAz` (`RMS/Astrometry/Conversions.py:464`). For a point directly above or below the station the normalized dot product is mathematically exactly 1, and rounding puts it above 1 for a large fraction of station positions:

```
zenith NaN rate, old formula: 3869/20000 = 19.34%
```

The altitude comes back NaN instead of +-90 degrees. Callers are `RMS/EventMonitor.py:695-697` and `:742`, which use it to decide whether a fireball trajectory is visible from a station, and `Utils/FOVSkyMap.py:340`.

## The rest — hardening, not observed failures

I checked each of these and could not produce a NaN from realistic inputs, but the arithmetic allows it and the guard costs nothing:

| Location | Quantity |
|---|---|
| `RMS/Math.py:161` `cartesianToPolar` | `z` component of a unit vector, at the poles |
| `RMS/Formats/AST.py:215`, `:226` | FOV width and aspect; `:226` divides by the result |
| `RMS/Formats/AllskyPlate.py:101` `AltAz2xy` | spherical cosine |
| `Utils/ShowerAssociation.py:375`, `:378`, `:381` | three meteor geometry angles, whose sines are then used as divisors |
| `RMS/Astrometry/CyFunctions.pyx:275` `equatorialCoordPrecession` | `sqrt(A**2 + B**2)` in the near-pole branch |

## Deliberately not changed

`RMS/Astrometry/ApplyAstrometry.py:100` — `np.arccos((10**(-0.4*delta_mag))**(1/4))`. This argument exceeds 1 only when `delta_mag` is negative, which is an invalid input rather than a rounding artefact. Clamping would hide the mistake, so it is left as is.

`Native/Hough/voting.cpp:264` — `acos(v.x)` on a normalized eigenvector carries the same theoretical exposure for axis-aligned lines. I rebuilt without a guard and ran 300 exactly-vertical lines through the kernel-based Hough transform: no NaN appeared, so it is not an observed problem. Left untouched.

Already clipped, no change needed: `RMS/Math.py:98` `angularSeparationVect`, `RMS/Astrometry/StarFilters.py:161`, `Utils/SkyFit2.py:119` and `:6079`.

## Style

Inline `np.clip(..., -1.0, 1.0)` matches the four places in the codebase that already do this. The Cython copy uses explicit comparisons instead, because C `fmin`/`fmax` return the non-NaN operand and would silently turn a NaN input into 0 or 180 degrees.

## Tests

New `Tests/TestArccosDomain.py`, 6 tests: `cartesianToPolar` at the poles and one ULP past unity, a round trip to confirm well-conditioned directions are untouched, `ECEF2AltAz` at the exact heights that used to fail plus 2000 random zenith points, and `equatorialCoordPrecession` at the pole. The `ECEF2AltAz` tests fail on `prerelease` and pass here.

The remaining sites need an AST plate, an all-sky plate or a full shower association run, so they are not covered directly.

## Checked

- `Tests/TestArccosDomain.py`, `TestMath.py`, `test_Grouping3Dcy.py`, `TestMatchStars.py` — 20 passed.
- Cython extension recythonized and rebuilt with Cython 3.2.4, no new warnings.


dvida:
Update on this branch, plus two things that came out of re-checking it.

## The C++ change has been dropped

The first version of this branch also guarded `acos(v.x)` in `Native/Hough/voting.cpp`. That has been reverted and the commit amended, so no C or C++ is touched here at all. The file is byte-identical to `prerelease`, CRLF line endings included.

That was the right call regardless of policy: I had already rebuilt the kernel-based Hough transform without the guard and pushed 300 exactly-vertical lines through it, and no NaN appeared. The eigenvector normalization keeps `v.x` inside the domain in practice, so there was nothing there to fix. It is now listed under "deliberately not changed" in the description.

The branch was force-pushed (single commit, `a7e46783` -> `7646e312`). Anyone who pulled it earlier should reset rather than merge.

## Verified against #994

Both branches touch `RMS/Astrometry/CyFunctions.pyx`, in different functions, so I merged them together onto `prerelease` to be sure:

- No conflicts in either direction.
- Both guards present and intact in the merged `CyFunctions.pyx`.
- Extension recythonized and rebuilt on the combined tree, then 25 tests pass across `TestArccosDomain.py`, `TestMath.py`, `test_Grouping3Dcy.py` and `TestMatchStars.py`.

Merge order does not matter.

## One honest note on what the clamp does in fully degenerate geometry

Two of these sites use the arc cosine result as a divisor: `AST.py:226` divides the FOV width by it, and `estimateMeteorHeight` divides by `sin(theta_met)`. If the geometry is genuinely degenerate — a zero-height field of view, or a meteor whose begin and end vectors coincide — the clamp gives `arccos(1.0) = 0` and the division then produces an infinity where it previously produced a NaN.

That is not a regression. The old code *already* produced an infinity in exactly that geometry whenever the cosine happened to land on exactly `1.0`, and only produced a NaN when rounding happened to push it above 1. So the practical effect of the clamp is to make the degenerate case behave consistently (always an infinity) instead of flipping between infinity and NaN depending on the last bit of a floating point result.

The underlying division by zero is pre-existing and orthogonal to this PR. It is worth a separate look at some point, but guarding it here would mean deciding what a zero-height FOV or a zero-length meteor should return, which is a different discussion.

dvida:
Recording two follow-ups that came out of this review, so they are not lost when these PRs are merged. Neither blocks the merge.

## 1. The field impact of the `ECEF2AltAz` NaN has not been assessed

This PR stops `ECEF2AltAz` returning NaN for a point directly above or below the station, which it did for about 19% of station positions. What has *not* been checked is what those NaNs were doing up to now.

`RMS/EventMonitor.py:695-697` and `:742` use that altitude to decide whether a fireball trajectory was visible from a station. A NaN elevation there does not raise — it silently fails whatever comparison it feeds. So it is worth going back over past event-monitor decisions for near-zenith trajectories and confirming none were dropped or accepted wrongly. `Utils/FOVSkyMap.py:340` is the other caller and is cosmetic by comparison.

Unlike the `angularSeparation` case in #994, this one is not a needle-in-a-haystack geometry. Straight overhead is ordinary.

## 2. There is no CI

`.github/` does not exist in this repository. Nothing builds the two Cython extensions or the C++ Hough module on a pull request, and nothing runs `Tests/`.

Everything claimed in #994 and in this PR was verified by hand on one machine, with one Python version, one NumPy and Cython 3.2.4. The regression tests added by both PRs only protect the codebase if something actually runs them.

Both of these bugs are exactly the class a minimal build-and-test job would have caught: a domain error with a deterministic reproducer, in code that compiles to a native extension where a stale `.so` can hide the problem entirely. A GitHub Actions workflow that installs `requirements.txt`, runs `pip install -e . --no-deps --no-build-isolation` and then `pytest Tests/` would be a small amount of work for a large amount of protection, and would also catch the pre-existing `testPointInsideConvexPolygonSphere` failure in `python -m RMS.Math` that both of these branches inherit from `prerelease`.

Worth raising as its own issue rather than letting it sit in a merged PR thread.

PR #998, opened 2026-09-20
SetCameraAddress non-default credentials

sanlupkim:


Cybis320:
Heads up: this overlaps with #922 (cli-tools-argparse), which also rewrites the `__main__` block of `Utils/SetCameraAddress.py` to argparse. #922 drops the manual `len(sys.argv) < 3` usage printout and moves that text into the parser's epilog, with `CURRENT_IP` / `NEW_IP` positionals. Both PRs replace the same lines, so whichever merges second will conflict.

One small note on the current version here: the `len(sys.argv) < 3` check runs before `parse_args()`, so `python -m Utils.SetCameraAddress -h` (or `-u admin` without both IPs) prints the old usage text instead of the argparse help that now documents `-u/-p`.

Possible ways to resolve:

1. Rebase this branch onto #922's version of the file and add just the credential bits on top: the two `add_argument` calls for `-u/--user` and `-p/--password`, plus passing `user=`/`password=` to both `DVRIPCam(...)` constructors. That is a ~6 line diff and removes the pre-parse length check.
2. Fold the two credential arguments into #922 directly and close this one.
3. Merge this first and let #922 rebase, keeping this PR's parser but dropping the manual usage block in favour of an epilog.

Option 1 or 2 seems cleanest since #922 already handles the help text. Happy to help with either.

sanlupkim:
Feel free to resolve this, I'm not up-to-date with PR/commits.

PR #1000, opened 2026-09-21
Merge master into prerelease (git credential-prompt hang fix, #985)

dvida:
First step of the September 2026 `prerelease` -> `master` release. `master` carries three commits that are not on `prerelease` (PR #985, the git credential prompt that blocked nightly processing during the 2026-09-03 GitHub outage). They conflict with `prerelease` in four files, so they are merged here first, ahead of the main release PR.

## Conflict resolution

- **`RMS/StartCapture.py`, `RMS/Reprocess.py`** - both sides added a name to the same `from RMS.Misc import ...` line and a call at the top of `__main__`. Kept both: `setMultiprocessingStartMethod()` (prerelease, must precede any `Process`) followed by `disableGitPrompts()` (master). `StartCapture.py` keeps its CRLF line endings; the diff against `prerelease` is 7 lines.
- **`RMS/Misc.py`** - no semantic overlap. Prerelease's multiprocessing helpers (`frameBufferShape`, `setMultiprocessingStartMethod`, `BoundedLock`, `AtomicFlag`) and its rewritten `runWithTimeout` are kept; master's `### Safe git invocation ###` block (`disableGitPrompts`, `runGitCommand`, host checks, URL sanitising) is appended after it. Import lines merged (`socket`, `tempfile` from master; `multiprocessing` from prerelease).
- **`RMS/Formats/ObservationSummary.py`** - the only substantive conflict. Both branches had rewritten the repository-lag check independently:
  - Master's credential-safe `RMS.Misc.runGitCommand` (returns `(returncode, stdout, stderr)`, hard timeout, prompts disabled) is used throughout. Prerelease's local `runGitCommand(command, cwd)` and `GIT_TIMEOUT_SEC` are removed, as the names collided and the signatures differed.
  - Master's `filterCloneableRemotes`, `updateCommitHistoryDirectory` (host reachability probe, per-remote fallback, raises `RuntimeError` when no remote works), `getCommit`, `getDateOfCommit` and `daysBehind` (returns `(None, None)` on failure, always cleans up the temporary clone) are taken wholesale.
  - `getRemoteBranchNameForCommit` is rewritten on top of `runGitCommand`, keeping prerelease's behaviour: drop symbolic refs such as `origin/HEAD -> origin/master`, prefer `origin/master` / `origin/prerelease`. Branches whose tip **is** the commit (`--points-at`) are checked before branches merely containing it (`--contains`), so a station exactly on the tip of `prerelease` reports `origin/prerelease` rather than `origin/master`.
  - `finalizeObservationSummary` keeps prerelease's dict-based `addObsParam(d, ...)` API and adds master's handling of the `(None, None)` case ("Not determined").
  - The `from __future__` line prerelease had dropped is restored; duplicate `import subprocess` and `from RMS.Logger import getLogger` removed.
- **`Scripts/RMS_Update.sh`** auto-merged (adds `git_net` wrapper with timeout and prompt suppression); result read through.

## Testing

- `python setup.py build_ext --inplace` builds all Cython modules.
- `python -m pytest Tests -o python_files="Test*.py test_*.py"`: 77 passed. (Note: the default pytest pattern only picks up `test_*.py`, so the `Test*.py` files need the explicit pattern.)
- Imported `RMS.Misc`, `RMS.Formats.ObservationSummary`, `RMS.Reprocess`, `RMS.StartCapture`.
- Ran the merged lag check against this checkout: `getRemoteBranchNameForCommit` returns `origin/prerelease` for the prerelease tip, `origin/master` for the master tip and `None` for an unknown hash; `daysBehind()` clones the history anonymously with prompts disabled and returns `(0.0, 'origin/prerelease')`.

### DIFF
--- a/RMS/Formats/ObservationSummary.py
+++ b/RMS/Formats/ObservationSummary.py
@@ -47,11 +47,20 @@ import tempfile
 import ephem
 
 from RMS.ConfigReader import parse
-from RMS.Misc import niceFormat, isRaspberryPi, sanitise, getRMSStyleFileName, getRmsRootDir, UTCFromTimestamp
+from RMS.Misc import niceFormat, isRaspberryPi, sanitise, getRMSStyleFileName, getRmsRootDir, \
+    UTCFromTimestamp
+from RMS.Misc import runGitCommand, GitCommandTimeout, hostReachable, gitUrlHost, \
+    gitUrlHostAndPort, sanitiseGitUrl, anonymousHttpsGitUrl, GIT_ALLOWED_HOSTS, \
+    GIT_NETWORK_TIMEOUT, GIT_LOCAL_TIMEOUT
+from RMS.Logger import getLogger
 from RMS.Formats.FFfits import filenameToDatetimeStr
 from RMS.Formats.Platepar import Platepar
 from RMS.CaptureDuration import captureDuration
 from RMS.CaptureModeSwitcher import SWITCH_HORIZON_DEG
+
+
+# Get the logger from the main module
+log = getLogger("rmslogger")
 
 
 if sys.version_info.major > 2:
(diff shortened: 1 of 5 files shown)


dvida:
Addressed the Codex finding in f7b424b3: `hostReachable()` now runs the name lookup and the connection together under `runWithTimeout` (existing helper in `RMS/Misc.py`), so a stalled resolver is cut off at `GIT_REACHABILITY_TIMEOUT` (5 s) instead of blocking in `getaddrinfo()`. A connection that completes after the deadline is closed by the late-completion hook. Added `Tests/TestHostReachable.py` (stalled lookup returns False within the deadline, late socket is closed, success and gaierror paths). Checked live: github.com reachable in 0.03 s, a blackholed address returns False at exactly the 1 s deadline, an unknown name fails fast. Full suite: 81 passed.

PR #1001, opened 2026-09-21
Prerelease update - September 2026

dvida:
> **Draft.** This is the September 2026 release of `prerelease` into `master`. It stays a draft until the review-pass PRs listed under **Review pass** below are merged into `prerelease`. The description is updated as they land.

`prerelease` is ~96 merged PRs (#762 to #997, March to September 2026) ahead of `master`: 108 files, about +26k/-5k lines. Contributors: @dvida, @Cybis320, @g7gpr, @GlassOnTin, @markmac99, @alextudorica, @sanlupkim, @novasharper, @walesto, @AstroEloy.

Headline items: a new nearest-neighbour RANSAC astrometric fit with headless auto-calibration (`AutoPlatepar`), a large SkyFit2 update (auto fit, brush masks, help tab, file manager, HiDPI, Qt6/pyqtgraph 0.13 compatibility), a set of multiprocessing deadlock and OOM fixes, NumPy 2 / Cython 3.3 / Python 3.14 support, and a per-night processing-status state machine replacing the leftover-backup heuristics.

## Operator-visible / breaking changes

| Area | Change |
|---|---|
| Config added | `time_server` (System, default `time.cloudflare.com`), `udp_buffer_size` (Capture, 16777216), `gst_queue_size` (Capture, 100), `video_scale_width` / `video_scale_height` (Capture, off), `video_crop` (Capture, off), `auto_reprocess_max_attempts` (Capture, 3; 0 = forever), `reprocess_if_archive_missing` (Capture, false), `ml_model_file` (MeteorDetection, `hyper_model.tflite`), `recalibration_min_match_fraction` (Calibration, 0.6) |
| Config removed | `kht_build_dir`, `kht_binary_name`, `kht_binary_extension` (KHT is now a Cython extension; keys are ignored) |
| Config defaults | `star_catalog_file` -> `GMN_StarCatalog`; `win_pc_weave` `-Wall` -> empty |
| Default distortion | New platepars default to `radial7-odd` (was `poly3+radial`) |
| Star catalog | Format v2 (20 columns, common and Bayer names); bundled `GMN_StarCatalog_LM9.0.bin` regenerated; v1 files still read |
| CALSTARS | Column order fixed to match the header (IntensSum before Ampltd) in writer and both extractors |
[...]

PR #1002, opened 2026-09-21
Review pass: astrometry, calibration and detection fixes

dvida:
Part of the September 2026 release review (#1001). Bug and maths fixes found while reading the astrometry, calibration, catalog and detection changes on `prerelease`. Behaviour changes are limited to confirmed defects; performance work is in a separate PR stacked on this one, and a style pass follows.

Every flagged spot was read in context first. The rejected ones are listed at the end so reviewers can see what was checked.

## Fixes

**`RMS/Astrometry/StarFilters.py`** (98dfce79)
- `filterBlendedStars` computed the FOV radius as `(diagonal/2) * F_scale` (px times px/deg), so the "stars behind the camera" pre-selection always saturated at the 90 deg cap and was a no-op. New helpers `fovRadiusDeg(platepar, margin)` and `catalogInFOVMask(ra, dec, platepar, margin)` divide by the plate scale; SkyFit2 had its own copy of the same expression and uses the helper in #1004. Tests in `Tests/TestStarFilters.py`.

**`RMS/Astrometry/StarFilters.py`** (8053fc78, found by the second review pass)
- The unit fix above exposed a second error in the same pre-selection: the cone was centred on the platepar's reference pointing `(RA_d, dec_d)`, which belongs to `platepar.JD`. A fixed camera's field moves by about 15 deg of RA per hour, so on a narrow lens (80 px/deg) the mask kept 100% of the in-image stars at the reference time, 81% one hour later, 28% after two and none after three. The old 90 deg radius had hidden this. `catalogInFOVMask` now takes the image JD and centres on the sky position of the image centre at that time; a test walks the image time up to six hours from the reference and fails with the old centring from one hour on. The RANSAC iteration callback also gets the pole-reflection position-angle offset that the fix below missed there.

**`RMS/Formats/Platepar.py`** (a8bec3a7)
- The declination wrap `-90 + (90*dec + 90) % 180.000001`, repeated at 13 sites, sent a declination of 90.000001 to the opposite pole (-90) instead of reflecting it. New module helper `normalizeRaDec(ra_d, dec_d)` reflects across the pole (`dec -> 180 - dec`, `RA += 180`) and wraps RA to [0, 360). All 13 sites use it. For values inside [-90, 90] the result is bit-for-bit the old one. Tests in `Tests/TestPlatepar.py` cover in-range, exactly +-90, +-95, 90.000001 and unit-vector equivalence.
- `np.random.seed(42 + iteration)` in the RANSAC loop reseeded the **global** NumPy RNG as a side effect of fitting. Replaced with a local `RandomState(42 + iteration)`; the draws are identical.
- `best_res.x` is now guarded by an explicit `if best_res is None` instead of relying on the `best_cost > 10` check.
- `write()` recomputed the FOV inside `except Exception: pass`. Narrowed to `(ValueError, TypeError, AttributeError, ArithmeticError)` and logged at debug (module now has a logger).
- Comment on `radial_scale` claimed it was updated with the distortion centre in the loop; it is computed once from the image centre on purpose (the early-exit "outliers stable" test relies on a fixed threshold). Comment corrected.

**`RMS/Formats/Platepar.py`** (59c930cb, from review feedback)
- Reflecting the pointing across a pole also turns the local tangent frame by 180 deg, because the direction of increasing declination at the reflected coordinates is opposite to the one carried across the pole. Keeping `pos_angle_ref` therefore flipped the field at the crossing and left the cost function discontinuous exactly where the reflection was meant to smooth it: on a synthetic field a star jumped 94 px between declination parameters 89.99 and 90.01. `normalizeRaDec` now returns `(ra_d, dec_d, pos_angle_offset)` and every call site that sets `pos_angle_ref` applies the offset; the same walk then steps 0.4 px. `testProjectionIsContinuousAcrossThePole` covers it and fails if the offset is forced to zero.

**`RMS/Astrometry/CheckFit.py`** (d4f4e34b)
- `_calcImageResidualsAstro` writes the trial parameters onto the caller's platepar on every evaluation, so when `minimize` returned without success the platepar carried the **last evaluated simplex point** and that corrupted state went on to `_handleFailure` / the NN refinement. The starting parameters are now restored right after `minimize`; the optimum is applied on success as before.

**`RMS/Astrometry/NNalign.py`** (feb1dbd9)
- `for calstars_file in file_list: ... break` left `calstars_file` pointing at the last file when nothing matched, so the `is None` check never fired. Separate loop variable, assigned on match.
- The years-since-J2000 value was built from a `datetime` where `int(round(ms*1000))` could reach 1000000 and raise. It is now computed from the JD already at hand via the new `jd2YearsFromJ2000` (see Conversions).
- Empty or one-dimensional CALSTARS entries raised `IndexError` on `shape[1]`; guarded, warns and returns the input platepar. Tests in `Tests/TestNNalign.py`.
- The inferred-LM constants (4.0, 12.0, +1.0 margin, percentile) are named `INFERRED_LM_*` and the `mag_lev != 1.0 and stddev > 0` "photometry fitted" test is explained (1.0 / 0.0 are the Platepar defaults).

**`RMS/Astrometry/Conversions.py`** (6a7e54dd)
- New `jd2YearsFromJ2000(jd)` used by NNalign and AutoPlatepar. Tests in `Tests/TestConversions.py`.

**`RMS/Astrometry/AutoPlatepar.py`** (6a7e54dd)
- The recursive wide-FOV fallback dropped `final_catalog_stars` from its return value.
- `(dt - J2000).days/365.25` truncated to whole days; uses `jd2YearsFromJ2000`.
- `saturation_fraction` actually held the non-saturated fraction (it enters the score positively, which is the right maths); renamed to `non_saturated_fraction`, including the details-dict key (no other consumers in the repo).
- Never-read `chunk_frames` and `removed` replaced with `_`; the 3.0 px association radius is now `NN_PAIR_ASSOC_RADIUS_PX` with a comment.

**`RMS/ExtractStars.py`** (f9390747)
- `extractStarsAuto` (used by the astrometry.net path) scaled the intensity threshold and the amplitude seed to the bit depth **and** passed `bit_depth` on, so `extractStars` scaled them again: 256x too high on 16-bit data. Values are now kept in 8-bit units and scaled exactly once, like every other caller.
- The `star_seg.size == 0` check ran after `np.indices` / `np.ones_like` on the empty segment; moved before.

**`RMS/Astrometry/MatchStars.py`** (c54db6ff)
- Docstring now states the semantics: `distance_upper_bound` is inclusive and no one-to-one enforcement is done (the removed Cython version had none either; the only caller, `matchStarsResiduals`, uses the pairs for residuals and counts and tolerates duplicates).

**`RMS/Formats/StarCatalog.py`** (de7d4620)
- Catalog version detection was `>= 20` columns vs everything else. Now `== 18` is v1, `== 20` is v2, anything else raises a clear `ValueError`. Tests in `Tests/TestStarCatalog.py`.
- Unknown `additional_fields` were silently dropped (they used to raise); a warning naming the unknown fields is now logged.

**`RMS/Routines/Image.py`** (0d4596bc)
- `CoordinateFilter` indexed the raw `coords` argument with the boolean mask instead of `coords_array`, failing for list input. Tests in `Tests/TestImage.py` (list, array, empty, masked).

**`RMS/Routines/Grouping3Dcy.pyx`** (768dd60d)
- `avg_std` was `unsigned int`; a negative intermediate wrapped to a huge value which the following `> 255` check then clamped by accident. Now signed with an explicit [0, 255] clamp. `Tests/test_Grouping3Dcy.py` gains a `j1 = -20` case (old code rejected, new accepts).
- `max_array_size` override: confirmed intended (the value is a minimum allocation, `propagateLine` stops when the array is full). Documented in the docstring and at the check.

**`RMS/Astrometry/AstrometryNet.py`** (2e57bbb1)
- Unconditional `print("DEBUG: fov_w_hint=...")` is now behind `verbose`.
- Star selection used the unseeded global `np.random.choice`; local `RandomState(0)` for reproducible solves.

**`RMS/Routines/SatellitePositions.py`** (e4a054a5)
- `datetime.utcnow()` (naive, deprecated) mixed with timezone-aware comparisons; both uses are now `datetime.now(timezone.utc)`. Exploratory comment ("Let's reuse its logic but...") replaced with a statement of what the code does.

**`RMS/Detection.py`** (2a8da8b6)
- No change in behaviour. The per-line `default_rng(0)` was briefly hoisted out of the loop and then restored, because each line drawing from the same seeded state is what makes a line's subsample independent of how many lines preceded it; the comment now says so.

## Checked and left alone

- `Platepar.getDistortionCentre` +0.5 px only in the forced branch: mirrors `CyFunctions.pyx` exactly (`x0 = 0.5/(x_res/2)` then `* x_res/2` in the forced case, `x_poly[0]*x_res/2` with no offset in the fitted case).
- `Platepar` `delta_Ho = (new - old) % 360`: RA is re-wrapped with `% 360` afterwards, so a signed delta would give the same RA; no jump for platepars written before `JD2HourAngle` was normalised.
- `_lstsqFit` with `method='lm'`: no bounds are passed anywhere.
- `ExtractStars` second "too many candidates" check: not dead, it is the `show_candidates` path after the diagnostic plot.
- `plotStars` / `plt.show(block=True)`: only reachable with `show_candidates=True`, which every batch caller leaves False; the CLI flag also forces sequential extraction.
- `StarCatalog` `mags != 0.0` as a missing-value test: both shipped catalogs (LM9.0 v2, LM12.0 v1) contain no exact 0.0 and no NaN in any band; missing bands are encoded as implausibly large magnitudes. The guard is inert and left in place rather than guessing a sentinel.
- `Detection._white_ratio_reject_count` per worker: already documented at the definition.
- `ApplyAstrometry` FOV radius mean -> max of corners: `subsetCatalog` has no cap, but NNalign caps at the 8000 brightest and `matchStarsResiduals` filters to on-image positions.

## Follow-up noted, not changed here

- `Conversions.date2JD` has the same latent `int(millisecond*1000)` overflow for `ms >= 1000`; not reachable from `getMiddleTimeFF`. Changing it would turn truncation into rounding, so it is left for a separate change.

## Testing

- `python setup.py build_ext --inplace` (rebuilt after the `.pyx` change).
- `python -m compileall -q RMS Utils`; imports of Detection, AutoPlatepar, Platepar, StarCatalog, NNalign, CheckFit, AstrometryNet, SatellitePositions, ExtractStars, Image.
- `python -m pytest Tests -q -o python_files="Test*.py test_*.py" --ignore=Tests/CompressionTimings.py`: 113 passed (new: TestStarFilters, TestPlatepar, TestStarCatalog, TestConversions, TestImage, TestNNalign, one Grouping3Dcy case).
- All touched files were LF and stay LF; no lines over 110 columns added.

## Reviewer notes

- `normalizeRaDec` only changes optimizer behaviour when a trial declination leaves [-90, 90]; RA is now wrapped in the same call (it already was `% 360` at every site).
- `extractStarsAuto` now finds fainter stars on >8-bit data through the astrometry.net auto-extraction path (thresholds were 2^(bit_depth-8) too high).
- `scoreFrameQuality` details key renamed `saturation_fraction` -> `non_saturated_fraction`; would break an out-of-tree script reading that key.


dvida:
Fixed in `59c930cb`. The finding is right, and the effect is larger than a tidy-up: reflecting the pointing across the pole turns the local tangent frame by 180 deg, so keeping `pos_angle_ref` left the objective discontinuous exactly where the reflection was meant to make it smooth. Measured on a synthetic 1920x1080 field at F_scale 27 px/deg with the boresight walking across the pole, a star jumped 94 px between declination parameters 89.99 and 90.01; the neighbouring steps move it 0.6 px.

`normalizeRaDec` now returns `(ra_d, dec_d, pos_angle_offset)` where the offset is 0 or 180 deg, and all 12 call sites that set `pos_angle_ref` add it (the 13th is the iteration print, which takes `_`). With the offset applied the same walk gives a 0.4 px step at the crossing. `Tests/TestPlatepar.py::testProjectionIsContinuousAcrossThePole` walks the parameter across the pole and requires every step to move the stars comparably; it fails if the offset is forced back to zero.

PR #1004, opened 2026-09-21
Review pass: SkyFit2 fixes and de-duplication

dvida:
Part of the September 2026 release review (#1001). Bug fixes and behaviour-preserving de-duplication in SkyFit2 found while reading the +11k lines added on `prerelease`. **Stacked on #1002** (it uses the `catalogInFOVMask` helper added there); the base will retarget to `prerelease` when #1002 merges. A style pass on the same files follows in a separate PR.

Every flagged spot was read in context; the ones checked and left alone are listed at the end. The WASD pan / pointing-indicator geometry was verified numerically (see "Geometry").

## Fixes

**`Utils/SkyFit2.py`** (06ea1099)
- The catalog FOV pre-selection in the star-detection tuning path computed the radius as `(diagonal/2) * F_scale * 1.5` (px times px/deg), so it always hit the 90 deg cap and did nothing. Replaced by `catalogInFOVMask` from `RMS.Astrometry.StarFilters`. It passes the image JD, so the cone follows the sky at the image time (see #1002).
- `geoMarkerSize` scaled by the `.config` width/height rather than the displayed image; uses the loaded image shape, config only as a fallback before an image is loaded.
- `filterBlendedStars` used `currentFrameTime()` while `fitPickedStars`, both outlier filters, the quick-align paths and the catalog projection all use `currentTime()` (chunk mid-time); now consistent. `currentFrameTime()` remains in the manual-reduction paths only.
- `updatePointingIndicator` swallowed every exception and just hid the HUD; it now prints the exception (the file's convention) before hiding.

**`Utils/SkyFit2.py`** (0d1bc222)
- Brush painting: the first point of a stroke was a disc of radius r while the interpolating segment was a line of width 2r with square ends, so stroke ends looked different from the middle; coordinates were not clamped; `brushStrokeBegin()` did not reset the last position, so a new stroke could connect to the previous one. Now a disc is stamped at every position plus a gap-filling line, the centre is clamped to the layer, and the reset happens in `brushStrokeBegin()`.
- `changeMode` (SkyFit <-> manual reduction) left brush mode and the panning lock active; it now exits brush mode, closes an open mask polygon and refreshes the panning state.
- `keys_pressed` grew on key auto-repeat (many press events, one release), leaving stale entries that the `Key_0 in keys_pressed` checks read. A key is only appended once. (The suspected double append via `eventFilter` was not the mechanism: the filter consumes the event.)
- `event.pos()` on a `QMouseEvent` is gone on Qt6; uses `event.position().toPoint()` when available.

**`Utils/SkyFit2.py`** (ab8246ae)
- `loadPairs` replaced the pairs without checking the saved JD against the current image; a warning dialog now appears if it differs by more than one frame block (pairs still load). Each saved star is re-looked-up in the loaded catalog by angular nearest neighbour (0.01 deg, the tolerance `fitBandRatio` uses) so exact catalog RA/Dec/mag are restored; the status message reports how many were not found.
- `pairsHint` re-read and parsed the pairs JSON on every image navigation; `_readPairsFile()` caches by (path, mtime) and is shared with `loadPairs`.
- `runInBackground` could be re-entered while a long operation (astrometry.net solve) was running; a busy flag now refuses the second request with a status-bar message and returns None (both callers already handle None), cleared in `finally`.

**`Utils/SkyFit2.py`, `RMS/Routines/CustomPyqtgraphClasses.py`** (c0c9913d, 3e795339)
- All bare `except:` are `except Exception:` (none remain in either file).
- Dead / commented-out code removed (including the whole commented "standard coordinates / TEST" block, not only the flagged lines).
- The "Auto Fit (placeholder)" button label now says what it does: "Auto Fit on placeholder image" when the best CALSTARS frame is not on disk and a striped placeholder is generated, plain "Auto Fit" otherwise.
- `HistogramLUTItem`: `setLevels(*self.saved_manual_levels)` with `saved_manual_levels` still `None` raised; guarded at both sites (toggle and `imageChanged`). Only reachable if `auto_levels` is set externally; the guard turns a crash into a no-op.
- `PointingIndicator.boundingRect` did a `mapToDevice` round-trip on every paint/geometry query; the device scale is cached and refreshed in `setData()` and `refresh()` (already connected to `sigRangeChanged`). Identical drawing.

**`Utils/CalibrationReport.py`** (06ea1099)
- `image_stars[:, 6]` (S/N) guarded for CALSTARS files with fewer columns (`lm_info = None`, already handled downstream).
- The limiting-magnitude fit re-derived `lm_radius` from scratch because `photometryFitRobust` silently overwrites `radius_arr` / `star_intensities` with the inlier subset. The full arrays are now captured before the fit as `lm_intens` / `lm_radius`; the previous code captured intensities after the filter, so the S/N fit saw an inconsistent pair.

**`Utils/SkyFit2.py`, `RMS/Routines/MaskImage.py`, `RMS/Routines/CustomPyqtgraphClasses.py`** (fe3d0f13, from review feedback)
- Loading a mask saved for another frame size decomposed the raster at its original size and scaled the polygon vertices, which does not reproduce the nearest-neighbour resize of the raster (a run of N pixels scales to N*scale, scaled vertices span one less), so the editable overlay disagreed with the mask given to star detection along the right and bottom edges. The raster is now resized first and everything derived from it; `resampleMaskLayers` has no caller left and is removed. Two tests replace its tests: decomposing the resized raster composites back to exactly that raster, and scaling source-size vertices does not.
- `PointingIndicator` assigned the cached device scale and step size that `boundingRect()` reads before calling `prepareGeometryChange()`; Qt needs the notification first so the stale bounds leave the scene index.

## De-duplication (behaviour preserving)

**`RMS/Routines/MaskImage.py`, `Utils/SkyFit2.py`, `Tests/TestMaskBrush.py`** (ee8b702c)
- Mask compositing (polygons + brush paint layer + inversion -> mask image) existed three times in SkyFit2 and a fourth time re-implemented in the test. `MaskImage.py` gains `compositeMaskLayers`, `decomposeMaskImage`, `resampleMaskLayers`, `paintBrushSegment` and the `PAINT_*` constants; all three SkyFit2 sites call them and the test imports them (27 tests). While doing so: `loadMaskFromFile` only resampled the paint layer when the mask size differed from the image; it now rescales polygons, paint layer and the detection `MaskStructure` together.

**`Utils/SkyFit2.py`** (e0fb989e)
- The `lm_callback`, `iteration_callback`, pre-final-fit photometric/blend filters and the post-fit sigma-clip block were near-verbatim copies across `tryQuickAlignment`, `testQuickAlignWithConfigMagLimit` and `getInitialParamsAstrometryNet`. Factored into `_makeLMCallback`, `_makeIterationCallback`, `_filterPairsBeforeFinalFit`, `_sigmaClipPairsAndRefit`; bodies moved verbatim.

## Geometry (verified, no change needed)

- `screenNudgeToAzAltDelta`: a step pans the view toward the sky shown in the pressed direction, exactly `key_increment` degrees of arc for any roll (tested to 1e-4 deg over 4 pointings x 4 keys, antisymmetric for opposite keys). No sign errors. `Tests/TestSkyFitHelpers.py` (21 tests) covers this, including the azimuth wrap across north.
- `near_pole` using `abs(alt)`: both poles are singular for `rotationWrtHorizon`, so the `skip_rot_update` regime is right at both; `checkParamRange` clamps alt to +-89.999 anyway.
- `az_centre += d_az` without a wrap: `checkParamRange()` wraps with `(az + 360) % 360` and `d_az` is already in (-180, 180], so the result is exact.
- Pointing-indicator compass: the notch moves toward the bar end that WASD pans toward; self-consistent, now documented in `paint()`. Inherent quirk: the notch jumps end to end when azimuth crosses 180 deg.
- Observation, not changed: the indicator evaluates `fovCentreZenithDirection` at the distortion centre, whose apparent az/alt differs from `az_centre/alt_centre` by ~0.1-0.3 deg in the RMS projection model, so its readout differs slightly from the Fit Parameters fields. By design per the code comment.

## Testing

- `python -m compileall -q RMS Utils Tests`; `QT_QPA_PLATFORM=offscreen python -c "import Utils.SkyFit2"`.
- `python -m pytest Tests -q -o python_files="Test*.py test_*.py" --ignore=Tests/CompressionTimings.py`: 141 passed on top of #1002.
- Headless smoke run on `Tests/ExampleStationData` AU000A: `QT_QPA_PLATFORM=offscreen python -m Utils.SkyFit2 . -c .config` loads mask, catalog, CALSTARS and platepar and runs without a traceback until the timeout.
- All touched files LF; ASCII only.

## Manual GUI checks for a reviewer

1. Mask tab: two quick separate strokes must not be joined; stroke ends match the middle; dragging off the image edge does not crash.
2. Enter brush mode, switch to manual reduction and back: cursor gone, panning restored, mask-tab pan lock still works.
3. Load a `mask.bmp` saved at a different resolution: polygons and paint layer line up with the image; Save writes an image-sized mask.
4. WASD near the zenith and across az = 0/360; the horizon-bar notch slides the way the view moves.
5. Load matched pairs saved under different fps/time settings: warning dialog; normal load shows the "(N not found)" suffix only when the catalog LM excludes saved stars.
6. Click Auto Fit twice while astrometry.net is solving: "Another operation is still running".
7. Hold a WASD key (auto-repeat), release, then use the `0` key path that reads `keys_pressed`.
8. Histogram tab: toggle auto levels before and after loading a new image.
9. Geo point marker sizes on a station whose image size differs from the `.config` width/height.


dvida:
Both findings fixed in `fe3d0f13`.

**Mask resampling**: correct, and the fix is the one suggested. `loadMaskFromFile` now resizes the raster with nearest-neighbour first and decomposes that, so the polygons, the residual paint layer and the `MaskStructure` all come from the same pixels by construction, instead of the polygons being scaled by a different rule than the raster. `resampleMaskLayers` has no caller left and is removed. Two tests replace its tests: one asserts that decomposing the resized raster composites back to exactly that raster, the other asserts that scaling the source-size polygon vertices does **not**, so the order cannot be swapped back unnoticed.

**Geometry notification**: correct, `prepareGeometryChange()` now runs before the cached device scale and step size that `boundingRect()` reads are replaced, in both `setData()` and `refresh()`.

PR #1005, opened 2026-09-22
Review pass: astrometric fit performance (RANSAC 2.8x faster, results bit-identical)

dvida:
Part of the September 2026 release review (#1001). Performance work on the new nearest-neighbour RANSAC astrometric fit and the helpers around it. **Stacked on #1002**; the base retargets to `prerelease` when #1002 merges. A style pass on the same files follows in a separate PR.

Every change was verified against a saved baseline dump of the fitted parameters, matched pairs, full RANSAC iteration log, filter outputs and keep masks for two seeds: all outputs are **bit-identical** before and after. Equivalence tests keep the old implementations as private reference functions.

## Timings

`Tests/BenchmarkAstrometryFit.py` (synthetic 1920x1080 `radial7-odd` platepar, F_scale ~27 px/deg, 3000 catalog stars in the FOV, 2275 detections including 206 false ones; best of 3):

| Routine | Before | After |
|---|---|---|
| Full RANSAC `fitAstrometry` | 11.9 s | 4.29 s |
| `fitPointingNN` | 0.314 s | 0.297 s |
| `filterBlendedStars` | 49 ms | 5 ms |
| AutoPlatepar detection association (2069 x 2275) | 12.3 ms | 0.57 ms |
| Final catalog match (2275 x 3000) | 118 ms | ~3 ms |
| ExtractStars duplicate removal (2616 detections) | 0.67 ms | 0.52 ms |

## Changes

**`RMS/Astrometry/ApplyAstrometry.py`, `RMS/Formats/Platepar.py`** (1b567fd2, largest win, found by profiling)
- `getPairedStarsSkyPositions` called `xyToRaDecPP` with a per-star time list, which did a per-star `date2JD` (5.6 million calls over a fit) and ran the per-star Python `calculateMagnitudes` / `correctVignetting` loop whose output was discarded: 11.5 s of a 14.4 s profile. The JD is now converted once (`date2JD(*jd2Date(jd))`, bit-identical to the per-star value) and broadcast with `jd_time=True`, and `xyToRaDecPP` accepts `level_data=None` to skip photometry, returning `magnitude_data=None`. Additive API: no existing caller passes `None`; the return tuple shape is unchanged.

**`RMS/Formats/Platepar.py`** (c9f38add)
- The per-iteration RANSAC scoring and the final match each built a full N x M `angularSeparation` matrix and then resolved one-to-one matches in a Python loop. New `_nearestCatalogStars` helper: `cKDTree` on unit vectors gives the nearest index, then the separation of each matched pair is recomputed with `angularSeparation`, so the numbers are exactly those the dense matrix produced (chord-to-angle would differ at the 1e-12 rad level). Non-finite detections get index 0 / NaN as a NaN row did. The greedy "closest claim wins" one-to-one rule is `np.unique(nearest[argsort(sep)], return_index=True)`, proven equal to the loop in the tests (600 detections onto 150 catalog indices, heavy conflicts).

**`RMS/Formats/Platepar.py`** (ac12e1e0)
- `fitPointingNN`: the residual is "each detection to its nearest *projected catalog* star", and the projected catalog moves on every Nelder-Mead evaluation, so the tree cannot be built once without changing the residual (that was considered and rejected). Instead the tree build uses `balanced_tree=False, compact_nodes=False` (0.31 -> 0.14 ms, exact nearest distances identical, tested) and the static detection array is stacked once per call. The Cython projection dominates the remaining cost. `Tests/BenchmarkAstrometryFit.py` lands here.

**`RMS/Formats/Platepar.py`** (64d83233)
- Two `copy.deepcopy(self)` per RANSAC iteration replaced by one `copy.copy(self)` reused for the FOV pre-filter and the scoring, after verifying that the projection functions only read the platepar, the Cython code never writes the polynomial arrays, and every attribute change on the copy is a rebinding rather than an in-place write. Measured as noise (a Platepar deepcopy is ~16 us), kept because it removes the question.

**`RMS/Astrometry/StarFilters.py`** (8f8be075)
- `filterBlendedStars` dense `(n_matched x n_catalog)` distance matrix -> `cKDTree.query_ball_point` with the per-star radius array, the exact `0.1 < d < r` rule re-applied on the candidates. A negative ball radius returns every point in SciPy, so non-positive / NaN radii are queried with 0 (tested with 0, -1, NaN and 40 px FWHM).

**`RMS/Astrometry/AutoPlatepar.py`** (b7d76dfc)
- Per-star `argmin` over all detections in a Python loop -> module-level `associateDetections()` using one `cKDTree.query` with `distance_upper_bound` and the same strict `< NN_PAIR_ASSOC_RADIUS_PX` test; returns -1 for unassociated stars. The old degenerate behaviour that any non-finite detection coordinate disables all associations (the argmin picked the NaN) is preserved on purpose; see notes.

**`RMS/ExtractStars.py`** (2bc7dacf)
- PSF duplicate removal extracted into `duplicateDetectionKeepMask()` and made cheap with Python lists. The rule is inherently order-dependent (a chain 0-1-2 with intensities [1, 2, 3] keeps {2} in one pair order and {0, 2} in the other), so it was **not** vectorised; doing so would change results.

**`RMS/Astrometry/StarFilters.py`** (0bc9f225, from review feedback)
- `cKDTree.query_ball_point` only accepts `return_sorted` from SciPy 1.2 while `requirements.txt` allows `scipy>=1.0.0`, so the argument would raise a `TypeError` instead of filtering. Only membership is used, so it is dropped.
- `requirements.txt` now declares `scipy>=1.6` (b0beea5d). `RMS/ExtractStars.py`, in code that arrived with `prerelease`, calls `query_pairs(..., output_type='ndarray')`, which exists only from SciPy 1.6, so the declared floor was below what the code needs. The set-based return is not a substitute: it gives the same set of pairs in a different order, and the duplicate-removal rule is order dependent. Raising the floor also drops Python 3.6, which costs nothing, as the pinned `pyqtgraph>=0.12` already requires Python 3.7 or newer.

**Skipped**: the extra `residual_func(x0)` call in `_lstsqFit`. Its value is used (the residual count selects `'lm'` vs `'trf'`, and `lm` raises when there are fewer residuals than parameters); it costs 6 of 553 residual evaluations in a whole RANSAC run.

## Remaining cost (not changed)

After this batch the RANSAC profile is 4.7 s, of which ~2.7 s is `cyXYToRADec` inside `xyToRaDecPP` and ~1.4 ms per 3000 stars is `raDecToXYPP`: the Cython projections themselves. The `iteration_callback` still deep-copies per iteration, deliberately, because the (debug-only) callback may retain the object.

## Testing

- `Tests/TestAstrometryPerfEquivalence.py`: 23 tests. Dense-matrix nearest vs KD-tree (indices equal, separations to 1e-12; NaN detections; prebuilt-tree path), greedy one-to-one loop vs `np.unique`, sliding-midpoint tree vs default tree vs brute force, end-to-end `fitPointingNN` recovering a 0.5 deg offset on the synthetic field, `filterBlendedStars` vs the copied dense implementation, `associateDetections` vs the argmin loop (incl. empty and NaN inputs), duplicate removal vs the inline loop on clustered chains with intensity ties.
- `python -m pytest Tests -q -o python_files="Test*.py test_*.py" --ignore=Tests/CompressionTimings.py`: 136 passed on top of #1002.
- `python -m compileall -q RMS Utils`; imports of Platepar, StarFilters, AutoPlatepar, ExtractStars.
- Independent diff review found no issues.

## Reviewer notes

- `xyToRaDecPP(level_data=None)` returns `magnitude_data=None`; no in-tree caller passes `None`.
- `_nearestCatalogStars` differs from the old code only if two catalog stars are at *exactly* the same distance from a detection (KD-tree tie order vs argmin lowest index).
- `associateDetections` keeps the old "any non-finite detection disables all associations" behaviour for strict preservation; arguably a latent bug worth a separate fix.
- The nested cost function in `fitPointingNN` now takes `img_coords` instead of `img_x, img_y`; it has no external callers.
- Commits 1b567fd2 and c9f38add carry only the `Co-Authored-By` trailer; the later five also carry the `Claude-Session` trailer.


dvida:
Fixed in `0bc9f225`: `return_sorted` dropped from the `query_ball_point` call, since only membership is used. Comment added so it is not reintroduced.

While checking the rest of the new code against `scipy>=1.0.0`, one related thing worth your decision, **not introduced by this PR**: `RMS/ExtractStars.py` uses `cKDTree.query_pairs(..., output_type='ndarray')`, which exists only from SciPy 1.6. That call came in with `prerelease` itself (it is in `origin/prerelease`, not in `master`), so the declared floor is already below what the code needs. I did not change it, because the duplicate-removal rule is order dependent and the set-based return has an unspecified iteration order, so swapping it would change which detection of a cluster survives. The choice is between raising the declared floor in `requirements.txt` to `scipy>=1.6` and adding a fallback that accepts the different ordering. Happy to do either.

dvida:
Floor raised in `b0beea5d`, as requested by the maintainer: `scipy>=1.0.0` becomes `scipy>=1.6`, the first release with `query_pairs(output_type=...)`.

I checked that the set-based return could not simply be substituted: it yields the same **set** of pairs but in a different order, and the duplicate-removal rule is order dependent, so a fallback would change which detection of a dense cluster survives.

Raising the floor also drops Python 3.6, which costs nothing here: the pinned `pyqtgraph>=0.12` already requires Python 3.7 or newer, so the 3.6 markers elsewhere in the file were already unsatisfiable. The blend-filter comment that referenced the old 1.2 caveat is updated.

PR #1007, opened 2026-09-22
Style pass: astrometry, calibration and detection code (no behaviour change)

dvida:
Part of the September 2026 release review (#1001). RMS style pass over the code that `prerelease` added or changed in the astrometry, calibration, catalog and detection modules. **Stacked on #1005**; the base retargets as the parents merge.

**No behaviour change.** Verification is stronger than a read-through: every one of the 36 touched `.py` files was parsed before and after, docstrings stripped, `__future__` imports dropped and the `print` to `log` renames normalised back, then the full ASTs compared. **All 36 are AST-identical.** The three `.pyx` files were compared with comments and docstrings stripped and are code-identical. The one functional commit is the benchmark import-path fix described at the end.

## What was done

726 comment lines added, 457 reworded or removed, 17 `### Section ###` banners, 22 new docstrings, 87 docstrings rewritten into RMS form, 8 `from __future__` headers added, and 266 existing docstrings left byte-identical. Largest footprints:

| File | Comments +/- | New docstrings | Rewritten | Banners |
|---|---|---|---|---|
| `RMS/Formats/Platepar.py` | +196 / -183 | 0 | 18 | 2 |
| `RMS/Astrometry/AutoPlatepar.py` | +87 / -47 | 0 | 7 | 6 |
| `RMS/Astrometry/AstrometryNet.py` | +51 / -51 | 0 | 3 | 0 |
| `RMS/Astrometry/StarFilters.py` | +32 / -1 | 0 | 3 | 2 |
| `Tests/TestAstrometryPerfEquivalence.py` | +30 / 0 | 0 | 7 | 2 |
| `RMS/Astrometry/NNalign.py` | +29 / -10 | 0 | 2 | 0 |
| `RMS/Routines/SatelliteTLEsDownload.py` | +25 / -8 | 0 | 2 | 0 |
| `RMS/ExtractStars.py` | +23 / -19 | 0 | 2 | 0 |
| `RMS/Astrometry/StarClasses.py` | +20 / 0 | 0 | 14 | 2 |
| `Tests/BenchmarkAstrometryFit.py` | +21 / 0 | 1 | 10 | 3 |

17 Google-style `Args:` / `Returns:` / `Raises:` headers inside changed hunks became RMS `Arguments:` / `Return:` sections (`AutoPlatepar` 6, `StarClasses` 9, `Platepar` 8, `StarFilters` 2, `AstrometryNet` 1, `Image` 1). `Tests/test_Grouping3Dcy.py` was renamed to `Tests/TestGrouping3Dcy.py` to match the convention, and `from __future__` headers were added to `MatchStars.py`, `SatelliteTLEsDownload.py` and six test files.

Nine further `Args:` / `Returns:` sites were checked against the changed-hunk ranges and are **master-era**, so they were left alone: `Conversions.py:578`, `AstrometryNetNova.py:703`, `StarCatalog.py:407`, `Math.py:462,490,519`, `ApplyRecalibrate.py:507`, `SatellitePositions.py:36,315`.

## Console output routed through the logger

In the NN-RANSAC fit path of `RMS/Formats/Platepar.py`, 21 `print` calls became logger calls with byte-identical message text: **16 `log.info`** (once per fit: input counts, RANSAC setup, the two distortion-type switches, the final catalog line, early-exit reasons, the RANSAC result, dropped duplicates, final fit size) and **5 `log.debug`** (per-iteration optimizer and outlier diagnostics).

Two prints in the forward-reverse loop were **deliberately left as prints**: `Utils/SkyFit2.py` wraps its cross-validation folds in `contextlib.redirect_stdout` to suppress exactly that output, and a logging handler holds its stream from construction time, so redirecting stdout would not reach it. Converting them would make every fold spam the log. That is the only stdout capture in SkyFit2, and it suppresses rather than displays, so nothing else depended on these messages.

`RMS/Astrometry/AutoPlatepar.py` kept all 80 of its prints: 68 are inside `if verbose:`, 8 inside `__main__`, and 12 are the body of `printFitResiduals()`, which exists to print and is only called under `verbose`. No module logger was added there, as it would be unused.

## Functional commit

`Tests/BenchmarkAstrometryFit.py` run directly (`python Tests/BenchmarkAstrometryFit.py`) put `Tests/` first on `sys.path`, so `RMS` resolved to an **installed** copy rather than the checkout: the duplicate-removal benchmark silently fell back to its inline reference and the RANSAC benchmark reported 3.0 s instead of 1.1 s. The repository root is now inserted first, so both invocations measure the working tree.

## Noticed while reading, not changed

- `Platepar.py:1972`: "only the astrometric parameters **where** fitted" should be "were". Master-era and a user-visible string, so left for a maintainer.
- Eleven lines in the Platepar logging block are 111-127 columns. They were already over the limit as prints; the rename added a few characters. Left rather than restructuring statements.
- `buildPairedStars(pp, ...)` in the benchmark never reads `pp`; kept for a uniform builder signature and documented as such.

## Test discovery

`pyproject.toml` gains `python_files = ["Test*.py", "test_*.py"]`. The `TestXxx.py` convention that most of the suite follows never matched pytest's default globs, so a plain `pytest Tests` collected a small fraction of the tests and silently skipped the rest; renaming a module to the convention in this PR would have hidden it too. With the setting, a plain `pytest Tests` collects the whole suite (137 tests, up from 20 here). The identical block is added on #1006, so the two merge without a conflict.

## Testing

- `python setup.py build_ext --inplace` (8 extensions); `python -m compileall -q RMS Utils Tests`.
- `python -m pytest Tests -q`: 137 collected and passed (20 were collected before the discovery fix).
- `python Tests/BenchmarkAstrometryFit.py` and `python -m Tests.BenchmarkAstrometryFit` now agree.
- `python -m RMS.Astrometry.AutoPlatepar --help`; all 39 touched modules import.
- LF endings and ASCII-only confirmed on every touched file; no branch-introduced line over 110 columns.


dvida:
Fixed in `fde1065f`, and the finding turned out to be bigger than the one renamed file: the `TestXxx.py` convention that 15 of the test modules already follow has never matched the default discovery globs, so `pytest Tests` collected **20 of 137 tests** on this branch and silently skipped the rest. That is also why the PR descriptions in this series quote `pytest -o python_files=...`.

Rather than renaming back to the minority spelling, `pyproject.toml` now sets `python_files = ["Test*.py", "test_*.py"]`, so both spellings are collected and a plain `pytest Tests` reports 137 passed. The identical block is added on #1006, which renamed `test_DeadlockFixes.py` for the same reason; the two files are byte-identical so they merge without a conflict.

PR #1010, opened 2026-09-22
Second review pass: astrometry, calibration and detection fixes

dvida:
Part of the September 2026 release review (#1001). A second, independent review of the whole release, run on the combined code of the first-pass PRs, found defects the first pass missed. This PR fixes those in the astrometry, calibration, catalog and detection code. **Stacked on #1007** (the tip of the astrometry stack) so it applies on top of everything already reviewed; the base retargets as the parents merge.

Each fix was reproduced first, has a regression test that fails without it, and has a commit message stating the defect, the evidence and the fix. An independent review of this branch found no P0 or P1 problems; its seven P2 findings are addressed in the last seven commits. 236 tests pass (142 before, 94 new).

## P0 / P1

- **Out-of-bounds write in fireball detection, default config (pre-existing, not from this release)**, `Grouping3Dcy.thresholdAndSubsample` (2e0fb35a, c7fc0789). The block-count array has `H//f` rows but the loop covers all `H` rows, with bounds checks off. On 1920x1080 with the default `subsampling_size: 16`, rows 1072-1079 wrote into the next frame's counts, and for maxframe 255 past the end of the buffer. This runs in the live Extractor on most stations. The partial edge block is now skipped; in-range pixels are unchanged. The shape variables are also typed, which makes the call 2.5x faster.
- **Stars lost when the configured bit depth does not match the data** (6fa2f3e7). Extraction thresholds, the brightness cutoff and the gamma white point scaled with `config.bit_depth`, but live FFs are always 8-bit, so a station with `bit_depth: 12` extracted **no stars** (57/37/0/0 stars for 8/10/12/16 on the same 8-bit image). New `Image.effectiveBitDepth(img, bit_depth)` caps the depth at the data's integer width; the same image now gives 57 stars at every setting.
- **Nightly recalibration rejects good fits on deep cameras** (4e3b7db0). The new star-coverage gate divided matches by all detections, but only catalog stars down to the catalog limit can match. On a real US05E1 night at LM 4.5 all 11 good fits were rejected and photometry was ~0.3 mag off. The denominator is now `min(detections, catalog stars in the image)` (`coverageMatchFraction`); all 11 are accepted, and a spurious sparse fit (31 of 308) is still rejected.
- **Wide-lens self-calibration could never succeed** (62e089f1, fe31087c). For fields wider than 90 deg the astrometry.net scale hint was built from the width of the centrally filtered stars and an inflated FOV, and excluded the true pixel scale for 100-150 deg lenses. The hint now contains the true scale for 60-180 deg, with and without an image centre.
- **Distortion coefficients one short for the `-all` radial models** (9d8755d7). `buildRadialCoeffs` dropped the highest-order term, so toggling a distortion flag on radial3/4/5-all changed the projection. Round-trip tested for 7 types x 3 flags; freeing a forced centre now writes the exact forced value.
- **A rejected NN-RANSAC fit left the platepar modified** (d540f56d, 89d9665b). Every rejection path now restores the platepar from a snapshot and empties `star_list`; the too-few-stars path no longer pairs detections with catalog stars by index, and poly distortion types are rejected before any fit runs.

## P2

- `fitPointingNN` (88c15cd2): pole reflection applied in the cost and the stored result (polar cameras); a refinement stage with the distance capped at the inlier threshold removes the bias from false detections (median error 0.34 -> 0.02 px at 30% false detections), at +0.2 s per call.
- Robustness (d9d83e8f, 1a89af1a, 0d515bf4): empty catalog, non-finite detections, non-positive intensities in RANSAC weights; out-of-bounds `k2` read in `CyFunctions` for radial3-odd (projections bit-identical).
- Photometry (d06ffd84, f8f3e89b): one intensity <= 0 made the zero-point fit return its initial guess with stddev 0; such stars are dropped, and a frame with a degenerate fit now receives its neighbours' average instead of keeping it. Neighbourhood averages are computed from the original fits and written back afterwards (d891021c, from Codex review), so overlapping neighbourhoods no longer double count and the zero points no longer depend on the meteor list order (one order gave 11.21, the reverse 11.185).
- astrometry.net rotation flipped 180 deg near RA 0 (53a67a0c).
- AutoPlatepar (a4097127): reference time is mid-FF (was FF start, 5.1 s off); CLI catalog gets proper motion for the data's date; frame scoring no longer penalises clear frames on sensitive cameras (unchanged on nights with a median of 100 stars or fewer).
- TLE cache (f6827404, 3c60e26e): offline fallback to the newest cache of the same group; atomic `.part` download validated as real TLE data; exact group matching; stale `.part` cleanup.
- Star catalog (c46c75c3): a newer catalog version is no longer deleted and re-downloaded on every start; permission errors keep the file.
- ExtractStars CLI printed amplitude and intensity swapped (a95f0764).
- Checked and not a defect: `NNalign.alignPlatepar` already works on a copy of the config; a test now guards it (5376c6ce).

## Behaviour changes a reviewer should know about

- Stations with `bit_depth` above 8 and 8-bit FFs get 8-bit thresholds: star counts go up, often from zero.
- More FFs on sensitive cameras become `auto_recalibrated`, and their photometry changes accordingly.
- NN-aligned pointings shift slightly (more accurate) wherever `alignPlatepar` is used.
- AutoPlatepar's reference time moves to mid-FF.
- On 1080p with f = 16 the bottom 8 rows are no longer searched for fireballs; they used to alias into other blocks.

## Merge note

When this PR and #1008's successor land in `prerelease` together, one comment in `Platepar.fitAstrometry` conflicts (this stack restyled it, the SkyFit2 second-pass PR extends it). Resolution: keep the restyled text plus the added clause "unless there are too few stars for the distortion fit". The combined second pass was trial-merged: it builds, imports and passes 422 tests.


dvida:
Fixed in `d891021c`. The finding is right, and it is broader than the repaired neighbour: the averages were written back while iterating, so where neighbourhoods overlap any already-averaged value (the repaired degenerate fit included) counted again as an independent measurement. That both double counts and makes the zero points depend on the meteor list order. With two meteor FFs two apart, one order gave 11.21 and the reverse 11.185.

`averageNeighbourPhotometry` now computes every neighbourhood average from a snapshot of the fits taken before any averaging, and only then writes them back. Neighbours are written first, and each meteor FF is written last with the average of its own neighbourhood, so the result is order independent. The same example now gives 11.05 / 11.175 in both orders. Two new tests cover order independence and that a repaired fit never contributes; both fail on the previous code. Full suite: 236 passed.

PR #1011, opened 2026-09-22
Second review pass: capture, logging, processing state and ops fixes

dvida:
Part of the September 2026 release review (#1001). A second, independent review of the combined release found defects the first pass missed in the capture, multiprocessing, logging, processing-state, config and ops-script code. This PR fixes them. **Stacked on #1006** (the tip of the pipeline stack); the base retargets as the parents merge.

Each fix was reproduced first, has a regression test that fails without it (unless noted), and has a commit message with defect, evidence and fix. An independent review of this branch found one regression introduced by the branch itself (Compressor shutdown grace period, item 13) plus P2 items; all are fixed in the last six commits. 167 tests pass (93 before); the new `Tests/TestPipelineReview2.py` (74 tests) passed three runs in a row, its multiprocessing tests under both fork and forkserver. CRLF files stay 100% CRLF.

## P0 / P1

1. **Child processes lost the bounded logging** (5d268346). `initChildLogging` replaced the drop-on-full `_DroppingQueueHandler` with a plain `QueueHandler`. With the queue full, `queue.Full` tracebacks went to stderr, which in a forked child is the `LoggerWriter`, so every log call recursed on its own error output: reproduced at 150 MB of stderr in 10 s with log calls never returning. That would stop FF production for the rest of the night. Children now use `_DroppingQueueHandler`.
2. **Exit could hang in the logging-queue feeder join** (534b37d2). `shutdownLogging` never cancelled the join, and children writing to an abandoned queue after a pipeline restart hung at normal exit; a wedged main process keeps the station lock, so respawns were refused. `shutdownLogging` joins the listener and cancels the join; children register an exit hook that runs the bounded flush before the feeder join, so no records are lost in the normal case (verified at 10 / 5000 / 50000 records, fork and forkserver); `EventMonitor.stop()` is bounded.
3. **Orphaned capture children could make a second capture** (23b39afd). Only RawFrameSaver armed a parent-death signal, so if StartCapture died (OOM, crash, SIGTERM) BufferedCapture and the Compressor kept running, the flock was released, and a respawn ran a second capture on the same camera. BufferedCapture, Compressor, UploadManager and EventMonitor now arm `PR_SET_PDEATHSIG` and check the parent PID at start; all four are started from the StartCapture main thread (verified, including the watchdog restarts). A SIGTERM handler was deliberately **not** added: the existing shutdown paths rely on KeyboardInterrupt unwinding.
4. **Unbounded joins after SIGKILL** in `BufferedCapture.stopCapture()` (18fb1eec) and `UploadManager.stop()` (dd8c656c): a process in D-state I/O wedged the end-of-night stop. Bounded, then abandoned with a warning.
5. **StartCapture crashed at high-latitude stations on the last night before polar day** (13c93b49). `next_setting` sat outside the new polar guards; `AlwaysUpError` escaped the main loop. Reproduced at 72/74/76/80 N, McMurdo and 89.9 N; when a sunrise exists and the Sun is below the horizon it is night now, so capture runs until sunrise. Swept 19 latitudes hourly over 2026: every case now returns a positive duration.
6. **A night could be reprocessed and re-uploaded on every start** with `reprocess_if_archive_missing` (7dff69fa). The attempt counter reset after success, so a processed night whose archive was purged by the free-space logic looped forever. Such nights are flagged and reprocessed once.
7. **`clock_ahead_ms` was twice the true offset** (81083a6a) on stations without chrony (including the Raspberry Pi OS default timesyncd). The remote time estimate is now `T4 + offset`; failures return three values and close the socket.

## P2

- ObservationSummary (345512a1, ffd6607c): if archiving failed after finalize, auto-reprocess rebuilt the summary without start-of-session metadata; the working dict is now seeded from the final JSON with only the start-of-session keys (`OBSERVATION_SUMMARY_SESSION_KEYS`), so conditional keys such as `photometry_good` are never carried over stale.
- UploadManager held its locks through a 1 s sleep, so the parent's acquisitions timed out with false "process died" warnings (d124bd5f).
- The nightly detector pool was created without `config`, so worker debug output went unfiltered into the bounded queue (be0446e5).
- `runWithTimeout` race that could skip the late-cleanup callback and leak an SSH client or socket (6078c53e).
- Continuous mode (f28f7587, dd495a19): capture no longer stops for a reboot wait of up to 4 h; it is stopped right before the reboot command, at dusk if the wait reaches night, or before capture resumes if no reboot happens. The "no second capture" guarantee of #919/#833 holds on every path (traced by the independent review). The retry loop moved into `tryRebootAfterProcessing()`, logic unchanged apart from the dusk stop.
- Compressor stop (0e5284c6, 4551bb58): the SIGINT step was a no-op (children ignore SIGINT); it is dropped, but the 5 s join before escalating is kept, which is the grace period a normally exiting Compressor needs to flush its logs.
- A stale lock file owned by another user blocked every start (5c15035e).
- An empty value for `udp_buffer_size`, `video_scale_*`, `auto_reprocess_max_attempts` or `ml_model_file` stopped StartCapture from starting (b2f26a77).
- Formats: float FITS sequences got a new offset per frame and NaNs cast undefined (ed6ce53c); 16-bit FF FITS could not be read with the default memmap (7dd3e258); `FRbin.write` always crashed and `read` leaked its file (2484aa61); FOVKML crashed on an empty or fully masked side (6ed5b3cf).
- Scripts: GRMSUpdater's kernel check asked for a spurious reboot on Raspberry Pi OS and missed updates on `-rpiN` kernel names (f95e3285, e454bc53); in RMS_Update.sh `((n++))` with n = 0 aborted the build retry after all `*.so` were deleted, and the interactive branch selection (6fd667d2).

## Operator-visible changes

- No more false nightly "Compression process still alive, forcing termination" warning.
- If StartCapture dies, capture, compression, upload and EventMonitor die with it.
- Polar stations start capture on the last night before polar day instead of StartCapture exiting.
- In continuous mode a pending reboot no longer stops the daytime capture while it waits.
- `clock_ahead_ms` halves on stations without chrony (it was reported double).
- Empty values for the options above keep their defaults.


dvida:
Both findings confirmed, fixes in progress on this branch.

**Parent death under forkserver:** reproduced. A child that arms `PR_SET_PDEATHSIG` is reparented to the fork server, and the fork server does not exit when StartCapture is SIGKILLed. Under `fork` the child dies at once; under `forkserver` it was still alive 3 s after the main process was killed. The earlier review had assumed the fork server exits with its parent, which is not the case. The fix keeps the death signal (it covers `fork` instantly) and adds a throttled logical-parent check (the recorded StartCapture PID) inside the long-running loops of BufferedCapture, Compressor, UploadManager and EventMonitor, with a forkserver regression test that SIGKILLs the parent.

**Float FITS mapping:** the sequence mapping will only be committed once a frame with a finite, nonzero range is seen, with a per-frame fallback until then.

dvida:
Both findings fixed and pushed.

- **Forkserver orphans** (`ae463828`): new `Misc.startParentWatch()` runs a daemon thread in `run()` of BufferedCapture, Compressor, UploadManager and EventMonitor. It checks the recorded StartCapture PID once a second. When the parent is gone it sets the child's own exit flag, so the child leaves through its normal exit path, and it exits hard after a grace period (15 s, 45 s for the Compressor, whose normal exit waits for its last buffers). A thread rather than checks in the loops, so the per-frame path gets no extra syscall and it still works when the main thread is blocked in camera I/O or an upload. `testOrphanStopsAfterMainSigkill` SIGKILLs the parent under fork and forkserver; with the watcher disabled the forkserver case fails. The fork server then also exits within about 1 s.
- **Float FITS mapping** (`834f0428`): the sequence mapping is committed only from an image with a finite, nonzero range; until then each image is mapped on its own. Test: a constant or all-NaN first image followed by a [0, 1] image maps to the full 16-bit range.

Full suite 173 passed; `TestPipelineReview2.py` passed three runs in a row.

